PR #5, opened 2017-09-09
NOJIRA Let external API clients see error statuses

raydavis:
Needed to distinguish Canvas API failures from lack of a Canvas user account.

### DIFF
--- a/boac/lib/http.py
+++ b/boac/lib/http.py
@@ -10,9 +10,7 @@ def request(url, headers):
         response.raise_for_status()
     except requests.exceptions.RequestException as e:
         app.logger.error(e)
-        return None
-    else:
-        return response
+    return response
 
 def sanitize_headers(headers):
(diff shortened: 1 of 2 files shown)


boac/lib/http.py
```diff
@@ -10,9 +10,7 @@ def request(url, headers):
         response.raise_for_status()
     except requests.exceptions.RequestException as e:
         app.logger.error(e)
-        return None
-    else:
-        return response
+    return response
```
paulkerschen: @raydavis Returning an informative error makes sense, but I think (without having tried it out) that `response` should be initialized to a default value; otherwise a timeout or network error might cause it not to get assigned, and we'll get a  `NameError: name 'response' is not defined` on attempting to return it.

raydavis: Thanks for the unexpected weekend review, @paulkerschen -- I'll close the PR & look into that.

PR #6, opened 2017-09-09
BOAC-35, code style fixes: import order, etc.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-35

Code style fixes and removed corresponding codes from `flake8 --ignore` list.

A question for the team: Where do we stand on [trailing commas in Python data structures](https://coderwall.com/p/4fvcaa/trailing-comma-in-python-data-structures) (as enforced by `flake8-commas`)? 

Also: https://wiki.python.org/moin/TupleSyntax (see "...essential element here is the trailing comma...")

### DIFF
--- a/boac/api/status_controller.py
+++ b/boac/api/status_controller.py
@@ -1,7 +1,6 @@
+from boac.externals import canvas
 from flask import current_app as app, jsonify
 from flask_login import current_user
-
-from boac.externals import canvas
 
 
 @app.route('/api/status')
@@ -11,10 +10,10 @@ def app_status():
         'is_authenticated': current_user.is_authenticated,
         'is_active': current_user.is_active,
         'is_anonymous': current_user.is_anonymous,
-        'uid': uid
+        'uid': uid,
     }
     resp = {
-        'authenticated_as': authn_state
+        'authenticated_as': authn_state,
     }
 
     if current_user.is_active:
--- a/tox.ini
+++ b/tox.ini
@@ -11,24 +11,27 @@ skipsdist = True
 [testenv]
 whitelist_externals = *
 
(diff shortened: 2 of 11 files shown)


boac/api/status_controller.py
```diff
@@ -11,10 +10,10 @@ def app_status():
         'is_authenticated': current_user.is_authenticated,
         'is_active': current_user.is_active,
         'is_anonymous': current_user.is_anonymous,
-        'uid': uid
+        'uid': uid,
```
johncrossman: Trailing comma looks out of place but there are some legit considerations:
* https://wiki.python.org/moin/TupleSyntax (see "...essential element here is the trailing comma...")
* See bug described in https://coderwall.com/p/4fvcaa/trailing-comma-in-python-data-structures

raydavis: I could get used to it, now that I know it's acceptable.

tox.ini
```diff
@@ -11,24 +11,27 @@ skipsdist = True
 [testenv]
 whitelist_externals = *
 
-[testenv:pytest]
+[testenv:test]
 commands =
     pytest {posargs}
 
-[testenv:eslint]
+[testenv:lint-js]
```
johncrossman: Fyi, renamed targets. Genericized! Suggestions welcome.

boac/api/status_controller.py
paulkerschen: The trailing commas are FREAKING ME OUT but I'm up for trying new thing.

FWIW here's what PEP8 has to say:
https://www.python.org/dev/peps/pep-0008/#when-to-use-trailing-commas

PR #8, opened 2017-09-11
NOJIRA Let external API clients see error statuses

raydavis:
For starters, we'll want to distinguish "this user has no Canvas account" from "Canvas is being unfriendly."


paulkerschen:
Thanks @raydavis. I like this implement-`__bool__` convention a lot.

PR #9, opened 2017-09-11
NOJIRA, envlist must match updated target names

johncrossman:
Three second build-time seemed wrong.

### DIFF
--- a/boac/lib/http.py
+++ b/boac/lib/http.py
@@ -3,11 +3,13 @@ import requests
 
 
 class ResponseExceptionWrapper:
-    def __init__(self, exception, original_response = None):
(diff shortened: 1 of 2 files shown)


johncrossman:
PR includes a couple code style fixes that were missed due to my `tox.ini` goof.

boac/lib/http.py
```diff
@@ -3,12 +3,14 @@
 
 
 class ResponseExceptionWrapper:
-    def __init__(self, exception, original_response = None):
```
raydavis: After my original PR, I noticed PyCharm complaining about this & wondered how it'd passed through Travis...

PR #11, opened 2017-09-11
BOAC-36 Client and tests for Canvas student summaries API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-36

Most of the mass here is test/demo support. BIG data means BIG fixtures, adding up to a real Canvas data distribution with 730 obfuscated IDs.

### DIFF
--- a/boac/lib/mockingbird.py
+++ b/boac/lib/mockingbird.py
@@ -1,6 +1,8 @@
 from contextlib import contextmanager
 from functools import partial, wraps
 import os
+import urllib
+
 from flask import current_app
 import httpretty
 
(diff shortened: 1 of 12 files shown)


boac/lib/mockingbird.py
```diff
@@ -117,6 +119,52 @@ def external_api_mock(hostname, resource_id):
         return MockResponse(200, {}, fixture)
 
 
+def paged_fixture(pattern):
+    """Convenience function to generate a mock response for an API with multiple pages. Fixture files should be
+    saved in the format:
+      - some_resource_page_1.json
+      - some_resource_page_1.json
```
raydavis: Why didn't flake8 catch this?

raydavis:
Every ID is special in its own way.

PR #12, opened 2017-09-12
BOAC-41, auth status of 'me' as global constant; landing page gets login form

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-41

Changes:
* `app.js` (bootstrap) loads 'me' per `/api/status`
* `routes.js` uses 'me' to determine available routes
* `landing.html` gets login form and template code removed from `dev_auth.py`

**Note:** `authFactory.js` posts JSON and I could only get `dev_auth.py` to play nice by using [Flask's get_json()](http://flask.pocoo.org/docs/0.12/api/#flask.Request.get_json). The previous code was not working for me. That said, I use `get_json()` as a fallback; the unit tests rely on `request.form()` and it's still supported.


boac/static/js/app.js
```diff
@@ -1,13 +1,25 @@
...
+
+  var loadConstants = function() {
+    var initInjector = angular.injector([ 'ng' ]);
+    var $http = initInjector.get('$http');
+    var $q = initInjector.get('$q');
+
+    return $q.all({
+      me: $http.get('/api/status')
+    }).then(function(results) {
+      boac.constant('me', results.me.data);
```
raydavis: Since session state will change on login / logout, shouldn't it be an AngularJS `value` instead of `constant`?

boac/static/templates/landing.html
```diff
@@ -1,8 +1,15 @@
...
-  <div data-ng-if="status.authenticated_as.is_authenticated">
-    Hello, UID {{status.authenticated_as.uid}}.
+<div class="body-text" data-ng-controller="LandingController">
+  <div data-ng-if="me.authenticated_as.is_authenticated">
+    Hello, UID {{me.authenticated_as.uid}}.
   </div>
-  <div data-ng-if="status.authenticated_as.is_anonymous">
+  <div data-ng-if="me.authenticated_as.is_anonymous">
     Would you mind logging in?
+    <div>
```
raydavis: You're probably already planning these bits as follow-ons, but just in case:

* We need a CAS button-link, which (glory be) will always be enabled.
* We need to hide the devauth login form if devauth isn't enabled. I figured those sort of app configs could be slipped pretty easily into the api/status payload à la Junction.

raydavis:
Thank you for getting us past my block, @johncrossman !

boac/static/js/app.js
johncrossman: @raydavis A good question. I use `constant` because it's needed in `routes.js`. A `value` is not available during the `module.config` phase. (Of course, this constant is refreshed when `index.html` reloads but that might not be good enough.) Let me experiment with `value`; I bet I can find a sensible approach.

boac/static/templates/landing.html
johncrossman: @raydavis Yes, these are some follow up tasks. I didn't want to overload this PR.

boac/static/js/app.js
raydavis: Ah yes, the old routes issue... Personally, I'd vote for one of these two options, in order of preference:

1. Always enable the AngularJS routes no matter what (which is what Junction does, for example), counting on template checks to hide navigations that don't make sense and on server-side authz to block any mischievous manually entered URLs.
2. Or if we really really feel a need to block AngularJS routes on session changes, move to `ui-router`.

@paulkerschen , what do you think?

johncrossman: @raydavis Yes, I was leaning towards option one. At this point, protecting routes is a solution in search of a problem.

paulkerschen: @johncrossman @raydavis I agree with the agreed.

boac/static/templates/landing.html
raydavis: Speaking as a recent victim of PR paralysis, this makes sense.

boac/static/js/controllers/landingController.js
```diff
@@ -2,14 +2,30 @@
...
   'use strict';
 
-  angular.module('boac').controller('LandingController', function(landingFactory, $scope) {
-    var getStatus = function() {
-      landingFactory.getAppState().then(function(appState) {
-        $scope.status = appState.data;
-      });
+  angular.module('boac').controller('LandingController', function(authFactory, me, $scope) {
+
+    $scope.me = me;
```
johncrossman: @raydavis Amendment is in. Follow up PR will include logOut (directive?) and other changes mentioned. **Note:**  With a proper logOut (directive?), I'd like `scope.me` assignment, as above, to be unnecessary.

PR #14, opened 2017-09-12
BOAC-45, /api/config provides devAuthEnabled (needed in templates) and version

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-45

Front-end bootstrap will use `/api/config` such that:
```
<div data-ng-if="config.devAuthEnabled && ...">
   Log in with dev-auth credentials.
```
`/api/config` is public and contains 'version'.

### DIFF
--- a/boac/api/config_controller.py
+++ b/boac/api/config_controller.py
@@ -1,0 +1,19 @@
+import json
+from flask import current_app as app, jsonify
+
+
+@app.route('/api/config')
+def app_config():
+    return jsonify({
+        'boacEnv': app.config['BOAC_ENV'],
+        'devAuthEnabled': app.config['DEVELOPER_AUTH_ENABLED'],
+        'version': _get_app_version(),
+    })
+
+
+def _get_app_version():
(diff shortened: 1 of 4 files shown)


boac/api/config_controller.py
```diff
@@ -0,0 +1,19 @@
+import json
+from flask import current_app as app, jsonify
+
+
+@app.route('/api/config')
+def app_config():
+    return jsonify({
+        'boacEnv': app.config['BOAC_ENV'],
+        'devAuthEnabled': app.config['DEVELOPER_AUTH_ENABLED'],
+        'version': _get_app_version(),
```
johncrossman: Having separate `/api/version` seems unnecessary – adherents of convention might disagree.

paulkerschen: Works for me, @johncrossman. Separate from this, we'll presumably want some kind of invariant app-status endpoint (distinct from the current `/api/status` which returns user status) for Nagios and such.

PR #16, opened 2017-09-12
BOAC-43 External API client for Canvas course site memberships

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-43

### DIFF
--- a/boac/externals/canvas.py
+++ b/boac/externals/canvas.py
@@ -2,6 +2,7 @@ import urllib
 
 from boac.lib import http
 from boac.lib.mockingbird import fixture, mockable, mocking, paged_fixture
+from flask import current_app
 
 
 @mockable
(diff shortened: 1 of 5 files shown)


boac/externals/canvas.py
```diff
@@ -2,6 +2,7 @@
 
 from boac.lib import http
 from boac.lib.mockingbird import fixture, mockable, mocking, paged_fixture
+from flask import current_app
```
raydavis: FWIW, I like the `import current_app as app` convention that someone not-me started. I think it reassures me that it would be easy to have the module rely on an explicit call rather than continuing to rely on magic words at import time.

boac/externals/canvas.py
```diff
@@ -57,3 +66,20 @@ def build_url(canvas_instance, path, query=None):
 def authorized_request(canvas_instance, url):
     auth_headers = {'Authorization': 'Bearer {}'.format(canvas_instance.token)}
     return http.request(url, auth_headers)
+
+
+def paged_request(canvas_instance, path, mock):
```
raydavis: Thank you for sorting this out...

raydavis:
Looks great, @paulkerschen . I'll wait a few more minutes to merge, though, in case you want to respond.

boac/externals/canvas.py
paulkerschen: @raydavis Oh yes, I think it was me who started that convention, copying someone else's code I saw somewhere else. I'll follow up to stick with it, then!

PR #18, opened 2017-09-12
NOJIRA, expand scope of code linting (py, js)

johncrossman:
As promised, `flake8` covers all py files. Bonus: eslint rules unleashed.

### DIFF
--- a/config/default.py
+++ b/config/default.py
@@ -1,5 +1,5 @@
-import os
 import logging
+import os
 
 
 # Base directory.
@@ -13,9 +13,9 @@ THREADS_PER_PAGE = 2
 
 # Some defaults.
 CSRF_ENABLED = True
-CSRF_SESSION_KEY = "secret"
+CSRF_SESSION_KEY = 'secret'
 # Used to encrypt session cookie.
-SECRET_KEY = "secret"
+SECRET_KEY = 'secret'
 
 # Override in local configs.
 SQLALCHEMY_DATABASE_URI = 'postgres://boac:boac@localhost:5432/boac'
@@ -24,7 +24,7 @@ HOST = '0.0.0.0'
 PORT = 5000
 
 DEVELOPER_AUTH_ENABLED = False
-DEVELOPER_AUTH_PASSWORD = "another secret"
+DEVELOPER_AUTH_PASSWORD = 'another secret'
 
 CAS_SERVER = 'https://auth-test.berkeley.edu/cas/'
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,5 +1,5 @@
-import boac.factory
 import os
+import boac.factory
 import pytest
 
 os.environ['BOAC_ENV'] = 'test'
@@ -12,8 +12,10 @@ class FakeAuth(object):
 
     def login(self, uid):
         self.app.config['DEVELOPER_AUTH_ENABLED'] = True
-        self.client.post('/devauth/login',
-            data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']})
+        self.client.post(
+            '/devauth/login',
+            data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']},
+        )
 
 
 # Because app and db fixtures are only created once per pytest run, individual tests
--- a/tox.ini
+++ b/tox.ini
@@ -24,21 +24,22 @@ commands =
 commands =
     # Flake8 error/violation codes:
     #   http://flake8.pycqa.org/en/latest/user/error-codes.html
-    #   http://www.pydocstyle.org/en/latest/error_codes.html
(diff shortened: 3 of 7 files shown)


tox.ini
```diff
@@ -24,21 +24,22 @@ commands =
...
         --max-line-length 155 \
-        boac
+        boac config tests run.py
 deps =
     flake8
-    flake8-docstrings
+    flake8-commas
     flake8-import-order
+    flake8-pytest
+    flake8-quotes
```
johncrossman: Fyi, `flake8-docstrings` removed; no need for `D` rules.

tests/conftest.py
```diff
@@ -12,8 +12,7 @@ def __init__(self, the_app, the_client):
 
     def login(self, uid):
         self.app.config['DEVELOPER_AUTH_ENABLED'] = True
-        self.client.post('/devauth/login',
-            data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']})
+        self.client.post('/devauth/login', data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']})
```
raydavis: I don't understand this one, @johncrossman -- do you know what the motivation was?

config/default.py
```diff
@@ -13,9 +13,9 @@
 
 # Some defaults.
 CSRF_ENABLED = True
-CSRF_SESSION_KEY = "secret"
+CSRF_SESSION_KEY = 'secret'
```
raydavis: Just curious -- was this flagged by a rule, or was it you as a human being consistent about the "textual matter in double-quotes" "computery strings in single" convention?

tests/conftest.py
johncrossman: @raydavis Violation was as below. I've amended PR to give it correct indentation.
```
E128 continuation line under-indented for visual indent
```

config/default.py
johncrossman: @raydavis The robots asked for this change. Would you prefer to omit `config/` from py linting?

johncrossman: For the record, "Q000 Remove bad quotes" is the rule that was broken.

tests/conftest.py
```diff
@@ -13,7 +13,7 @@ def __init__(self, the_app, the_client):
     def login(self, uid):
         self.app.config['DEVELOPER_AUTH_ENABLED'] = True
         self.client.post('/devauth/login',
-            data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']})
+                         data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']})
```
raydavis: Ugh, personally I hate those super-long indents, and always turn off the formatters that default to them. @johncrossman , @paulkerschen , is there an alternative? PEP8 itself (as opposed to automated enforcers) just says there should be some visual distinction.

Maybe robocop would accept

```
        self.client.post(
            '/devauth/login',
            data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']}
        )
```

config/default.py
raydavis: So we're basically deciding to centralize on single-quotes no matter what (except, I hope, when the text includes a single quote)? OK by me.

johncrossman: @raydavis Good news. As an example, Flake8 allows:
```
CANVAS_HTTP_TOKEN = "BOAC's secret"
```

tests/conftest.py
paulkerschen: @raydavis's compromise proposal makes sense to me, assuming the robot will play ball.

config/default.py
paulkerschen: @johncrossman: so to be clear, the rules to keep the robots happy are:
- Double quotes for thrice-quoted docstrings;
- Single quotes for other strings;
- Unless the string contains a single quote, in which case double quotes may e used;
- If your string contains both single and double quotes, you've been at work too long and need to go home;
?

tests/conftest.py
johncrossman: @raydavis @paulkerschen Robocop seems to like:
```
self.client.post(
    '/devauth/login',
    data={'uid': uid, 'password': self.app.config['DEVELOPER_AUTH_PASSWORD']},
)
```
Works for me. Commit amended.

config/default.py
johncrossman: @paulkerschen That's right. Note: If the string contains single quote(s) then there is the option of escaping them. For example :
```
CANVAS_HTTP_TOKEN = 'BOAC\'s secret'
```
The robots are flexible. (I prefer escaping, as shown above, over the _sometimes-we-use-double-quotes_ principle. But I, too, am flexible.)   
cc: @raydavis

paulkerschen:
Thanks for this, @johncrossman; I make bold to declare that coexistence with the robots is possible.

PR #19, opened 2017-09-12
NOJIRA An app is an app is an app

paulkerschen:
Per comments in https://github.com/ets-berkeley-edu/boac/pull/16.


johncrossman:
Appsolutely

PR #20, opened 2017-09-13
BOAC-45, 'config' constant available in templates; module value 'me' refreshed on logOut

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-45

Eventually, `authFactory.refreshStatus` might move to an event-listener. The next PR will have CAS linkage.


raydavis:
Suspense!

PR #21, opened 2017-09-13
NOJIRA Make it easy to use fixtures outside source control

paulkerschen:
This change makes the fixtures path configurable per environment. For example, including a `FIXTURES_PATH` value in your `demo-local.py` will load fixtures from that path under `BOAC_ENV=demo`, rather than from the default path in source control. This allows us to demo from large bodies of local unobfuscated JSON if need be.

The BASE_DIR config value is also changed to a better path.

### DIFF
--- a/boac/api/config_controller.py
+++ b/boac/api/config_controller.py
@@ -13,7 +13,7 @@ def app_config():
 
 def _get_app_version():
     try:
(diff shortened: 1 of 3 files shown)


johncrossman:
Better, indeed. Thx, @paulkerschen

PR #25, opened 2017-09-14
NOJIRA Ensure mocks are unregistered on assertion failure

paulkerschen:
As it turns out, without this little fix test failures will cascade in confounding ways.


raydavis:
Makes sense, in hindsight.

PR #27, opened 2017-09-14
BOAC-47, header and footer with compact login/logout; switch to angular-ui-router

johncrossman:

Make way for students & cohorts. Screenshot of footer:
image


raydavis:
It's starting to look real...

PR #28, opened 2017-09-14
BOAC-8 @fixture decorator is your one-stop mock shop

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-8

The `@mockable` and `@mocking` decorators still exist if needed, but the `@fixture` decorator now provides the equivalent of both, taking a template string which is dynamically evaluated at request time in order to find the proper fixture filename. 

Apart from concision, what's the benefit of introducing more magic into the mix? Associating an evaluable pattern directly with the request function means that the function "knows" what its fixtures ought to look like, and can be told (in a follow-up PR) to generate them as it goes, VCR-style.

### DIFF
--- a/boac/externals/canvas.py
+++ b/boac/externals/canvas.py
@@ -1,23 +1,18 @@
 import urllib
 
 from boac.lib import http
-from boac.lib.mockingbird import fixture, mockable, mocking, paged_fixture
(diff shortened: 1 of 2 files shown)


johncrossman:
Looks good from here, @paulkerschen  Js style fix and we're good to go.

PR #29, opened 2017-09-14
BOAC-8 Allow scripts to quickly pull down a whole lotta JSON

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-8

Thanks @raydavis for working out the scriptify part of this!

Usage as shown in `scripts/example_generate_fixtures.py` is straightforward. Request functions must be decorated with `@fixture` in order to write fixture files; but then you can rest easy that the function will find the right file when it comes time to read the fixture back.

It's also possible to generate fixtures by starting the app with a `FIXTURE_OUTPUT_PATH` environment variable and clicking around; but I assume that will be a less common case for us.

### DIFF
--- a/scripts/example_generate_fixtures.py
+++ b/scripts/example_generate_fixtures.py
@@ -1,0 +1,15 @@
+import os
+# Boilerplate allowing scripts in the /scripts directory to find the boac module.
+import sys
+sys.path.append(os.path.abspath(os.path.join(os.path.dirname(__file__), '..')))
+from boac.lib import scriptify
+
+os.environ['FIXTURE_OUTPUT_PATH'] = os.path.expanduser('~/tmp')
+
+@scriptify.in_app
+def main(app):
+    from boac.externals import canvas
+    for uid in range(1, 20):
+        canvas.get_user_for_uid(app.canvas_instance, uid)
+
+main()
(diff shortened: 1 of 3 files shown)


sandeepmjay:
Very nice @paulkerschen , @raydavis ! I will wait for a second pair of eyes for merge.

scripts/example_generate_fixtures.py
```diff
@@ -0,0 +1,15 @@
...
+
+os.environ['FIXTURE_OUTPUT_PATH'] = os.path.expanduser('~/tmp')
+
+@scriptify.in_app
+def main(app):
+    from boac.externals import canvas
+    for uid in range(1, 20):
+        canvas.get_user_for_uid(app.canvas_instance, uid)
+
+main()
```
johncrossman: @paulkerschen  All looks good to me, too. Would you mind amending this PR such that 'scripts' dir is included in `lint-py`.  The py file above might need a style fix.

paulkerschen:
@johncrossman Unfortunately, including scripts in the linter has opened this can of worms: https://stackoverflow.com/questions/36827962/pep8-import-not-at-top-of-file-with-sys-path

Of the available workarounds, is there one that strikes you as least worst?

johncrossman:
@paulkerschen Ignore my suggestion is the best workaround. No linting of `scripts/`

PR #30, opened 2017-09-14
BOAC-47, draft version of '/student/:uid' view

johncrossman:

`student.html` iterates through sample data as a proof of concept. It is not an intended page design. Plus, the `me` object lives solely in `$rootScope` and is managed by authFactory only.


paulkerschen:
Looking good, @johncrossman!

PR #31, opened 2017-09-15
NOJIRA Make python scripts lintable

paulkerschen:
This seems to work - put the funny business in `scriptpath.py` and tell flake8 to ignore it.


johncrossman:
@paulkerschen You rock!!

johncrossman:
The linters linted.

PR #33, opened 2017-09-15
BOAC-51 Simple API feeds for cohort data

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-51

Since the whole backing of the cohort model is likely to change, I kept tests simple and at the controller level.

However, these are BOAC's first tests that directly depend on database-backed fixtures, and some additional changes were needed to get things working.

### DIFF
--- a/boac/api/error_handlers.py
+++ b/boac/api/error_handlers.py
@@ -1,0 +1,33 @@
+import boac.api.errors
+from flask import current_app as app, jsonify
+
+
+@app.errorhandler(boac.api.errors.BadRequestError)
+def handle_bad_request(error):
+    return error.to_json(), 400
+
+
+@app.errorhandler(boac.api.errors.UnauthorizedRequestError)
+def handle_unauthorized(error):
+    return error.to_json(), 401
+
+
+@app.errorhandler(boac.api.errors.ForbiddenRequestError)
+def handle_forbidden(error):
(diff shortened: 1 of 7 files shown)


boac/api/error_handlers.py
```diff
@@ -0,0 +1,33 @@
+import boac.api.errors
```
paulkerschen: These error handlers are moved into a separate file so that other code (such as models used in tests) can `import boac.api.errors` outside of an app context.

tests/conftest.py
```diff
@@ -66,10 +69,9 @@ def db_session(db, request):
     _session = db.create_scoped_session(options=options)
     db.session = _session
 
-    # Roll back transaction and close connection when the test is complete.
+    # Roll back transaction when the test is complete.
     def teardown():
         transaction.rollback()
-        connection.close()
```
paulkerschen: For as-yet-unknown reasons, this line (originally pulled from http://alexmic.net/flask-sqlalchemy-pytest/) worked in Snakeskin, but in BOAC causes a complaint that the connection is already closed. Perhaps it has something to do with our changed development_db structure. In any case, the tests seem to perform properly with it taken out.

raydavis:
BOAC recognizes the right to peaceful public assemblies.

PR #35, opened 2017-09-17
BOAC-47, tolerant_jsonify will not put NaN to JSON response; student view ng-repeats

johncrossman:

Standard (Flask) encoder does not produce valid JSON when given `float('nan')`.  The front-end would choke when, for example, feed included: `"zscore": NaN`.   [simplejson](https://simplejson.readthedocs.io) has a solution whereby _float('nan')_ translates to `null` in the feed.

### DIFF
--- a/.eslintrc.yml
+++ b/.eslintrc.yml
@@ -44,7 +44,7 @@ rules:
     - all
   default-case: 2
   dot-notation:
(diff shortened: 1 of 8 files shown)


raydavis:
Hey, this is starting to look real...

PR #36, opened 2017-09-18
BOAC-52, npm install tox deps only if necessary (faster linting)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-52

The `commands` section of tox.ini file does not allow for BASH-like conditional. Delegating to a script seems acceptable if we get faster linting.


paulkerschen:
Bash it, hash it, Grandmaster Flash it.

PR #38, opened 2017-09-18
NOJIRA Script to generate fixtures for CSV-stored cohorts

paulkerschen:


johncrossman:
Easy scripting? Python delivers!

PR #39, opened 2017-09-19
BOAC-56 Pad boxplot ever so slightly

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-56

Because outlier dots are important too.


johncrossman:
Every dot is sacred...

PR #42, opened 2017-09-19
BOAC-57, tolerate user with no courses in /api/user/:uid/analytics response

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-57

Review with `?w=1`.


boac/static/templates/student.html
```diff
@@ -1,47 +1,49 @@
...
+          </li>
+        </ul>
       </div>
     </div>
+    <div data-ng-if="!student.courses.length">
+      No courses
+    </div>
   </div>
   <div class="container-left" data-ng-if="!student.isLoading && error">
     <h1>Error</h1>
```
johncrossman: I removed a stray `</div>` in this file.

tests/test_api/test_user_controller.py
```diff
@@ -102,26 +103,27 @@ def test_course_with_enrollment(self, authenticated_response):
         assert course_with_enrollment['analytics']['participations']['courseDeciles'][9] == 6.0
         assert course_with_enrollment['analytics']['participations']['courseDeciles'][10] == 12.0
 
+    def test_empty_canvas_course_feed(self, client, fake_auth):
+        """returns 200 if user is found and Canvas course feed is empty"""
+        non_student_uid = '2040'
+        fake_auth.login(non_student_uid)
+        response = client.get('/api/user/{}/analytics'.format(non_student_uid))
```
paulkerschen: @johncrossman I like your having turned `api_path` into a format string above - may as well use it here too?

boac/api/user_controller.py
```diff
@@ -40,26 +40,27 @@ def user_analytics(uid):
...
+            else:
+                course_analytics['analytics'] = analytics_from_summary_feed(student_summaries, canvas_id, course)
+            course_analytics_feed.append(course_analytics)
 
-    cohort_data = Cohort.query.filter_by(member_uid=uid).first()
-    if cohort_data:
-        cohort_data = cohort_data.to_api_json()
+        cohort_data = Cohort.query.filter_by(member_uid=uid).first()
+        if cohort_data:
+            cohort_data = cohort_data.to_api_json()
```
paulkerschen: @johncrossman Looks like `cohort_data` will only get assigned if `user_courses` returns populated. Perhaps we may as well make them independent, so we can still get obfuscated name out of `cohort_data`?

boac/static/templates/student.html
```diff
@@ -1,47 +1,49 @@
...
-                     data-ng-init="drawBoxplot(course.canvasCourseId, 'participations')"></div>
-              </div>
-            </li>
-          </ul>
-        </div>
-      </div>
-      <div data-ng-if="!student.courses">
-        No courses
+    <img class="profile-summary-avatar" data-ng-src="{{student.canvasProfile.avatar_url}}" data-ng-if="student.canvasProfile.avatar_url"/>
+    <h1 data-ng-bind="student.cohortData.name || student.canvasProfile.name"></h1>
```
paulkerschen: @johncrossman This isn't clear in the code (and ought to be signalled with a better variable name or at least a comment), but `canvasProfile.name` was changed to `cohortData.name` in this template so that we wouldn't surface real names from fixtures. (I suppose the same problem exists for avatars, but in practice no one seems to populate their avatars).

paulkerschen:
Looks good @johncrossman - left a couple of questions.

tests/test_api/test_user_controller.py
johncrossman: Thanks, @paulkerschen  A refactor unfinished.

johncrossman:
@kerschen PR amended.

PR #43, opened 2017-09-20
BOAC-9 Factor course analytics out of controller

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-9

Per-user analytics fetching is moved to `boac.lib.analytics.course_analytics_for_user` to make that logic accessible from other points in the app, including scripts.


sandeepmjay:
Thanks @paulkerschen. Merging

PR #46, opened 2017-09-21
BOAC-61, upon user auth, emit authenticationSuccess event

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-61

Simple `$rootScope.$emit('authenticationSuccess')` for any controller that wants it.

### DIFF
--- a/boac/static/app/landing/landingController.js
+++ b/boac/static/app/landing/landingController.js
@@ -2,12 +2,29 @@
 
   'use strict';
 
-  angular.module('boac').controller('LandingController', function(cohortFactory, $scope) {
+  angular.module('boac').controller('LandingController', function(cohortFactory, $rootScope, $scope) {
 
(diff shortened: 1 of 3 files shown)


boac/static/app/landing/landingController.js
```diff
@@ -2,12 +2,29 @@
...
+    });
+
+    /**
+     * Refresh
+     */
+    $rootScope.$on('authenticationFailure', function() {
+      $scope.alertMessage = 'Log in failed. Please try again.';
     });
 
+    loadCohorts();
```
johncrossman: A previously known issue is that `cohortFactory.getCohorts()` is called when user is not authenticated. The js console will log an error but user experience is undisturbed. The issue highlights a larger concern: `$rootScope.me` is not available in controller due to order of events in bootstrap.  Future fix is needed.

paulkerschen:
Thanks @johncrossman, makes sense.

PR #55, opened 2017-09-29
BOAC-66 Cohort fixtures script: don't duplicate per-course API calls

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-66

When generating fixtures for a large student set, we get a lot of repeat courses.

### DIFF
--- a/scripts/cohort_fixtures.py
+++ b/scripts/cohort_fixtures.py
@@ -35,9 +35,12 @@ def main(app):
                 continue
 
             for course in user_courses:
-                student_summaries = canvas.get_student_summaries(app.canvas_instance, course['id'])
-                if not student_summaries:
-                    failures['student_summaries'].append(course['id'])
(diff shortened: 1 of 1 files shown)


scripts/cohort_fixtures.py
```diff
@@ -35,9 +35,12 @@ def main(app):
                 continue
 
             for course in user_courses:
-                student_summaries = canvas.get_student_summaries(app.canvas_instance, course['id'])
-                if not student_summaries:
-                    failures['student_summaries'].append(course['id'])
+                if os.path.isfile(f"{os.environ['FIXTURE_OUTPUT_PATH']}/canvas_student_summaries_for_course_{course['id']}.json"):
```
johncrossman: @paulkerschen do we have a stray 'f'?

paulkerschen: @johncrossman As of Python 3.6, "f" is for "fancy" "formatting".

https://docs.python.org/3/whatsnew/3.6.html#whatsnew36-pep498

johncrossman: Fab, not frivolous.

PR #57, opened 2017-10-02
BOAC-72 Only display Dev Auth login if enabled

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-72

Also added ".DS_Store" to gitignore.


johncrossman:
Mac some noise! Thx, @raydavis

PR #58, opened 2017-10-03
BOAC-73 Include Canvas term in user courses query; translate term name to SIS id

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-73

First step in the grand old game of matching official sections to bCourses sites.


raydavis:
New singer, same old song...

PR #60, opened 2017-10-04
BOAC-74 Client code, fixture, tests for SIS enrollments API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-74

We include a realistic, exhaustively obfuscated Hub API payload (verbosity, empty objects and all).


raydavis:
What took you so long, @paulkerschen ? We got API access almost an hour ago!

PR #62, opened 2017-10-05
BOAC-75 Include SIS enrollment data in user API feeds when present

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-75

It's a good old Junction-style data munge, with some extra refactoring to keep the analytics module focused on analytics rather than overall feed transformation.


raydavis:
It's a merged model, all right....

PR #64, opened 2017-10-05
BOAC-75 Crude front-end templating for enrollment and grade data

paulkerschen:
Last step in https://jira.ets.berkeley.edu/jira/browse/BOAC-75.


johncrossman:
A tenet of agile: start with crude.

PR #65, opened 2017-10-10
BOAC-59 Distinguish zero from no data in scatter plot

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-59

Points with no y-axis data are plotted below the x-axis. Points with no data at all get a separate list off the grid. It ain't the prettiest, but I don't suppose any of these dev designs are going out on the road.


johncrossman:
Null versus nil. Zero versus none. Works for me, @paulkerschen

PR #66, opened 2017-10-10
BOAC-79 Merge CalNet attributes with Cohort lists

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-79

Plus: our biggest fixture yet!


paulkerschen:
I've never before opened those portals onto how CalNet is implemented, and I'm now going to quietly close them. Thanks @raydavis.

PR #69, opened 2017-10-18
BOAC-103 Standardize external API configuration

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-103

Just simplifying some API support code before adding more complexity...

Also adds Athlete Descriptors API.

### DIFF
--- a/boac/lib/http.py
+++ b/boac/lib/http.py
@@ -1,3 +1,5 @@
+import urllib
+
 from flask import current_app as app
 from flask import Response
 import requests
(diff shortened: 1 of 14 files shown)


boac/lib/http.py
```diff
@@ -13,6 +15,20 @@ def __bool__(self):
         return False
 
 
+def build_url(url, query=None):
+    encoded_query = urllib.parse.urlencode(query, doseq=True) if query else ''
+    url_components = urllib.parse.urlparse(url)
+    print(repr(url_components))
```
paulkerschen: @raydavis Debugging line or intended log?

raydavis: That never happened & you never saw it.

paulkerschen:
Woo, emergent conventions! That Canvas instance really was getting passed around like a relay baton; thanks for giving it some rest.

PR #70, opened 2017-10-20
BOAC-97, .ebextensions and apache conf; buildspec (draft) for CodePipeline

johncrossman:

There are a few outstanding tasks:
* Proper `python3 run.py` on deploy (the `99_post_deployment_hook` is not a likely solution)
* Tweaks to buildspec
* More Apache config work

### DIFF
--- a/.ebextensions/00_ami.config
+++ b/.ebextensions/00_ami.config
@@ -1,0 +1,28 @@
+#
+# AWS configuration for BOAC
+#
+packages:
+  yum:
+    gcc: []
+    git: []
+    postgresql95-devel: []
+
+option_settings:
+  aws:elasticbeanstalk:application:
+    Application Healthcheck URL: null
+
+  aws:autoscaling:launchconfiguration:
+    InstanceType: t2.small
+    EC2KeyName: cloudlrs-aws-eb
+    IamInstanceProfile: aws-elasticbeanstalk-ec2-role
+
+  aws:elasticbeanstalk:container:python:
+    WSGIPath: run.py
+    NumProcesses: 1
+    NumThreads: 15
+
+  aws:elasticbeanstalk:container:python:staticfiles:
+    /static/: boac/static/
+
+  aws:elasticbeanstalk:application:environment:
+    PYTHONPATH: "/opt/python/current/app/boac:$PYTHONPATH"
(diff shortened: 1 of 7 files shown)


config/apache-boac.conf
```diff
@@ -0,0 +1,44 @@
...
+</Directory>
+
+<VirtualHost *:80>
+    DocumentRoot "/opt/python/current/app/boac/static"
+
+    Header set Access-Control-Allow-Origin "*"
+
+    ErrorLog /var/log/httpd/boac_error.log
+    CustomLog /var/log/httpd/boac_access_log common
+    CustomLog /var/log/httpd/boac-combined_log combined
```
paulkerschen: @johncrossman Hyphen instead of underscore intentional?

johncrossman: No. Thx, @paulkerschen  Fixing...

.ebextensions/10_install.config
```diff
@@ -0,0 +1,35 @@
+#
+# Install Python and front-end dependencies
+#
+container_commands:
+  01_install_python36:
+    command: |
+      sudo wget http://www.python.org/ftp/python/3.6.1/Python-3.6.1.tgz
```
paulkerschen: @johncrossman Why 3.6.1 specifically? buildspec.yml below has 3.6.2.

johncrossman: Good catch.

.ebextensions/00_ami.config
```diff
@@ -0,0 +1,29 @@
...
+  aws:elasticbeanstalk:container:python:
+    WSGIPath: run.py
+    NumProcesses: 1
+    NumThreads: 15
+
+  aws:elasticbeanstalk:container:python:staticfiles:
+    /static/: boac/static/
+
+  aws:elasticbeanstalk:application:environment:
+    FLASK_DEBUG: true
```
paulkerschen: @johncrossman I assume we won't want this variable set in production; do you think it will be useful while we're setting up on the cluster? If so, we'll just have to remember to change it later.

johncrossman: Removing. That is leftover from my own test deployments.

paulkerschen:
Thanks @johncrossman!

PR #73, opened 2017-10-20
BOAC-97, regex to get all files in subdir

johncrossman:

Get the regex right.


paulkerschen:
Shape it up! Get straight!

PR #74, opened 2017-10-20
BOAC-91, eb deploy step needs 'bower install'

johncrossman:

A couple more artifacts needed for deploy step.


raydavis:
Just waitin' for Mr. T...

PR #75, opened 2017-10-23
BOAC-103 Add DB-based caching

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-103
Also:
- Avoid Flask-SQLAlchemy session deadlocks in tests.
- Enable automatic transaction handling.

Known minuses:
- Because HTTP responses aren't JSON-serializable, failure information is discarded on the way up the call stack.
- Extra functions are defined to work around the lack of clean decorator nesting.


paulkerschen:
@raydavis Growing pains are to be expected, and the wrapped/uncached vs. unwrapped/cached distinction makes enough sense that I could see a convention forming around it. I think it looks good to merge once flake8 is satisfied.

PR #77, opened 2017-10-25
BOAC-103 Add ability to create Cohorts from ASC data export

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-103 (although in retrospect this really should have been broken out into its own task)

Besides loading relevant rows from an ASC-provided spreadsheet, this PR add columns to the Cohorts table for ASC's in-house team names, which differ quite a bit from the Athletes API.

### DIFF
--- a/boac/externals/asc_cohorts.py
+++ b/boac/externals/asc_cohorts.py
@@ -1,0 +1,77 @@
+import csv
+
+from boac import db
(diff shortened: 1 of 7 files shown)


paulkerschen:
Looking good, @raydavis. (Though I miss the f-strings already.)

PR #78, opened 2017-10-26
BOAC-96, eb supports python 3.4, not 3.6; Apache config for boac-dev

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-96

* no f-strings (Python 3.6), for now
* `run.py` is oddly required to use var name `application`
* Apache configs are not needed on localhost so they can be moved to `.ebextensions/`

I suggest we do https/ssl in pair-programming exercise


.ebextensions/00_ami.config
```diff
@@ -26,3 +19,4 @@ option_settings:
 
   aws:elasticbeanstalk:application:environment:
     PYTHONPATH: "/opt/python/current/app/boac:$PYTHONPATH"
+    PATH: "$HOME/.local/bin:$PATH"
```
johncrossman: `python3 get-pip.py` (below) puts pip3 in `~/.local/bin`

johncrossman:
@paulkerschen @raydavis  Two things:
1. I'd like to spend 30 min tomorrow (or Tuesday) pair-programming on the https/ssl configs. What I thought was the proper approach did not work. 
1. `.ebextensions/20_configure_environment.config` has the command below. I tested it with target filename `production-local.py` but the configs were not picked up. (Filename `dev-local.py` did not work either.) Only `development-local.py` works. What's the proper scheme here?      
```
aws s3 cp s3://boac-config/boac-${EB_ENVIRONMENT}.py config/${EB_ENVIRONMENT}-local.py
```

paulkerschen:
@johncrossman

paulkerschen:
@johncrossman What value is `EB_ENVIRONMENT` set to in the AWS console? On the suitec model, it would be set to "boac-dev", and the .ebextensions command would read:

```
aws s3 cp s3://boac-config/${EB_ENVIRONMENT}.py config/production-local.py
```

In order to pick up that config, the environment variable `BOAC_ENV` should be set to "production",  along the lines of what SuiteC does with `NODE_ENV` here:
https://github.com/ets-berkeley-edu/suitec/blob/master/.ebextensions/03_set_node_environment.config

johncrossman:
@paulkerschen `EB_ENVIRONMENT` is set to 'dev'. I'll change it (to 'boac-dev') per suitec convention. PR amend on the way.

.ebextensions/02_set_boac_environment.config
```diff
@@ -0,0 +1,8 @@
+#
+# BOAC expects certain environment variables
+#
+
+# BOAC_ENV is set to production regardless of environment. EB_ENVIRONMENT has current Elastic Beanstalk environment name.
+option_settings:
+  - option_name: 'BOAC_ENV'
+    value: 'production'
```
johncrossman: @paulkerschen @raydavis  BOAC_ENV above. Below you'll see that 10_install.config was split in two (pip3_install and bower_install).

paulkerschen:
Thanks for the extra retroPython work, @johncrossman.

PR #82, opened 2017-10-27
BOAC-96 Elastic Beanstalk load balancer and SSL configurations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-96

To @johncrossman's extensive SSL work I've added a couple of commits restoring deploy commands that got lost in #78.


sandeepmjay:
Looks good @paulkerschen. Merging now.

PR #83, opened 2017-10-27
BOAC-96 Further .ebextensions fixes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-96, and most especially https://jira.ets.berkeley.edu/jira/browse/BOAC-107.

Hard-won discoveries:

- The Python EB environment looks for a requirements.txt file and sics pip on it by default; having an explicit pip command in .ebextensions is duplicative. An explicit command is necessary to bring pip up to a current version.
- An Apache restart isn't necessary because our new configs get added sufficiently early, in the "files" phase; and said Apache restart turns out to be the thing that was preventing Python code from seeing the BOAC_ENV variable. Fortunately the same thing happened to one other person, who blogged about it: https://blog.writelab.com/blog/2015/5/20/django-elastic-beanstalk-apache-config-and-environmental-variables.


raydavis:
Thank you for going through this instead of me!

PR #87, opened 2017-10-30
BOAC-111 WSGI needs secret sauce or numpy will destroy it

paulkerschen:
As seen on https://jira.ets.berkeley.edu/jira/browse/BOAC-111.


paulkerschen:
Locally tox'd.

johncrossman:
What a difference a config makes. Thanks, @paulkerschen

PR #88, opened 2017-10-31
BOAC-108 Effective logging under Elastic Beanstalk/WSGI

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-108

### DIFF
--- a/boac/models/json_cache.py
+++ b/boac/models/json_cache.py
@@ -50,11 +50,13 @@ def stow(key_pattern, for_term=False):
             )
         stowed = JsonCache.query.filter_by(key=key).first()
         if stowed:
+            app.logger.debug('Returning stowed JSON for key {key}'.format(key=key))
             return stowed.json
         else:
             app.logger.info('{key} not found in DB'.format(key=key))
             to_stow = func(*args, **kw)
             if to_stow is not None:
+                app.logger.debug('Will stow JSON for key {key}'.format(key=key))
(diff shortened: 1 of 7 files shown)


boac/models/json_cache.py
```diff
@@ -50,11 +50,13 @@ def _stow(func, *args, **kw):
             )
         stowed = JsonCache.query.filter_by(key=key).first()
         if stowed:
+            app.logger.debug('Returning stowed JSON for key {key}'.format(key=key))
```
paulkerschen: This line in particular gets very busy when loading something like a cohort view, which relies on a whole bunch of stowed JSON. But this may be appropriate for debug mode, and while we're still getting to know the app, I find exhaustive description of its activities more reassuring than several seconds of silence.

boac/routes.py
```diff
@@ -33,3 +33,14 @@ def handle_unmatched_api_route(**kwargs):
     @app.route('/<path:path>')
     def front_end_route(**kwargs):
         return make_response(open('boac/templates/index.html').read())
+
+    @app.after_request
+    def log_api_requests(response):
```
paulkerschen: The access logs that we're used to seeing when running BOAC locally come not from Flask, but from the Werkzeug development server, which isn't active under WSGI. For endpoint calls to show up in the logs, the app needs to log them explicitly.

johncrossman:
Effective indeed, @paulkerschen

PR #91, opened 2017-11-01
BOAC-114 Allow for weird suffixes in SIS section ids

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-114

Observed in the wild in PHYSICS 10.


johncrossman:
We like weird.

PR #92, opened 2017-11-02
BOAC-90 List/matrix toggle for cohort view

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-90

Following in-progress designs.

screen shot 2017-11-02 at 12 16 31 pm
screen shot 2017-11-02 at 12 13 51 pm

### DIFF
--- a/boac/static/app/cohort/cohort.html
+++ b/boac/static/app/cohort/cohort.html
@@ -1,35 +1,50 @@
 <div data-ng-include="'/static/app/shared/header.html'"></div>
 
 <div class="container">
-  <h1 data-ng-bind="cohort.name"></h1>
-
-  <h2>Student Performance Assessment Graph</h2>
   <div class="container" data-ng-if="errored">
     There was an error retrieving cohort data.
   </div>
   <div class="container" data-ng-if="!errored && !cohort">
     Loading....
   </div>
-  <div id="scatterplot"></div>
-  <div class="container" data-ng-if="!errored && cohort">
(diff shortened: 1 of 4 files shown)


boac/static/app/cohort/cohort.html
```diff
@@ -1,36 +1,51 @@
 <div data-ng-include="'/static/app/shared/header.html'"></div>
 
 <div class="container">
-  <h1 data-ng-bind="cohort.name"></h1>
```
paulkerschen: Per designs, team name no longer appears at top, since a cohort can be specified in numerous other ways. Current designs have a more detailed cohort selection bar (not yet implemented) that shows how the current cohort is defined.

boac/static/app/cohort/cohort.html
```diff
@@ -1,36 +1,51 @@
...
-        <a data-ng-href="/student/{{member.uid}}">
-          <img class="cohort-avatar" data-ng-src="{{member.avatar_url}}">
-          <span data-ng-bind="member.name"></span>
-        </a>
-      </li>
-    </ul>
-
-    <div class="body-text">
-      <h2 data-ng-if="cohort">Students by name</h2>
+  <div class="container" data-ng-show="!errored && cohort">
```
paulkerschen: Here and below, we use `ng-show` rather than `ng-if` because the `scatterplot` element needs to be present in the DOM for d3 to do its work.

johncrossman:
Houston, we have toggle.

PR #95, opened 2017-11-08
NOJIRA, upgrade flake8-import-order per Travis; two lint-py fixes

johncrossman:
Cause of mysterious Travis failure (https://travis-ci.org/ets-berkeley-edu/boac/builds/298832425) was unexpected `flake8-import-order` version. Devs will get latest with this PR.


paulkerschen:
Naturally!

PR #94, opened 2017-11-07
BOAC-92, rename 'cohorts' to 'teams' (w/ db modify script) - make way for custom cohorts 

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-92

See JIRA comments.

Refactor will continue as we develop search and custom-cohorts.

PR #96, opened 2017-11-08
BOAC-92, rename 'cohorts' to 'team_members' (w/ db modify script)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-92

First try was PR #94 

db modify script is `scripts/db/migrate/20171107-BOAC-92/rename_cohorts_table.sql`


paulkerschen:
The new fall style is in!

PR #97, opened 2017-11-09
BOAC-92, cohort_filters table, many-to-many with authorized_users

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-92

We don't need to roll out many-to-many in BOAC UX and yet having such support on the backend gives us the option to manually create shared filters (e.g., low_gpa_filter). Plus, we avoid a future refactor.

**Note:** I'll amend this with a few more tests if we like this direction.

### DIFF
--- a/boac/models/development_db.py
+++ b/boac/models/development_db.py
@@ -1,9 +1,10 @@
 import csv
 from boac import db
 from boac.models.authorized_user import AuthorizedUser
+from boac.models.authorized_user import CohortFilter
 # Needed for db.create_all to find the model.
-from boac.models.team_member import TeamMember # noqa
 from boac.models.json_cache import JsonCache # noqa
+from boac.models.team_member import TeamMember # noqa
 
 
 def clear():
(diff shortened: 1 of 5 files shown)


boac/models/development_db.py
```diff
@@ -38,12 +39,24 @@ def load_schemas():
...
 def load_development_data():
     csv_reader = csv.DictReader(_default_users_csv.splitlines())
+    cohort_filter_crew = create_cohort_filter('Men and women\'s crew', 'CRM', 'CRW')
+    cohort_filter_soccer = create_cohort_filter('Men and women\'s soccer', 'SCW', 'SCW')
     for row in csv_reader:
-        obj = AuthorizedUser(**row)
-        db.session.add(obj)
+        user = AuthorizedUser(**row)
+        db.session.add(user)
+        cohort = cohort_filter_crew if int(user.uid) < 100000 else cohort_filter_soccer
```
johncrossman: This `uid < 100000` condition is my (odd) way of creating two cohort_filters, each shared by multiple users.

johncrossman:
@paulkerschen Amendments are in.

paulkerschen:
Thanks, @johncrossman, looking good.

PR #98, opened 2017-11-09
BOAC-93, BOAC/api to support GET custom cohorts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-93

All teams are cohorts but not all cohorts are teams. 
* I generalized `/api/cohort/<cohort_code>`
* the custom cohorts are always available in `current_user` object
* next step is to rework the front-end


paulkerschen:
Mixable, matchable, shareable cohorts! Looks good to me @johncrossman.

PR #100, opened 2017-11-14
BOAC-94, prep page header for cohort search

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-94

* Page header with search/custom-cohort placeholders
* User profile added to `me`
* search page with initial controller


raydavis:
It lives!

PR #103, opened 2017-11-14
BOAC-140, if not state.isPublic then force location.path to '/'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-140

None (without auth) shall pass!


raydavis:
Nice & simple scheme

PR #110, opened 2017-11-20
BOAC-95, cohort counts on landing; wrap-reverse puts myCohorts on top

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-95

* `myCohorts` is not needed in `/api/profile`
* wrap-reverse on landing puts my-cohorts on top
* isLoading for dropdown in header nav


paulkerschen:
Me and my cohorts, strolling down the avenue...

PR #111, opened 2017-11-20
BOAC-148, fix db.Table def of cohort_filter_owners

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-148

This matches the intention, as seen in [scripts/db/migrate/20171107-BOAC-92/create_cohort_filters_table.sql](https://github.com/ets-berkeley-edu/boac/blob/master/scripts/db/migrate/20171107-BOAC-92/create_cohort_filters_table.sql).

On localhost, a stale schema (in which `cohort_filter_owners.user_id` was varchar) was masking the problem.  Once I refreshed my local schema I could reproduce (then fix) the bug.


paulkerschen:
The local schema was quiet, too quiet...

PR #112, opened 2017-11-20
BOAC-149, when get_cohort_list, include member_count per cohort and exclude Canvas user profile

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-149

Furthermore, do not cache team data if load_canvas_profile=False. (The change in `development_db.py` allows use to load test-cohort-data when test-user-data is already present in the db – e.g., boac-dev)

### DIFF
--- a/boac/models/development_db.py
+++ b/boac/models/development_db.py
@@ -41,9 +41,13 @@ def load_development_data():
     csv_reader = csv.DictReader(_default_users_csv.splitlines())
     team_codes = list(TeamMember.team_definitions.keys())
     for row in csv_reader:
-        user = AuthorizedUser(**row)
-        db.session.add(user)
-        # A subset of users get saved cohorts
+        # This script can be run more than once. Do not create user if s/he exists in BOAC db.
+        user = AuthorizedUser.find_by_uid(row['uid'])
+        if not user:
+            user = AuthorizedUser(**row)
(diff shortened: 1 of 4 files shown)


boac/static/app/cohort/manageCohortsController.js
```diff
@@ -37,8 +37,6 @@
     $scope.deleteCohort = cohortFactory.deleteCohort;
 
     var init = function() {
-      $scope.isLoading = true;
-
```
johncrossman: We only need `isLoading = true` on initial page render. It is not needed when page refreshes due to cohort renaming or deletion.

boac/models/development_db.py
```diff
@@ -41,9 +41,13 @@ def load_development_data():
     csv_reader = csv.DictReader(_default_users_csv.splitlines())
     team_codes = list(TeamMember.team_definitions.keys())
     for row in csv_reader:
-        user = AuthorizedUser(**row)
-        db.session.add(user)
-        # A subset of users get saved cohorts
+        # This script can be run more than once. Do not create user is s/he exists in BOAC db.
```
paulkerschen: @johncrossman: `is`/`if` comment typo

johncrossman: Thanks. Corrected.

PR #114, opened 2017-11-23
BOAC-153, dev_auth need only support json POST

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-153

Why support `request.form` if we don't need it?


raydavis:
I let it go GET in case it would be useful for scripting, but if @pfarestveit doesn't need it, I don't either.

PR #117, opened 2017-11-28
BOAC-94, remove obsolete searchController.js and related route

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-94

`cohortController.js` does the work of view/create cohort and search. PR includes postgres version bump in `.travis.yml`.


johncrossman:
Fyi, we're using postgres 9.6 in AWS land:
image

PR #118, opened 2017-11-28
BOAC-156, cohort view must have totalMemberCount for all cohorts, incl. teams

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-156

The importance of `totalMemberCount` on cohort view:
* if equals 0 or more then you're viewing search results or saved cohort
* if null then you're starting from scratch (ie, 'create a cohort')


paulkerschen:
How many were going to Saint Ives?

PR #119, opened 2017-11-28
BOAC-151 Back-end percentile rounding to one decimal place

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-151

In some cases the front end will decide to round further, but there's no need to pass all those floating-point digits into the API.


raydavis:
The Travis issue is unrelated to @paulkerschen 's changes & a local run of ``tox`` succeeds.

PR #120, opened 2017-11-28
BOAC-158, use data-ng-show so #scatterplot is reachable by d3

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-158


raydavis:
Oh, all right...

PR #121, opened 2017-11-28
BOAC-157 Fix db session handling in tests

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-157

The particular breakage here addressed was caused by upgrading pytest from to 3.2.5 to 3.3.0. It may have something to do with changes to the `pytest_fixture_post_finalizer` hook, but that's getting pretty deep: https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23

Note that explicit use of transactions seems not to be necessary, as nowadays an SQLAlchemy session behaves transactionally by default.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -61,24 +61,23 @@ def db(app, request):
 def db_session(db, request):
     """
     Fixture database session used for the scope of a single test. All executions are wrapped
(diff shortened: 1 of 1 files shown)


johncrossman:
Skillful sleuthing, @paulkerschen

PR #124, opened 2017-11-30
BOAC-119, UX for cohort/search filters and create modal

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-119

UX: https://confluence.ets.berkeley.edu/confluence/display/LRSASC/Cohort+View+Page

Note: pytest-catchlog has been merged to pytest.

PR #125, opened 2017-11-30
BOAC-119, play nice when isCreateCohortMode becomes save search mode

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-119

Follow up (fixes) on PR #124


raydavis:
Cohort, behave
That's what they say when we're together...

PR #126, opened 2017-11-30
BOAC-147 Begin collecting Canvas grade data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-147

More data is good data!


johncrossman:
Mo' betta data.

PR #127, opened 2017-11-30
BOAC-121 BOAC-127 Show enrollments from all terms with best-effort Canvas site association

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-121
https://jira.ets.berkeley.edu/jira/browse/BOAC-127

Apologies for the large PR, but these two wanted to come together.

A few UX topics:
- Unmatched Canvas sites from a given term appear immediately after courses for that term. This means that prior to Fall 2016, we show a bunch of courses with no sites and then a bunch of sites with no courses. Suppress some or all of this?
- Our current enrollments-fetching code only looks at primary sections, so Canvas sites associated with secondary sections only will remain unmatched.
- Phys Ed courses taken by athletes now appear in the mix. Suppress these, even if (as occasionally happens) there's a Canvas site associated? Should any other enrollments be considered noise? 
- Dropped enrollments are included back to the beginning of time. Probably we should suppress these for prior terms? For the current term, are they more useful?
- This list of sites plus courses-with-whiskers is still a far cry from the last elaborated design for the student page. How to reconcile?

### DIFF
--- a/boac/lib/berkeley.py
+++ b/boac/lib/berkeley.py
@@ -14,3 +14,14 @@ def sis_term_id_for_name(term_name=None):
                 'Fall': '8',
             }
             return '2' + match.group(2) + season_codes[match.group(1)]
+
+
+def term_name_for_sis_id(sis_id=None):
+    if sis_id:
+        sis_id = str(sis_id)
+        season_codes = {
+            '2': 'Spring',
+            '5': 'Summer',
+            '8': 'Fall',
+        }
+        return season_codes[sis_id[3:4]] + ' 20' + sis_id[1:3]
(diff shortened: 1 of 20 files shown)


boac/lib/berkeley.py
```diff
@@ -14,3 +14,14 @@ def sis_term_id_for_name(term_name=None):
...
+
+def term_name_for_sis_id(sis_id=None):
+    if sis_id:
+        sis_id = str(sis_id)
+        season_codes = {
+            '2': 'Spring',
+            '5': 'Summer',
+            '8': 'Fall',
+        }
+        return season_codes[sis_id[3:4]] + ' 20' + sis_id[1:3]
```
johncrossman: There it is. The screwy scheme we can't escape.

johncrossman:
Quite good for a "best-effort". Thx, @paulkerschen

PR #129, opened 2017-11-30
BOAC-151 Round all percentile figures to whole numbers

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-151


johncrossman:
Whole numbers are wholesome.

PR #130, opened 2017-11-30
BOAC-169, track user/browser events with Google Analytics

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-169

More info: https://developers.google.com/analytics


paulkerschen:
And the big data itself, in turn, has bigger data to go on....

PR #131, opened 2017-11-30
BOAC-177 Stop trying to calculate NaN stats on unvaried datasets

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-177


paulkerschen:
Calling yourself a number doesn't make you a number, kid.

PR #135, opened 2017-12-01
[python36] BOAC-193, bump python-path (in 01_create_apache_conf.config) to python3.6

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-193


paulkerschen:
What a difference a number makes. Waiting for Travis...

PR #136, opened 2017-12-01
BOAC-192 Reduce contention on DB sessions caused by lengthy API calls

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-192


paulkerschen:
woodsy-owl-original

PR #138, opened 2017-12-04
BOAC-147 Bring Canvas scores APIs in line with BOAC-127 terms handling

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-147

For a while on Thursday, these expensive feeds were being cached with a "term_None" prefix.


johncrossman:
Thanks for the test coverage.

PR #141, opened 2017-12-05
BOAC-159 BOAC-171 BOAC-195 Secondary section handling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-159
https://jira.ets.berkeley.edu/jira/browse/BOAC-171

Since this change will result in more courses with multiple sites attached, per-site headings are added to address https://jira.ets.berkeley.edu/jira/browse/BOAC-195.

It appears that the SIS enrollments API returns secondary sections only back to Fall 2016.

While the page structure will change following deployment of this PR, the underlying data will not show on boac-dev until the enrollments API cache is refreshed.

### DIFF
--- a/boac/lib/merged.py
+++ b/boac/lib/merged.py
@@ -34,10 +34,25 @@ def merge_sis_enrollments(canvas_course_sites, cs_id, matriculation):
 def merge_sis_enrollments_for_term(canvas_course_sites, cs_id, term_name):
     term_id = sis_term_id_for_name(term_name)
     enrollments = sis_enrollments_api.get_enrollments(cs_id, term_id)
+
     if enrollments:
-        enrollments_feed = [api_util.sis_enrollment_api_feed(enrollment) for enrollment in enrollments.get('studentEnrollments', [])]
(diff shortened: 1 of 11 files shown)


boac/lib/merged.py
```diff
@@ -34,10 +34,25 @@ def reverse_terms_until(stop_term):
...
+        for enrollment in enrollments.get('studentEnrollments', []):
+            # Skip this class section if we've seen it already.
+            section_id = enrollment.get('classSection').get('id')
+            if section_id in term_section_ids:
+                continue
+            else:
+                term_section_ids[section_id] = True
+            # SIS class id (as distinct from section id or course id) is not surfaced by the SIS enrollments API. Our best
+            # unique identifier is the class display name.
+            class_name = enrollment.get('classSection', {}).get('class', {}).get('displayName')
```
raydavis: Good call, @paulkerschen -- this level of disambiguation wasn't enough for CalCentral, but it should work in BOAC's context. (Especially since we're already weeding out the most common sort of duplicate class display names.)

PR #142, opened 2017-12-05
BOAC-173 Suppress dropped enrollments from past terms

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-173


johncrossman:
Out of term, out of mind.

PR #144, opened 2017-12-05
BOAC-194 Hush, JavaScript error

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-194


johncrossman:
Sleepin' with the fishes.

PR #147, opened 2017-12-06
BOAC-208 Restore per-term filter to Canvas cache reload

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-208

We lost this filter from boac.canvas.externals in #127.


johncrossman:
And we're back. Thanks, @paulkerschen

PR #149, opened 2017-12-06
BOAC-207 Include majors of all sorts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-207

Compare https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/my_academics/college_and_level.rb#L255


johncrossman:
Ah, the Ruby cookbook.

PR #153, opened 2017-12-06
BOAC-204 Add limited support for Canvas score/grade loading

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-204

Cain't show what ain't thar.

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -39,6 +39,45 @@ def load_canvas_externals(uid, sis_term_id):
     return success_count, failures
 
 
+def load_canvas_scores(sis_term_id=None):
+    from boac.externals import canvas
+    from boac.lib import berkeley
+    from boac.models.team_member import TeamMember
+
+    success_count = 0
+    failures = []
+
+    if sis_term_id is None:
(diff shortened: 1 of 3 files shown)


boac/api/cache_utils.py
```diff
@@ -39,6 +39,52 @@ def load_canvas_externals(uid, sis_term_id):
...
+                        ))
+                        continue
+                    success_count += 1
+                    # if not canvas.get_assignments_analytics(site_id, uid, sis_term_id):
+                    #     failures.append('canvas.get_assignments_analytics failed for UID {}, site_id {}'.format(
+                    #         uid,
+                    #         site_id,
+                    #     ))
+                    #     continue
+                    # success_count += 1
```
johncrossman: @raydavis The lines above can go?

johncrossman:
Looks good, @raydavis  One question posted.

boac/api/cache_utils.py
raydavis: Thanks, @johncrossman . I left them there by way of a reminder, but you're right that it doesn't match our standard practice. I'll amend.

PR #155, opened 2017-12-07
BOAC-185 BOAC-190 Add SIS profile and current-term courses to cohort

paulkerschen:
You said the cohort members API feed wasn't slow or bulky enough... and we listened!

Back end only for:
https://jira.ets.berkeley.edu/jira/browse/BOAC-185
https://jira.ets.berkeley.edu/jira/browse/BOAC-190

Querying and filtering on level and GPA, as opposed to just surfacing them, probably calls for additional work to extract those values into new database columns, unless we want to try the wild world of JSONB.

### DIFF
--- a/boac/models/team_member.py
+++ b/boac/models/team_member.py
@@ -4,6 +4,8 @@ from boac import db
 from boac.api.util import canvas_courses_api_feed
 from boac.externals import canvas
 from boac.lib.analytics import mean_course_analytics_for_user
+from boac.merged.sis_enrollments import merge_sis_enrollments_for_term
+from boac.merged.sis_profile import merge_sis_profile
 from boac.models.base import Base
 from flask import current_app as app
 from sqlalchemy import func, UniqueConstraint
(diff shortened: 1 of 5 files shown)


boac/models/team_member.py
```diff
@@ -164,37 +166,14 @@ def team_group_summary(row):
...
-                            canvas_courses = canvas_courses_api_feed(student_courses_in_current_term)
-                            if canvas_courses:
-                                member['analytics'] = mean_course_analytics_for_user(canvas_courses,
-                                                                                     canvas_profile['id'],
-                                                                                     current_term)
+                member = row.to_api_json(details=include_member_details)
                 summary['members'].append(member)
 
             summary['totalMemberCount'] = TeamMember.query.distinct(o).filter(f).count()
-            db.session.commit()
```
paulkerschen: Turns out this line in a read-only method was upsetting pytest by persisting json_cache entries between tests.

raydavis:
If you want slow-n-bulky (& who doesn't?), just wait for those Canvas scores to be merged!

tests/conftest.py
```diff
@@ -66,6 +66,8 @@ def db_session(db, request):
     # Mixing SQL-using test fixtures with SQL-using decorators seems to cause timing issues with pytest's
     # fixture finalizers. Instead of using a finalizer to roll back the session and close connections,
     # we begin by cleaning up any previous invocations.
+    # This fixture is marked 'autouse' to ensure that cleanup happens at the start of every test, whether
```
raydavis: Thanks for the explanation, @paulkerschen

PR #160, opened 2017-12-08
BOAC-188, order by firstName/lastName requires code and schema change

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-188

Notice SQL scripts in `scripts/db/migrate/20171207-BOAC-188`. The plan is:
* add columns to team_members table: firstName, lastName (no null constraint)
* populate columns based on `member_name`
* add 'not null' constraint and then drop member_name column

Of course, the code changes depend on the db-modify.  End result is sorting by first and last name in search results and more natural display of names in UI.


boac/externals/asc_cohorts.py
```diff
@@ -55,7 +55,8 @@ def load_cohort_from_csv(app, csv_file='tmp/FilteredAscStudents.csv'):
             sis_sport_code = SPORT_TRANSLATIONS[asc_sport_code_core]
             record = TeamMember(
                 member_csid=r['SID'],
-                member_name=r['cName'],
+                first_name=r['FirstName'],
+                last_name=r['LastName'],
```
johncrossman: `FilteredAscStudents.csv` will need a change but where is it kept?!  Didn't find it in S3://pdg-share

raydavis: I believe I just extracted it from the Excel file, @johncrossman, after eliminating non-students, etc. I'll go ahead and upload it there.

PR #163, opened 2017-12-08
BOAC-226 Show mid-term grades if available

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-226

It's not fire-engine red and blinking, but it's there.


paulkerschen:
Risk! Risk!

PR #169, opened 2017-12-11
BOAC-215 Improve cohort view performance with in-memory cache

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-215

First time through with a big team can still be painful, but at least it gets nicer afterwards.


paulkerschen:
Sorely needed, and thank you, @raydavis!

PR #171, opened 2017-12-12
BOAC-204 Merge Canvas site "current score" stats into cohort-wide means

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-204

Currently only shows on the scatterplot matrix via "?yAxis=analytics.courseCurrentScore".
Changed display name to "Assignment grades" to avoid confusion with other "score" labels.

### DIFF
--- a/boac/static/app/cohort/cohortController.js
+++ b/boac/static/app/cohort/cohortController.js
@@ -8,6 +8,7 @@
     cohortFactory,
     cohortService,
     googleAnalyticsService,
+    $location,
     $rootScope,
     $scope,
     $state,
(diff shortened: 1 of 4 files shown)


boac/static/app/cohort/cohortController.js
```diff
@@ -11,7 +11,8 @@
     $rootScope,
     $scope,
     $state,
-    $stateParams
+    $stateParams,
+    $location
```
paulkerschen: @raydavis Not that it much matters, but we've been trying to keep these injected dependencies alphabetical.

raydavis: Which is a lot clearer now that it's not a super long line. I'll amend.

paulkerschen:
In the bucket!

PR #174, opened 2017-12-12
BOAC-224 Fix KeyError when matriculation not present

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-224

Durn idioms.


raydavis:
Shouldn't that be "Danged"?

PR #177, opened 2017-12-12
BOAC-222 Sort sections per enrollment

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-222


raydavis:
There's something oddly familiar about all this....

PR #183, opened 2017-12-13
BOAC-241 Cohort list view: change "Activity" to "Page Views"

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-241

Not the meatiest burger, this one.


raydavis:
But at least the bun is labeled correctly.

PR #184, opened 2017-12-13
BOAC-116 Disallow cohort matrix CSS from romping all over the app

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-116

Fun fact: the letter "I" was disappearing from the Berkeley logo because it's defined as an SVG `rect`, whereas all other letters are `path` or `polygon`.


johncrossman:
There is no "i" in Team BOAC.  So...

PR #198, opened 2017-12-18
BOAC-260 BOAC-262 Student profile: reformat class listings

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-260
https://jira.ets.berkeley.edu/jira/browse/BOAC-262

Designs: https://jira.ets.berkeley.edu/jira/browse/BOAC-257

Rebased version of #195.

PR #205, opened 2017-12-19
BOAC-262 Tame double border separator

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-262 (following up #198 and #199)


raydavis:
Lose those whiskers, mister.

PR #207, opened 2017-12-19
BOAC-184, additional cohort filter criteria supported in cohort.create()

johncrossman:
UX: https://jira.ets.berkeley.edu/jira/browse/BOAC-12

What remains:
* expand associated db query to support new params: gpa_ranges, levels, majors, unit_ranges_eligibility, unit_ranges_pacing
* front-end work (dropdowns on cohort view)


boac/lib/util.py
```diff
@@ -0,0 +1,6 @@
+"""Generic utilities"""
+
+
+def get(_dict, key, default_value=None):
+    value = _dict and key in _dict and _dict[key]
+    return value or default_value
```
paulkerschen: @johncrossman It's our own baby Lodash...

tests/test_models/test_cohort_filter.py
```diff
@@ -13,11 +15,46 @@ def test_no_cohort(self):
...
+        cf = cohort['filterCriteria']
+        assert expected == cf
+        # Intended use of GPA and unit ranges
+        gpa_range = cf['gpaRanges'][0]
+        assert 1.99 in numpy.arange(*gpa_range)
+        assert 2.5 not in numpy.arange(*gpa_range)
+        unit_range = cf['unitRangesEligibility'][0]
+        assert 59 not in numpy.arange(*unit_range)
+        assert 60 not in numpy.arange(*unit_range)
+        assert 5000 in numpy.arange(*cf['unitRangesPacing'][1])
```
paulkerschen: @johncrossman I had to go check the numpy.arange documentation to get clear that this was a `[start, stop, step]` format. If that's the format we're going to use, it might be worth documenting in the controller or elsewhere. But are we sure that a `step` parameter is called for here? When it comes to translating these ranges into SQL, I'd think we would be more likely to use a filter like `gpa > 2.5 AND gpa <= 4`, rather than expanding the range into an array of all possible GPA values and testing for membership.

paulkerschen:
Looks good, @johncrossman, and thanks for all the test examples. I had one question about how we're specifying range criteria, which naturally depends on how the db will implement them.

tests/test_models/test_cohort_filter.py
johncrossman: @paulkerschen Yes, a good point on "translation into SQL".  Perhaps a non-numpy convention...?
```
gpaRange (2.5,4] => gpa > 2.5 AND gpa <= 4
unitRangesPacing [30,∞) => 30 units and above
```

paulkerschen: @johncrossman That makes sense. Perhaps we should expect the API param to be a text representation of the Postgres `numrange` type?

https://www.postgresql.org/docs/9.6/static/rangetypes.html

And then if we're constructing raw queries, the syntax is apparently `gpa <@ gparange`? https://www.postgresql.org/docs/9.6/static/functions-range.html
Don't know if SQLAlchemy is any help here or should be circumvented.

johncrossman: @paulkerschen We're zeroing in on something sensible. Are we agreeing that the string representation, in both API and `cohort_filters::filter_criteria` JSONB column, would be `numrange(1.0, 14.0, '(]')`?  Including the "numrange" prefix makes it self-documenting. (The SQLAlchemy question can be tackled later, methinks.)

paulkerschen: I approve, @johncrossman!

johncrossman: Amendment coming soon...

johncrossman: Fyi, SQLAlchemy has [numrange support](http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html?highlight=numrange#sqlalchemy.dialects.postgresql.NUMRANGE). cc: @paulkerschen

tests/test_api/test_cohort_controller.py
```diff
@@ -120,14 +120,40 @@ def test_create_cohort(self, authenticated_session, client):
...
 
+    def test_create_cohort_with_complex_filters(self, authenticated_session, client):
+        """creates custom cohort, with many non-empty filter_criteria"""
+        label = 'Complex'
+        gpa_ranges = [[0.0, 2.00, 0.01], [2.00, 2.49, 0.01]]
+        group_codes = []
+        levels = ['Junior']
+        majors = ['Environmental Economics & Policy', 'Gender and Women\’s Studies']
+        unit_ranges_eligibility = [[30, 59, 1]]
+        unit_ranges_pacing = [[0, 5, 1], [30, 1000000]]
```
johncrossman: I'll have this amended very soon.  I'm adding a test on 'numrange' validation

tests/test_api/test_cohort_controller.py
```diff
@@ -120,14 +120,53 @@ def test_create_cohort(self, authenticated_session, client):
...
+        bad_range_syntax = 'numrange(2, BLARGH, \'[)\')'
+        custom_cohort = {
+            'label': 'Problematic Cohort',
+            'gpaRanges': [bad_range_syntax],
+            'levels': ['Sophomore'],
+        }
+        response = client.post('/api/cohort/create', data=json.dumps(custom_cohort), content_type='application/json')
+        assert 500 == response.status_code
+        assert 'BLARGH' in str(response.data)
+        assert 'does not match expected' in str(response.data)
```
johncrossman: The new test is in and Travis is green.

PR #211, opened 2017-12-20
BOAC-184, validate all filter_criteria before /cohort/create

johncrossman:

Let's do our best to avoid bad `filter_criteria` data.


raydavis:
Let's!

PR #212, opened 2017-12-20
BOAC-247 BOAC-273 Show all team names in student profile and search dropdown

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-247
https://jira.ets.berkeley.edu/jira/browse/BOAC-273

Note that surfacing multiple team memberships requires changing the structure of the /api/students/all feed, which in turn requires the search dropdown directive to do more work to parse multiple memberships. 

Per the design suggestion in https://jira.ets.berkeley.edu/jira/browse/BOAC-247, students on multiple teams will appear multiple times in the dropdown. If we decide we don't like this behavior, it's easy to change; the important thing is that we've decoupled this logic from the API feed structure and located it entirely in the front end.

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -68,7 +68,7 @@ def user_analytics(uid):
     student = Student.query.filter_by(uid=uid).first()
     if student:
         sis_profile = merge_sis_profile(student.sid)
-        athletics_profile = student.to_api_json()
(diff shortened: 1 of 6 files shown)


johncrossman:
One feed to rule them all. Thx, @paulkerschen

PR #213, opened 2017-12-20
BOAC-138 Add admin API endpoints to manage refresh jobs

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-138

This PR knowingly makes an utter hash of the cache_utils module. The goal is to support a non-destructive test of this background job approach on AWS. Once we have something that seems OK, I'll tidy the place up.


johncrossman:
Nice 'n tidy. Thanks, @raydavis

PR #214, opened 2017-12-20
BOAC-242 Show name, not UID, in the header

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-242


raydavis:
You can call me Ray, or you can call me Unnamed, but please don't call me 211159.

PR #218, opened 2017-12-21
BOAC-184, cohort view gets two more filters: majors and gpaRanges

johncrossman:
UX: https://jira.ets.berkeley.edu/jira/browse/BOAC-12

* GPA options have `numrange` syntax
* the Majors dropdown (usefulness is TBD) has options hard-coded in dedicated `majorsFactory.js`.  (Values copied from CalCentral's [app/models/berkeley/majors.rb](https://github.com/ets-berkeley-edu/calcentral/blob/master/app/models/berkeley/majors.rb).) If we keep this dropdown we'll probably want to strip it down a bit.


boac/static/app/student/majorsFactory.js
```diff
@@ -0,0 +1,226 @@
...
+  'use strict';
+
+  var boac = angular.module('boac');
+
+  boac.factory('majorsFactory', function() {
+
+    var getAllMajors = function() {
+      return [
+        {name: 'Undeclared'},
+        {name: 'Undeclared - Engineering'},
```
johncrossman: I put 'Undeclared' options on top. I assume they'll be most interesting to ASC advisors.

raydavis:
Yes, that Majors dropdown is more of a skydive, isn't it? Well, let's run it up the tallest flagpole in the world and see who salutes!

PR #219, opened 2017-12-21
BOAC-268 Clearly distinguish midpoint grade from final grade

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-268

Grading basis translations are the same as those used by CalCentral.


johncrossman:
It's clear to me. Thx, @paulkerschen

PR #221, opened 2017-12-21
BOAC-282 Do not lose Canvas data after a course site has been concluded

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-282

Without this, we run a high risk of 500 errors due to enrollment list mismatches between Canvas APIs, as well as losing important analytics.


raydavis:
The bug unfortunately means we'll have to re-load all cached Canvas data feeds.

paulkerschen:
Ugh!

PR #222, opened 2017-12-21
BOAC-278 Normalized database cache for SIS profile

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-278

Before deploy, the included SQL script should be run on boac-shared to create new tables.

After deploy, the `load_merged_sis_profiles` method can be used to fill the tables. Since it will be running off cached JSON, it should go quickly.

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -1,6 +1,7 @@
 from threading import Thread
 from boac import db
 from boac.lib import berkeley
+from boac.merged.sis_profile import merge_sis_profile
 from boac.models.job_progress import JobProgress
 from boac.models.json_cache import JsonCache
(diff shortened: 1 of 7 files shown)


raydavis:
@paulkerschen Since this is a new table, could you make sure it gets into models/development_db.py so that it's easy to create a fresh DB from scratch?

I forgot to do that myself when I added the JobProgress table, and it looks like we might be missing the db_relationships pair as well. :(

raydavis:
so, uh, maybe you could fix our bugs while you're at it thankyouverymuch...

paulkerschen:
@raydavis Done!

raydavis:
We're all normalized and we want our freedom.

PR #225, opened 2017-12-22
BOAC-285 Add non-current-term option to admin data refresh API

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-285

Also fixes bug in Canvas scores loader.


johncrossman:
Past and present - a nondiscriminatory API.

PR #226, opened 2017-12-22
BOAC-206, style '.dropdown-menu > li > span' like an anchor

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-206

I hope Teena Marie will approve.

### DIFF
--- a/boac/static/app/cohort/myCohortsDropdown.html
+++ b/boac/static/app/cohort/myCohortsDropdown.html
@@ -9,8 +9,12 @@
       role="menu"
       aria-labelledby="btn-append-to-single-button"
       uib-dropdown-menu>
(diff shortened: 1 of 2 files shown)


boac/static/app/cohort/myCohortsDropdown.html
```diff
@@ -10,7 +10,9 @@
       aria-labelledby="btn-append-to-single-button"
       uib-dropdown-menu>
     <li role="menuitem" aria-disabled="true" data-ng-if="isLoading"><a>Loading....</a></li>
```
raydavis: Is it OK that this other not-a-link text is still a link element?

johncrossman: Good catch. I'll amend

PR #228, opened 2017-12-23
BOAC-206, fix to myCohorts dropdowns broke filters dropdowns

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-206

Cohort filters were impacted by new `.dropdown-menu > li > span` style. 

PR includes necessary `None` check `photo_api`.

### DIFF
--- a/boac/externals/cal1card_photo_api.py
+++ b/boac/externals/cal1card_photo_api.py
@@ -11,7 +11,7 @@ def get_cal1card_photo(uid):
     if response:
         return response.content
     else:
-        if hasattr(response, 'raw_response') and response.raw_response.status_code == 404:
+        if hasattr(response, 'raw_response') and hasattr(response.raw_response, 'status_code') and response.raw_response.status_code == 404:
             return False
         else:
             return None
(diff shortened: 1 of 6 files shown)


paulkerschen:
@johncrossman I opened #229 for a fix to the photo_api issue that should play nice with tests.

raydavis:
@paulkerschen @johncrossman  Can one of you say more about "the photo_api issue"? I don't find any explanation in JIRA or code comments, but I gather from PR https://github.com/ets-berkeley-edu/boac/pull/229 that we can somehow end up with a raw HTTP response which has no `status_code` attribute, which seems kind of interesting.

paulkerschen:
@raydavis In cases such as a timeout, raw_response can come back `None`.

As outlined at #229, a simple `if response.raw_response` doesn't work as a check for this case since error responses are falsey. That's what's causing the failing test at this PR.

#229 duck-types it by checking for a `status_code` attribute. I suppose `is not None` would work equally well.

raydavis:
Thanks, @paulkerschen . I knew about not relying on a `response.raw_response` boolean (which is why that wasn't being used in other caching code), but hadn't seen the timeout response yet.

PR #229, opened 2017-12-23
NOJIRA Safe, sane, consistent 404 checks

paulkerschen:
Our externals code needs to survive cases where `response.raw_response` is `None`. But since error responses evaluate to falsey, the simple check `response.raw_response and response.raw_response.status_code == 404` will never return true.


raydavis:
Question left at https://github.com/ets-berkeley-edu/boac/pull/229 .

PR #232, opened 2017-12-27
BOAC-291 Add Canvas Course User Assignments Analytics API to data loads

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-291


johncrossman:
More goodness in the mix. Thanks, @raydavis

PR #236, opened 2018-01-02
BOAC-292 Exceptions during a commit require explicit session rollback

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-292


paulkerschen:
Orderly alchemy with an explicit structure, just like Ben Jonson used to make.

PR #237, opened 2018-01-02
BOAC-297 Matrix view: make axis labels visible

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-297

PR #239, opened 2018-01-03
BOAC-312 Switch the cohort matrix Y-axis to Assignment Grades

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-312

I've checked and this doesn't break @johncrossman 's https://github.com/ets-berkeley-edu/boac/pull/237 fix.


paulkerschen:
The thinnest of lines between sense and nonsense.

PR #242, opened 2018-01-03
BOAC-288 Hide preferred name when identical to primary name

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-288


johncrossman:
+1

PR #243, opened 2018-01-03
BOAC-287 Redefine 'primary section' as 'section with a grading basis'

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-287

The Hub doesn't explicitly tell us which sections are primary. We've been using "nonzero unit count" as a proxy, but it looks like the occasional zero-unit primary section is a thing. Let's try a different metric! Fundamental as this change is, I haven't found any issues in spot checks on various students.


raydavis:
SIS hermeneutics marches on.

PR #241, opened 2018-01-03
BOAC-298 Performance improvements to cohort boxplot rendering

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-298

PR #245, opened 2018-01-03
BOAC-298 Custom cohorts need boxplots too

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-298

Follow-up to #241. Without this explicit nudging, boxplots won't draw for custom cohorts.


johncrossman:
Wink, wink, nudge, nudge. Thanks, @paulkerschen

PR #246, opened 2018-01-04
BOAC-277, bookmarkeable matrix view; cohort object is always defined, even in search

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-277

The 'cohortController` is simplified if we make a few assumptions:
* `cohort` is always defined, even if we are execute-search mode
* `cohort.code` is always defined for team and custom cohort
* the block of code under `$scope.$watch('$location' ...` might be used for 'Return to results', too.     Note: `$location` is not immediately available during init phase

### DIFF
--- a/boac/static/app/cohort/cohort.html
+++ b/boac/static/app/cohort/cohort.html
@@ -14,13 +14,13 @@
       <span class="faint-text" data-ng-if="error.message">(<span data-ng-bind="error.message"></span>)</span>
     </div>
   </div>
-  <div class="cohort-label" data-ng-if="!isLoading && !error && !isSaving && (search.results.totalCount >= 0)">
+  <div class="cohort-label" data-ng-if="!isLoading && !error && !isSaving && (cohort.totalMemberCount >= 0)">
     <h1 data-ng-if="cohort">
-      <span data-ng-bind="cohort.label || cohort.name"></span>
-      <span class="faint-text">(<span data-ng-bind="search.results.totalCount"></span> members)</span>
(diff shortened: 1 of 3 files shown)


raydavis:
We have path!

boac/static/app/cohort/cohort.html
```diff
@@ -14,13 +14,13 @@ <h1>Error</h1>
...
   </div>
-  <div class="cohort-label" data-ng-if="!isLoading && !error && !isSaving && (search.results.totalCount >= 0)">
+  <div class="cohort-label" data-ng-if="!isLoading && !error && !isSaving && (cohort.totalMemberCount >= 0)">
     <h1 data-ng-if="cohort">
-      <span data-ng-bind="cohort.label || cohort.name"></span>
-      <span class="faint-text">(<span data-ng-bind="search.results.totalCount"></span> members)</span>
+      <span data-ng-bind="cohort.name"></span>
+      <span class="faint-text">(<span data-ng-bind="cohort.totalMemberCount"></span> members)</span>
     </h1>
     <h1 data-ng-if="!cohort && !isLoading && !isCreateCohortMode">
```
paulkerschen: @johncrossman Does the `data-ng-if` on these h1 elements need to be tweaked? Now that `cohort` is always defined, I'd expect we would always get the upper option and never the lower.

johncrossman: I believe you're right, @paulkerschen. I'll fix in follow up PR. Thank you

PR #247, opened 2018-01-04
BOAC-322 Fix API header check

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-322

Earlier commits have broken the caching job API on boac-dev. So far as I can tell from my local environment, this should be a fix, but I can't be sure until it lands on the server.


paulkerschen:
`KeyError` is one unforgiving shibboleth, isn't it.

PR #248, opened 2018-01-04
BOAC-322 One bug fix, some diagnostic logging for cache refresh

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-322

Here's hoping this gives enough information to go on...


paulkerschen:
Hm, that seems like an unwelcome mystery.

PR #252, opened 2018-01-04
BOAC-333, fix workflow to 'create cohort'; put 'Intensive' name to feed

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-333

Fix regressions caused by 'bookmarkability' work.


raydavis:
Looks reasonable from here, @johncrossman -- I'll merge post-Travis

PR #254, opened 2018-01-04
BOAC-327 Cohort dropdown: use ui-router to force reload

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-327

This ensures that selecting a dropdown item forces a reload of the original criteria, even if the browser location hasn't changed in the meantime.


johncrossman:
Nifty use of `ui-sref`, @paulkerschen

PR #255, opened 2018-01-04
BOAC-314 Merge Course Student Assignments Analytics data (server-side)

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-314

The merged data does not yet appear in the UX, so this change is not aimed for QA and the first production release.


paulkerschen:
Thanks, @raydavis, looks right from here.

PR #256, opened 2018-01-04
BOAC-334 Skip concurrent-enrollment academic status

paulkerschen:
See https://jira.ets.berkeley.edu/jira/browse/BOAC-334 for rationale on why skipping this part of the feed is probably better than trying to fold it in.


raydavis:
I agree that skipping makes sense in this particular context. Thanks, @paulkerschen .

PR #257, opened 2018-01-05
BOAC-319, selected filters are reflected in $location.search (browser URL); prep for 'Return to results'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-319

Noteworthy:
* cohort APIs `/api/cohorts/all` and `/api/cohorts/my` no longer include full student list per cohort, but `totalMemberCount` is included.
* `/cohort/:code` route is now `/cohort?c=`. This allows us to change `$location` args without re-init of Angularjs controller. See 'reloadOnSearch: false'.
* Bookmarkable searches is precursor to 'Return to results' from student profile. A second PR will need to finish the job. (Parameter names are quite concise – e.g., 'g' for 'gpaRanges' – in order to shorten URLs as much as possible.) 

Use `w=1` for easier PR read.


boac/static/app/cohort/cohortController.js
```diff
@@ -367,35 +375,66 @@
...
+                $scope.isLoading = false;
+                // Track view event
+                if (isNaN($scope.cohort.code)) {
+                  googleAnalyticsService.track('team', 'view', $scope.cohort.code + ': ' + $scope.cohort.name);
+                } else {
+                  googleAnalyticsService.track('cohort', 'view', $scope.cohort.name, $scope.cohort.code);
+                }
+              });
+            });
+          }
```
johncrossman: Required restructuring of `init()` above because these steps are necessary:
1. what are search-filter options
2. per request parameters, which options should be 'selected=true'
3. perform search with requested args

boac/static/app/routes.js
```diff
@@ -25,9 +25,10 @@
         isPublic: true
       })
       .state('cohort', {
-        url: '/cohort/:code',
+        url: '/cohort?c',
         templateUrl: '/static/app/cohort/cohort.html',
-        controller: 'CohortController'
+        controller: 'CohortController',
+        reloadOnSearch: false
```
johncrossman: From AngularJS docs:
_**reloadOnSearch:** Boolean (default true). If false will not retrigger the same state just because a search/query parameter has changed. Useful for when you'd like to modify $location.search() without triggering a reload._

paulkerschen:
Thanks for the documentation on reloadOnSearch, @johncrossman. Looks good to me!

PR #273, opened 2018-01-08
BOAC-288 Clean up extra whitespace from SIS Student API

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-288

Follow-up to #242.


johncrossman:
[insert vacuum sound here]  Thx, @paulkerschen

PR #287, opened 2018-01-09
BOAC-342, natural level order is Freshman, Soph, Junior, etc., not alpha

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-342

Followed tip: https://stackoverflow.com/questions/866465/order-by-the-in-value-list


tox.ini
```diff
@@ -48,7 +48,7 @@ exclude =
 format = ${cyan}%(path)s${reset}:${yellow_bold}%(row)d${reset}:${green_bold}%(col)d${reset}: ${red_bold}%(code)s${reset} %(text)s
 ignore = E731,I201
 import-order-style = google
-max-complexity = 11
+max-complexity = 10
```
johncrossman: `max-complexity` is back to what it was.

raydavis:
Nice!

PR #295, opened 2018-01-09
BOAC-349 Fix clip path, overflow for users on matrix

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-349


johncrossman:
Users on the matrix thank you, @paulkerschen

PR #300, opened 2018-01-10
BOAC-355, gulp dist --allow-root in same block as other npm installs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-355

Okay, one last try – I'm hoping `--allow-root` is the secret sauce. And, I neglected to include necessary `gulpfile.js` in buildspec.

If does not work, I'll revert, set up my own pipeline, get it right and then submit proper PR.

### DIFF
--- a/.ebextensions/01_create_apache_conf.config
+++ b/.ebextensions/01_create_apache_conf.config
@@ -1,19 +1,6 @@
 #
 # Apache configuration files and keys.
 #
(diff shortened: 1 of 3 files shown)


raydavis:
Thank you for continuing to hammer on this, @johncrossman . Visibility is sure poor inside that cloud...

PR #301, opened 2018-01-10
BOAC-349 Even expanded heads do not deserve to be cut off

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-349

Follow-up to #295.


johncrossman:
The Coneheads (from France) thank you, @paulkerschen

PR #303, opened 2018-01-10
BOAC-355, revert changes to rev js and css files, then do it right

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-355

Reverting all commits of BOAC-355.

`eb-deploy` has won this battle. I will set up a personal Pipeline and consider using buildspec to do the `gulp dist` / bower install.


raydavis:
You'll whup 'em yet, @johncrossman !

PR #304, opened 2018-01-10
BOAC-341 Add team memberships to cohort list view

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-341


johncrossman:
And just like that... it's done.

PR #308, opened 2018-01-10
BOAC-390, click outside cohort-filter dropdown to close it

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-360

`auto-close="outsideClick"` is a well documented feature (see http://angular-ui.github.io/bootstrap/) and it has [a known issue with workaround](https://github.com/angular-ui/bootstrap/issues/6038), as seen in this PR.


paulkerschen:
Thanks @johncrossman -  I suspect those `cohort-filters-column` divs might want to be directives one day, but looks good to me.

PR #310, opened 2018-01-11
BOAC-370, order by group_name, not group_code

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-370

Most of the diff is under `tests`.  Front-end no longer passes 'group_code'.


paulkerschen:
I'm encouraged that our history major has such a strong GPA.

PR #312, opened 2018-01-11
BOAC-373 Clear temporary query strings from browser history

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-373


johncrossman:
Looks right, @paulkerschen

PR #314, opened 2018-01-11
BOAC-360, Angular ui-bootstrap needs menu-options in DOM for auto-close=outsideClick to work

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-360

`auto-close=outsideClick` works on initial page load cuz DOM is in order. But, due to `ng-if`, the DOM is tweaked post-search and a click on menu-options is seen as an "outsideClick".


paulkerschen:
So tricky!

PR #316, opened 2018-01-12
BOAC-332 Integrate authentication with ui-router

paulkerschen:
A robust fix for https://jira.ets.berkeley.edu/jira/browse/BOAC-332 requires introducing a pattern where login status is checked on every front-end state change. The standard way of doing this in ui-router 0.x is with resolvers:

https://github.com/angular-ui/ui-router/wiki#resolve

An advantage of the resolver pattern is that it makes authentication entirely the responsibility of routes.js, so we don't have to remember to include auth wrapping in our controllers.

### DIFF
--- a/boac/static/app/auth/authFactory.js
+++ b/boac/static/app/auth/authFactory.js
@@ -2,7 +2,7 @@
 
   'use strict';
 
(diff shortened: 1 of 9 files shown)


johncrossman:
Very nice. Certainly more elegant, @paulkerschen

PR #321, opened 2018-01-12
BOAC-377, get_calnet_user_for_uid used by '/api/cohorts/all' and '/api/profile'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-377
https://jira.ets.berkeley.edu/jira/browse/BOAC-369

Profile data of `authorized_users` is pulled from CalNet. Benefit to the customer: 'Everyone's Cohorts' page displays names, not UIDs.

### DIFF
--- a/boac/api/cohort_controller.py
+++ b/boac/api/cohort_controller.py
@@ -1,6 +1,7 @@
 from boac.api.errors import BadRequestError, ForbiddenRequestError, ResourceNotFoundError
 from boac.lib import util
 from boac.lib.http import tolerant_jsonify
+from boac.merged import calnet
 from boac.merged import member_details
 from boac.models.cohort_filter import CohortFilter
 from boac.models.student import Student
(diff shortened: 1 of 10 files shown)


boac/api/cohort_controller.py
```diff
@@ -17,7 +18,14 @@ def all_cohorts():
             if uid not in cohorts:
                 cohorts[uid] = []
             cohorts[uid].append(cohort)
-    return tolerant_jsonify(cohorts)
+    owners = []
+    for uid in cohorts.keys():
+        owner = calnet.get_calnet_user_for_uid(app, uid)
```
paulkerschen: @johncrossman I don't know too much about LDAP, but isn't it optimized for bulk queries? At least in CalCentral, the usual pattern would be a single `get_calnet_user_for_uids` method call that takes an array of uids and returns an array of users.

johncrossman: Good point, @paulkerschen  Amending

paulkerschen:
Discussed with @johncrossman and @raydavis, and we're agreed that given the small size of our current user population, trying to finesse cache lookup with bulk queries is complexity without payoff.

PR #322, opened 2018-01-12
BOAC-369, order by first_name, last_name on 'Everyone's Cohorts'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-369


paulkerschen:
Yep.

PR #324, opened 2018-01-13
BOAC-374 Cohort ordering for multiple teams, majors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-374

Easy part: order team groups and majors alphabetically.
Hard part: ensure that, in the edge case where group or major is both a filter criterion and an ordering criterion, alphabetical order is still respected. This requires an extra join in the query; I'm the first to admit this is a little persnickety for an edge case, but it is at least discreetly tucked away under a conditional.


johncrossman:
Very well done, @paulkerschen   Not so persnickety... Users would notice.

PR #326, opened 2018-01-16
BOAC-379, flake8-docstrings, flake8-strict, flake8-tidy-imports and necessary fixes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-379

Many files touched and yet few significant changes. The one debatable – and IMO, desirable – rule is `D400: First line should end with a period`. The D400 downside is **specifically in test classes**, the `tox -e test` output looks a bit odd:
```
Cohort API. Returns all cohorts per owner.
Cohort API. Returns a well-formed response with custom cohort.
...
```
In test classes, you can omit class-level docs and you will get useful output:
```
TestCohortDetail Returns all cohorts per owner.
TestCohortDetail Returns a well-formed response with custom cohort
...
```

### DIFF
--- a/boac/__init__.py
+++ b/boac/__init__.py
@@ -6,11 +6,11 @@ db = SQLAlchemy()
 
 
 def std_commit(allow_test_environment=False):
-    """Commit failures in SQLAlchemy must be explicitly handled. This function follows the suggested default,
-    which is to roll back and close the active session, letting the pooled connection start a new transaction
-    cleanly.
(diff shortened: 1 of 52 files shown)


paulkerschen:
I don't especially mind periods in my test outputs. Thanks for all the cleanup, @johncrossman!

raydavis:
I prefer leaving out the class-level doc anyway.

PR #333, opened 2018-01-18
BOAC-251, blue skies outage page when HTTP status code 502-504

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-251

See designs in BOAC-251. The `fa-plane` icon has a 45 degree tilt:
image

I've also included `favicon.ico`:
image


.gitignore
```diff
@@ -12,6 +12,7 @@ boac/static/lib
 # Local configs
 config/*-local.py
 config/build-summary.json
+config/*-local.conf
```
johncrossman: Apache work is verified locally with `*-local.conf` file.  Git should ignore.

paulkerschen:
That calming moment when the pilot says, "Well, this is awkward," at 30,000 feet. Thanks @johncrossman.

PR #338, opened 2018-01-18
BOAC-385, put EB_ENVIRONMENT in config file (03_download_local_configuration)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-385

`EB_ENVIRONMENT` is put to production-local.py at deploy time. The value of that config:
* surfaces in `/api/config`
* will be used by upcoming BOAC-385 script

See `ebEnvironment` in https://asc-pilot-dev.berkeley.edu/api/config  (deployed with my CodePipeline)

### DIFF
--- a/.ebextensions/03_download_local_configuration.config
+++ b/.ebextensions/03_download_local_configuration.config
@@ -3,4 +3,6 @@
 #
 container_commands:
   01_get_configuration_file:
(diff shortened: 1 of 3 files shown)


paulkerschen:
The app that knows its own name. Clever!

PR #341, opened 2018-01-19
BOAC-385, script uses EB_ENVIRONMENT to deploy s3://boac-config/__.py to production-local.py

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-385

`AWS_PROFILE=boac` is not needed in this particular approach. Tested on boac-dev.


paulkerschen:
mariobros 1vinelol

PR #342, opened 2018-01-19
BOAC-413 Redirect http to https

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-413

Hat tip for https://stackoverflow.com/questions/14693852/how-to-force-https-on-elastic-beanstalk/38751749#38751749 for a workable solution.

### DIFF
--- a/.ebextensions/01_create_apache_conf.config
+++ b/.ebextensions/01_create_apache_conf.config
@@ -12,6 +12,12 @@ files:
       WSGISocketPrefix run/wsgi
       WSGIRestrictEmbedded On
 
+      # Redirect http to https.
+      RewriteEngine On
+      <If "-n '%{HTTP:X-Forwarded-Proto}' && %{HTTP:X-Forwarded-Proto} != 'https'">
+        RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R,L]
+      </If>
+
       Listen 443
 
       <VirtualHost *:443>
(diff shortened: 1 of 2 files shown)


.ebextensions/01_create_apache_conf.config
```diff
@@ -12,6 +12,12 @@ files:
       WSGISocketPrefix run/wsgi
       WSGIRestrictEmbedded On
 
+      # Redirect http to https.
+      RewriteEngine On
+      <If "-n '%{HTTP:X-Forwarded-Proto}' && %{HTTP:X-Forwarded-Proto} != 'https'">
+        RewriteRule (.*) https://%{HTTP_HOST}%{REQUEST_URI} [R,L]
+      </If>
```
johncrossman: @paulkerschen Just curious...  Recent Apache docs have https://wiki.apache.org/httpd/RewriteHTTPToHTTPS   Is the above equivalent or does it do more?

paulkerschen: @johncrossman The Elastic Beanstalk load balancer defeats more straightforward solutions. See https://stackoverflow.com/questions/14693852/how-to-force-https-on-elastic-beanstalk for a long discussion, but the meat of it is that the `X-Forwarded-Proto` header is our way of determining, on the instance, how clients are connecting to the load balancer.

johncrossman:
Elasticity has its cost. Thanks for the edubication, @paulkerschen

PR #344, opened 2018-01-22
BOAC-420, add/remove from watchlist on profile and list view

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-420
https://jira.ets.berkeley.edu/jira/browse/BOAC-417

UI samples:
image

image

Designs will likely evolve but the front-end wiring is in place.


boac/static/app/user/watchlistToggleDirective.js
```diff
@@ -0,0 +1,58 @@
...
+    };
+
+    return {
+      // @see https://docs.angularjs.org/guide/directive#template-expanding-directive
+      restrict: 'E',
+
+      // @see https://docs.angularjs.org/guide/directive#isolating-the-scope-of-a-directive
+      scope: {
+        sid: '=',
+        view: '=bind',
```
paulkerschen: @johncrossman Do you want `=bind` or `=` here? I'd expect `=bind` to look for an attribute called `data-bind` in the HTML.

johncrossman: @paulkerschen  Ah, I now see the prob I hit earlier. On landing page, for example, I pass 'student' expecting to get fa styling of profile page. However, I missed the fact that 'student' is defined in `$scope` and that causes a conflict of sorts. The `'=bind'` worked around the conflict w/o understanding the problem. (Note: use of 'bind' should not change attribute naming)

johncrossman: Amendment is on the way...

boac/static/app/user/watchlistToggleDirective.js
```diff
@@ -0,0 +1,56 @@
...
+
+    return {
+      // @see https://docs.angularjs.org/guide/directive#template-expanding-directive
+      restrict: 'E',
+
+      // @see https://docs.angularjs.org/guide/directive#isolating-the-scope-of-a-directive
+      scope: {
+        sid: '=',
+        watchlist: '='
+      },
```
johncrossman: @paulkerschen Fix is in. I choose to go with more explicit naming of attribute ('faSize') and remove from `scope` above.

paulkerschen:
Thanks @johncrossman!

PR #347, opened 2018-01-22
 BOAC-402 Make base64 URL-safe, fix console error

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-402

Two commits for what turned out to be two unrelated front-end issues.

The first one honors gulp-rev-replace's expectation that a given reference will only appear once per file.


johncrossman:
What's better than encoding? Answer: more encoding. Thanks, @paulkerschen

PR #348, opened 2018-01-22
BOAC-361, boac/__init__.py is our single source for version info

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-361

This convention is used by [`pip` project](https://packaging.python.org/guides/single-sourcing-package-version/)


paulkerschen:
If it doesn't fit, you must __init__. Thanks @johncrossman.

PR #350, opened 2018-01-22
BOAC-435, the large spinner is not too big to fail

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-435

Collateral damage of PR #333


paulkerschen:
That spinner is its own boss, all right.

PR #352, opened 2018-01-23
BOAC-428, SID in findStudent options; NOJIRA, 'Watchlist' is now 'My List'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-428


paulkerschen:
A nonthreatening name. Thanks @johncrossman

PR #353, opened 2018-01-24
BOAC-397, search by first and last name, even if student has middle name

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-397

Two filters needed:
* filter list of students per input
* put `<strong>` on matching substrings

Default filters were not good enough.


raydavis:
That's some hard-working JavaScript....

PR #354, opened 2018-01-24
[QA] BOAC-407 Support merging of updates from ASC

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-386

Currently `merge_tsv` has to be run manually rather than from a server, but since this version of code will hit Production and QA data, we might as well make it official.


paulkerschen:
Students old, students new...

PR #356, opened 2018-01-25
BOAC-422 BOAC-423 BOAC-424 BOAC-425 Back end for alerts framework

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-422
https://jira.ets.berkeley.edu/jira/browse/BOAC-423
https://jira.ets.berkeley.edu/jira/browse/BOAC-424
https://jira.ets.berkeley.edu/jira/browse/BOAC-425

Models, controller and tests for showing alerts to their viewers. Not included: 
- How is viewership of a given alert decided? (requirements TBD)
- How are alerts generated in the first place? (spoiler: they will probably have to be tied to our data update logic)

### DIFF
--- a/boac/models/alert.py
+++ b/boac/models/alert.py
@@ -1,0 +1,111 @@
+import datetime
+
+from boac import db, std_commit
+from boac.models import authorized_user
+from boac.models.base import Base
+from boac.models.db_relationships import AlertView
+from sqlalchemy import text
+
+
+class Alert(Base):
+    __tablename__ = 'alerts'
+
+    id = db.Column(db.Integer, nullable=False, primary_key=True)
+    sid = db.Column(db.String(80), db.ForeignKey('students.sid'), nullable=False)
+    alert_type = db.Column(db.String(80), nullable=False)
+    key = db.Column(db.String(255), nullable=False)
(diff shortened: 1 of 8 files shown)


boac/models/alert.py
```diff
@@ -0,0 +1,106 @@
...
+    ),)
+
+    @classmethod
+    def create(cls, sid, alert_type, key=None, message=None, active=True):
+        # Alerts must contain a key, unique per SID and alert type, which will allow them to be located
+        # and modified on updates to the data that originally generated the alert. The key defaults
+        # to a string representation of today's date, but will more often (depending on the alert type)
+        # contain a reference to a related resource, such as a course or assignment id.
+        if not key:
+            key = datetime.datetime.now().strftime('%Y-%m-%d')
```
johncrossman: @paulkerschen Edge case: What if key is nothing but whitespace? We will not get the default above, right?

paulkerschen: @johncrossman Indeed, it's thinkable. I'll tighten to `is not None`.

johncrossman:
Very nice, @paulkerschen (Thank you for throwing in the stern warning.) I'll merge when green.

PR #361, opened 2018-01-26
BOAC-441 Per-advisor alert visibility logic

paulkerschen:
As arrived at in https://jira.ets.berkeley.edu/jira/browse/BOAC-441, this more or less flips the role of our "AlertView" association objects. Instead of explicitly associating advisors with displayed alerts, they are now used only to explicitly dismiss alerts for a particular advisor.

Since the latest designs bundle in alert counts on the homepage with cohort and list displays, we can use cohort and list memberships to work out alert visibility logic on the fly, and can dispense with a separate '/api/alerts/current' feed. (Per-student alert feeds are still useful and present.)

### DIFF
--- a/boac/lib/util.py
+++ b/boac/lib/util.py
@@ -1,6 +1,15 @@
 """Generic utilities."""
 
 
+def camelize(string):
+    def lower_then_capitalize():
+        yield str.lower
+        while True:
+            yield str.capitalize
+    string_transform = lower_then_capitalize()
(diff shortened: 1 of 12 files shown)


boac/lib/util.py
```diff
@@ -1,6 +1,15 @@
 """Generic utilities."""
 
 
+def camelize(string):
```
raydavis: Hello camel, my old friend...

raydavis:
I'm impressed that the flip didn't require a DB schema change.

PR #368, opened 2018-01-30
BOAC-439, 'Undeclared' filter option leads to SQL: m.major LIKE ...

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-439

Keep biz logic of 'Undeclared' out of front-end code.

### DIFF
--- a/boac/static/app/cohort/cohort.html
+++ b/boac/static/app/cohort/cohort.html
@@ -148,14 +148,17 @@
             uib-dropdown-menu
             data-ng-show="!isLoading && !isSaving">
           <li class="cohort-filter-list-item"
+              data-ng-class="{'divider': !major}"
               role="menuitem"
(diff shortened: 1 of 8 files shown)


boac/static/app/cohort/cohort.html
```diff
@@ -148,14 +148,17 @@ <h1>
             uib-dropdown-menu
             data-ng-show="!isLoading && !isSaving">
           <li class="cohort-filter-list-item"
+              data-ng-class="{'divider': !major}"
               role="menuitem"
               data-ng-repeat="major in search.options.majors">
```
johncrossman: `null` entry in list of filter-options causes `hr` divider style per uib.

paulkerschen:
Well threaded!

PR #370, opened 2018-01-30
BOAC-451, add 'Declared' menu option with wacky select/deselect logic

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-451

As described in BOAC-451, when user selects...
* "Declared" then deselect all other 'majors' checkboxes
* "Undeclared" then deselect "Declared"
* a specific major then deselect "Declared"

### DIFF
--- a/boac/static/app/cohort/cohortController.js
+++ b/boac/static/app/cohort/cohortController.js
@@ -217,57 +217,150 @@
     };
 
     /**
+     * @param  {Array}     allOptions     All options of dropdown
+     * @param  {Function}  isSelected     Determines value of 'selected' property
+     * @return {void}
+     */
+    var setSelected = function(allOptions, isSelected) {
+      _.each(allOptions, function(option) {
+        if (option) {
+          option.selected = isSelected(option);
+        }
+      });
+    };
+
+    /**
+     * @param  {String}    menuName     For example, 'gpaRanges'
+     * @param  {Object}    option       Has been selected or deselected
+     * @return {void}
+     */
+    var onClickOption = function(menuName, option) {
+      var delta = option.selected ? 1 : -1;
+      var existingValue = _.get($scope.search.count, menuName, 0);
+      _.set($scope.search.count, menuName, existingValue + delta);
+    };
+
+    /**
+     * Ordinarily, the value of 'selected' (per dropdown menu option) is managed by uib-dropdown-toggle in the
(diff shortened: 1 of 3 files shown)


boac/static/app/cohort/cohortController.js
```diff
@@ -216,58 +216,151 @@
       }
     };
 
+    /**
```
johncrossman: `w=1` will make diff below a bit more readable.

boac/static/app/cohort/cohortController.js
```diff
@@ -216,58 +216,151 @@
...
+       *    unitRange.onClick = onClickOption;
+       *  });
+       *  // Units, pacing
+       *  var selectedUnitRangesP = _.get($scope.cohort, 'filterCriteria.unitRangesPacing', []);
+       *  _.map($scope.search.options.unitRangesPacing, function(unitRange) {
+       *    unitRange.selected = _.includes(selectedUnitRangesP, unitRange.value);
+       *    unitRange.onClick = onClickOption;
+       *  });
+       *  $scope.search.count.unitRanges = selectedUnitRangesE.length + selectedUnitRangesP.length;
+       */
```
johncrossman: Soon we will deal with `units` logic above.

boac/static/app/cohort/cohortController.js
```diff
@@ -216,58 +216,151 @@
...
+            $scope.search.count.majors = 1;
+            setSelected($scope.search.options.majors, function(major) {
+              return major.name === optionGroup.name;
+            });
+          } else if (optionGroup.name === 'Undeclared') {
+            // If user selects "Undeclared" then "Declared" is deselected
+            manualSetSelected(menuName, 'Declared', false);
+            onClickOption(menuName, optionGroup);
+          }
+        } else {
```
paulkerschen: @johncrossman Just to check: is this `else` down here handling the case where `optionGroup.selected` is false, meaning either "Declared" or "Undeclared" has been deselected? In which case no special behavior is needed, and we just pass it on to `onClickOption`?

johncrossman: @paulkerschen Correct, no special logic – we only decrement 'count'.

paulkerschen:
And if you select "Declared" and "Undeclared" very quickly one after the other, you get fireworks and an extra 2000 points. Thanks!

PR #375, opened 2018-01-31
BOAC-461 Restrict student status alerts by term

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-461

Because a call to action for last October is nothing but a cruel taunt.


johncrossman:
With alerts, less is more.  Thanks @paulkerschen

PR #376, opened 2018-01-31
BOAC-437 Show units and grading basis for waitlisted classes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-437

Most of the code change is actually just rearrangement because tox thought merge_sis_enrollments_for_term was too complex.


johncrossman:
Waitlisted of the world unite!

PR #377, opened 2018-02-01
BOAC-438, put myWatchlist to scope before rendering matrix

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-438

First things first.


paulkerschen:
Ayup.

PR #379, opened 2018-02-01
BOAC-464 Follow the wandering current_term_id

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-464


johncrossman:
term_id can run but it cannot hide...

PR #380, opened 2018-02-01
BOAC-463, keep the list/matrix toggle when zero results in matrix view

johncrossman:
See most recent comments in https://jira.ets.berkeley.edu/jira/browse/BOAC-463


paulkerschen:
It was a feature all along! Thanks @johncrossman

PR #381, opened 2018-02-01
BOAC-466, menu options on cohort view get property-naming convention

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-466

The goal: Angular directive to render cohort-view dropdown. In order to have such a directive we must standardize property naming in menu options. Eg, `option.id` that plays nice in `<input id="team-{{option.id}}" />`


paulkerschen:
Whew. Thanks for covering all the angles, @johncrossman.

PR #382, opened 2018-02-02
BOAC-326, 'inIntensiveCohort' is new attribute in cohortFilters object and checkbox on cohort view

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-326

UX is in the Jira. I ripped out `/api/intensive_cohort` endpoint in favor of optional 'intensive' arg in `/api/student`. A remaining task: the manageMyCohorts view must include proper "pill" icon if intensive is true.

### DIFF
--- a/boac/api/cohort_controller.py
+++ b/boac/api/cohort_controller.py
@@ -35,23 +35,6 @@ def my_cohorts():
     return tolerant_jsonify(CohortFilter.all_owned_by(current_user.get_id(), include_alerts=True))
 
 
-@app.route('/api/intensive_cohort')
-@login_required
-def get_intensive_cohort():
-    order_by = util.get(request.args, 'orderBy', None)
-    offset = util.get(request.args, 'offset', 0)
-    limit = util.get(request.args, 'limit', 50)
-    results = Student.get_students(in_intensive_cohort=True, order_by=order_by, offset=offset, limit=limit)
-    member_details.merge_all(results['students'])
-    return tolerant_jsonify({
-        'code': 'intensive',
-        'label': 'Intensive',
-        'name': 'Intensive',
-        'members': results['students'],
-        'totalMemberCount': results['totalStudentCount'],
-    })
-
-
 @app.route('/api/inactive_cohort')
 @login_required
 def get_inactive_cohort():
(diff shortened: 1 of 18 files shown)


boac/models/cohort_filter.py
```diff
@@ -43,13 +43,17 @@ def create(
             levels=None,
             majors=None,
             unit_ranges=None,
+            in_intensive_cohort=None,
     ):
+        # If in_intensive_cohort is True then search intensive cohort; if equals False then search
+        # non-intensive students; if equals None then search all students.
```
johncrossman: The front-end plays by the rules described above. IMO, having this standard (True, False or None) on the backend might be beneficial for future use cases (eg, search for `is_inactive`).

boac/static/app/cohort/cohortController.js
```diff
@@ -519,6 +542,13 @@
       }
     };
 
+    $scope.disableSearchButton = function() {
+      // Disable button if page is loading or no search criterion is selected
+      var count = $scope.search.count;
+      return $scope.isLoading || $scope.isSaving || !count.gpaRanges && !count.groupCodes && !count.levels &&
+        !count.majors && !count.unitRanges && ($scope.showIntensiveCheckbox && !$scope.search.options.intensive);
```
paulkerschen: @johncrossman Would more parentheses make precedence on these booleans clearer? I always get lost when both `&&` and `||` are involved.

johncrossman: @paulkerschen Yes, but our linter disagrees. I can disable the rule and fix. Opinion?

johncrossman:
```
  548:53  error  Gratuitous parentheses around expression  no-extra-parens
```

johncrossman: The `no-extra-parens` rule is off. Let the entropy begin.

boac/api/cohort_controller.py
```diff
@@ -92,6 +75,7 @@ def create_cohort():
     levels = util.get(params, 'levels')
     majors = util.get(params, 'majors')
     unit_ranges = util.get(params, 'unitRanges')
+    in_intensive_cohort = bool(util.get(params, 'intensive'))
```
johncrossman: I just noticed that the line above needs `to_bool_or_none` instead of `bool()`.  Amend coming in moments.

paulkerschen: @johncrossman Looks like a test might need to be updated to match the change?

johncrossman: @paulkerschen  Ah yes

johncrossman: Fyi, fix coming in ~10 min.

johncrossman:
Fyi, Travis green.  cc: @paulkerschen

paulkerschen:
And BOAC goes ternary. Thanks @johncrossman.

PR #383, opened 2018-02-02
BOAC-407 Support updates of ASC Athletes data from API

raydavis:

* The ASC API uses a request parameter to hold credentials, necessitating some new censorship in the `http` module.
* Sorry about the length of the `import_asc_athletes` module. When we stop having to use TSV import on production, it'll be easier to reduce its weight.
* I dropped the now obsolete "-AA" suffixes from the team group codes.

Next step: Add the ASC import to the Admin API so that it can join the cache-loading cron job.


boac/lib/http.py
```diff
@@ -49,7 +50,16 @@ def request(url, headers={}, auth=None):
     response = None
     try:
         # TODO handle methods other than GET
-        response = requests.get(url, headers=headers, auth=auth)
+
+        # By default, the urrlib3 package used by Requests will log all request parameters at DEBUG level.
```
paulkerschen: @raydavis Developer's choice whether to leave the sort of pleasing `urrlib3` typo in.

paulkerschen:
People say short fiction is dead, but _I_ like those tests.

PR #384, opened 2018-02-02
BOAC-472 Move tooltip from icon to button element

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-472

Per https://github.com/twbs/bootstrap/issues/11359, Firefox doesn't like tooltips on elements within buttons.


johncrossman:
A fair reqt. Thx @paulkerschen

PR #388, opened 2018-02-05
BOAC-407 Add admin API support for ASC imports

raydavis:

This also makes the cachejob status report less verbose and more informative.


johncrossman:
Very nice, @raydavis  Thx

PR #389, opened 2018-02-05
BOAC-477, filter out college-specific 'undeclared' major options

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-477
https://jira.ets.berkeley.edu/jira/browse/BOAC-476


raydavis:
The Undeclared are strangers everywhere.

PR #394, opened 2018-02-06
BOAC-414 BOAC-415 Add ability to view Inactive students

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-409
https://jira.ets.berkeley.edu/jira/browse/BOAC-414
https://jira.ets.berkeley.edu/jira/browse/BOAC-415


paulkerschen:
And that's how we learned that, in the end, an inactive student is much like you or me.

PR #395, opened 2018-02-06
BOAC-473, tertiary sort order for get_students

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-473

See Jira for use case.


paulkerschen:
Three-layer cake.

PR #397, opened 2018-02-06
BOAC-485 Drop "Participations" metrics from the UX

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-485

Sometimes I feel AngularJS 1 pushes us into unnecessary replication. This isn't one of those times.


johncrossman:
Yes we have no replication. Thanks @raydavis

PR #403, opened 2018-02-07
BOAC-488 De-duplicate sids within students query

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-488

Using GROUP BY instead of DISTINCT preserves our ordering logic but de-duplicates within SQL. This eliminates the need for an extra function and ensures that de-duplication works across pagination offsets.


raydavis:
So many times the answer is "you shoulda used GROUP BY"...

PR #408, opened 2018-02-08
BOAC-489, You have no cohorts? Create one.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-489

A call to action! (We do not use 'Create your first...' because it's not necessarily the user's first.)

### DIFF
--- a/boac/static/app/landing/landing.css
+++ b/boac/static/app/landing/landing.css
@@ -18,6 +18,10 @@
   background-color: #ccc;
 }
 
+.landing-alert {
+  font-size: 16px;
+}
+
 .landing-list {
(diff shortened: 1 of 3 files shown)


boac/static/app/landing/landing.css
```diff
@@ -18,6 +18,10 @@
   background-color: #ccc;
 }
 
+.landing-alert {
+  font-size: 16px;
```
johncrossman: font-size was 20px, due to `body-text` – too much for such a simple alert.

raydavis:
Encouraging yet discreet!

PR #409, opened 2018-02-08
BOAC-502 Adjust alert due dates for sunny California

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-502


raydavis:
I anticipate a utility function in our future...

PR #411, opened 2018-02-09
BOAC-164 Configure & verify Data Loch connection

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-164

I'll add it to the Dev configuration after the merge (if any).


paulkerschen:
`SELECT 1`! It's a datum!

PR #415, opened 2018-02-15
[QA ONLY] BOAC-520 Revert assignment alerts until we have stricter criteria

paulkerschen:
QA-only PR for https://jira.ets.berkeley.edu/jira/browse/BOAC-520.

Apart from telling the analytics module not to create new assignment alerts, we take the "Issues" column out of the landing page so it doesn't show a whole stack of zeros.


raydavis:
Thank you for the battlefield surgery, Dr. @paulkerschen . The idea is that we delete existing alerts with SQL before deploying?

paulkerschen:
That would be my recommendation, exactly.

PR #417, opened 2018-02-15
BOAC-521 Tighten criteria for assignment alerts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-521

This PR inserts more stringent criteria for what counts as a late or missing assignment, in an attempt to screen out false positives as discovered in https://jira.ets.berkeley.edu/jira/browse/BOAC-519.

If the code passes muster and is merged, I'd like to follow it up with a cache refresh on boac-dev. We should then be able to take a broad look at which alerts survived the cull, and decide on this feature's fate for the sprint.

### DIFF
--- a/boac/lib/analytics.py
+++ b/boac/lib/analytics.py
@@ -1,6 +1,9 @@
+from datetime import datetime
 import math
 from statistics import mean
+
 from boac.externals import canvas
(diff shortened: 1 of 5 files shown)


raydavis:
Who says you can't squeeze blood from a stone? Thanks, @paulkerschen .

PR #423, opened 2018-02-21
BOAC-528 Display feedback link

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-528

The `hideFeedbackLink` is pretty cheesy, but alternative approaches I considered brought in too many redundancies for comfort. Let me know if you'd rather have the duplications, or if you see a better way.


johncrossman:
Cheese? Get the wine and Scottish biscuits!  Looks good, @raydavis

PR #424, opened 2018-02-21
BOAC-484, click to see course details causes other courses to collapse

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-484

* one and only one course can have `isShowingDetails = true`
* if no canvasSites then message: 'No additional info'
* remove stray 'Dropped'


raydavis:
Nice changes, @johncrossman -- thank you.

PR #426, opened 2018-02-22
BOAC-484, student profile accordion UX action: close-others=false

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-484

After "soaking" in dev for a day, the feature gets a tweak: do not auto-collapse course details when clicking elsewhere.


paulkerschen:
That special marinade.

PR #433, opened 2018-02-23
BOAC-539, if demoMode then demo-mode-blur style on sensitive info in UI

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-539

We can improve/iterate on this approach. A good-enough impl for purpose of upcoming demos.


johncrossman:
@raydavis @paulkerschen Travis green. Concerns?

raydavis:
I was concerned @paulkerschen might be about to merge it, but I'll overcome that.

PR #435, opened 2018-02-25
BOAC-539, demo-mode-blur for names in matrix view; student names in search replaced with random gen

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-539

One more bit of blurring.

### DIFF
--- a/boac/static/app/shared/utilService.js
+++ b/boac/static/app/shared/utilService.js
@@ -78,11 +78,35 @@
       return str.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
     };
 
+    /**
+     * @param  {String}     str      Word, phrase or whatever
+     * @return {String}              Unrecognizable representation of str
+     */
+    var obfuscate = function(str) {
+      var words = _.map(str.split(' '), function(word) {
+        var length = word.length;
+        if (length <= 1) {
+          // A dash or similar
+          return word;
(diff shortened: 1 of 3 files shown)


boac/static/app/shared/utilService.js
```diff
@@ -78,11 +78,20 @@
       return str.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
     };
 
+    /**
+     * @param  {String}     str      Word, phrase or whatever
+     * @return {String}              Unrecognizable representation of str
+     */
+    var obfuscate = function(str) {
+      return btoa(str);
```
raydavis: Using "btoa" as an obfuscation technique worries me a little. Deobfuscation is trivial, but it's easy to slip into a false sense of security.

paulkerschen: Assuming obfuscation doesn't need to be repeatable, perhaps random strings are best? Everyone's got their own favorite way to do that client-side, looks like: https://stackoverflow.com/questions/1349404/generate-random-string-characters-in-javascript

johncrossman:
A proper `demo-mode` impl might use something like http://chancejs.com or server/db side obfuscation. This impl was hurried and, for now, might qualify as "good enough".

raydavis:
Thanks, @johncrossman .

PR #439, opened 2018-02-27
BOAC-512 Data Loch integration, with mocking support

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-512

Next steps: Stow in JsonCache, handle DB exceptions, and get boiled-down data into the Analytics JSON (but not into the UX).


raydavis:
@paulkerschen , to reduce duplication I tried to fit the CSV-based code into `mockingbird`, but it just got too ugly for me. Let me know if you'd prefer that I try again & better.

raydavis:
I have no idea why Travis-CI is freaking over the test-exception, since local `tox` works fine. Guess I'll investigate...

raydavis:
Whoops, forgot my new fixture. Amendment coming.

paulkerschen:
Thanks, @raydavis. I can see how trying to get an either-or going with MockResponse and MockRows would get ugly fast. It might be nice to give the consolidation another go in the future, but since `mockingbird` could use some tuning up anyway (like preserving function signature with `decorator`), and since you've thoughtfully given `mockingdata` a consistent API, I'm fine with leaving that attempt to a future follow-up and getting on with Delivering Value.

PR #441, opened 2018-02-27
BOAC-468 CAS redirect: parse URLs right

paulkerschen:
Fix for the reopened https://jira.ets.berkeley.edu/jira/browse/BOAC-468; and because there's always something extra, make sure the added CAS login param doesn't leak into constructed return URLs.


johncrossman:
The delicate biz of query args.

PR #447, opened 2018-02-28
BOAC-512 Integrate Data Loch "page views" with API feed (not UX)

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-512

Also standardizes Data Loch function results to match the fixtures' list-of-dicts data structure.
And stows the Data Loch query, since it will be hit multiple times when more than one athlete is in a course site.


johncrossman:
"Loch" has landed. The ceremonial gold spike is in.

PR #450, opened 2018-02-28
BOAC-544 Links to Berkeley Academic Guide should open in a new tab

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-544


johncrossman:
Only thing better than a tab is more tabs.

PR #451, opened 2018-03-01
[QA] BOAC-554, unmatchedCanvasSites rendered with same template code as term.enrollments

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-554

We decorate the list of `unmatched` and then combine it with `term.enrollments`.  A dozen fewer lines in the template.


paulkerschen:
Elegant.

PR #453, opened 2018-03-01
BOAC-532, tweak to feed, class page sidebar and shared CSS

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-532

Not all `student-profile-*` styles will be shared but moving all of 'em to profile.css is simplest.


paulkerschen:
With tox content, who am I to stand in the way?

PR #455, opened 2018-03-02
BOAC-515 Clarify online assignment timeliness metric

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-515


johncrossman:
Clarified.

PR #456, opened 2018-03-02
BOAC-534, matrix view on class page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-534
https://jira.ets.berkeley.edu/jira/browse/BOAC-541

Bonus: email and phone icons


paulkerschen:
red-phone

PR #462, opened 2018-03-06
BOAC-532, class page sidebar; return-to link has informative label

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-532
https://jira.ets.berkeley.edu/jira/browse/BOAC-537


paulkerschen:
Thanks @johncrossman. So many pages, so subtly different...

PR #466, opened 2018-03-07
BOAC-536, phantom student, 'Class Average', using mock data for now

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-536

* icon (svg) is from BOAC-394 (design Jira)
* changes in `courseController` allow graceful fail when 'no such section'

### DIFF
--- a/boac/static/app/course/courseController.js
+++ b/boac/static/app/course/courseController.js
@@ -49,6 +49,70 @@
       utilService.goTo('/student/' + uid, $scope.section.displayName);
     };
 
+    var getClassAveragePseudoStudent = function() {
+      var mockDeciles = [
+        0,
+        80,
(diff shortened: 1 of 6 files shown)


boac/static/app/course/courseController.js
```diff
@@ -49,6 +49,70 @@
...
+          grade: 'B-',
+          gradingBasis: null,
+          midtermGrade: 'B+'
+        },
+        isClassAverage: true,
+        lastName: 'Class Average',
+        uid: 0
+      };
+    };
+
```
johncrossman: The mock data above is temporary. Good enough for demo.

raydavis:
Never trust the demos, as Plato used to say...

PR #467, opened 2018-03-07
BOAC-579 Catch & log errors in background job thread

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-579

No need to be shy about it.


johncrossman:
Not shy about merging. Thanks @raydavis

PR #468, opened 2018-03-08
BOAC-580, if demo-mode then exclude student name from page title and breadcrumb

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-580

`course.html` has blur-in-demo-mode logic but controller was missing `$scope.debugMode = ...`


raydavis:
Breadcrumbs are notoriously leaky.

PR #469, opened 2018-03-08
BOAC-583, commit the db delete before proceeding to db inserts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-583

Bulk deletion with SQL Alchemy has some gotchas according to [their own docs](http://docs.sqlalchemy.org/en/latest/orm/query.html#orm-specific-query-constructs) (see section on `delete synchronize_session=False`). I've encountered no such `IntegrityError` locally but boac-dev is a much better load test on this code. We want guaranteed order of operations: delete then add.


raydavis:
+1 to that, @johncrossman !

PR #470, opened 2018-03-08
BOAC-579 Do not let ASC API failures block cache refresh

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-579


johncrossman:
Do not let troublemakers make trouble. Thx @raydavis

PR #471, opened 2018-03-08
BOAC-569, avatar-large for list view; fix drawBoxplots() ref

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-569

The change in cohortController fixes a small bug: we had a stranded call to `drawBoxplots()` – not evident in the diff below. Furthermore, the function does not need to live in `utilService`.


paulkerschen:
Big heads!

PR #473, opened 2018-03-09
BOAC-545, boxplotPageViewsDirective replaces scope.postDigest strategy

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-545

Reusable code with less complexity.


paulkerschen:
A problem made for directives to solve. Thanks @johncrossman

PR #476, opened 2018-03-12
BOAC-584 Stow and preload merge_enrollments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-584

From local testing, it looks like NormalizedCacheEnrollment logic can get expensive when run in series on many students in a cohort. We should get performance improvements by treating it like other normalized cache tables, updating only when a cache reload updates the source data.


johncrossman:
Thank you for taking it to another level, @paulkerschen  A performance boost is needed.

PR #477, opened 2018-03-13
BOAC-536, begin to move average-student logic to server

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-536

Move extraneous stuff out of `courseController.js`. Mock data is still present but the structure is taking shape. Includes small enhancements on front-end.

### DIFF
--- a/boac/api/course_controller.py
+++ b/boac/api/course_controller.py
@@ -26,10 +26,11 @@ ENHANCEMENTS, OR MODIFICATIONS.
 
 from boac.api.errors import ResourceNotFoundError
 import boac.api.util as api_util
+from boac.lib import util
 from boac.lib.http import tolerant_jsonify
 from boac.merged import member_details
 from boac.models.normalized_cache_enrollment import NormalizedCacheEnrollment
-from flask import current_app as app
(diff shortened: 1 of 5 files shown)


boac/api/course_controller.py
```diff
@@ -39,14 +40,80 @@ def get_section(term_id, section_id):
     row = NormalizedCacheEnrollment.get_course_section(term_id=term_id, section_id=section_id)
     if not row:
         raise ResourceNotFoundError(f'No section {section_id} in term {term_id}')
-
     section = api_util.course_section_to_json(term_id=term_id, section=row)
-    if 'students' in section:
-        member_details.merge_all(section['students'], section['termId'])
+    if 'students' in section and section['students']:
+        students = section['students']
```
paulkerschen: @johncrossman FYI I've found that `get` offers a bit more concision on paths like these. You could do:

```
students = section.get('students')
if students:
```

Or `students = section.get('students', [])` and avoid the conditional entirely.

paulkerschen:
Average Joe is looking fine. Left one pedantic style q but won't kick up a fuss over it.

PR #482, opened 2018-03-14
BOAC-592 Restrict cohort boxplot directive to draw in list mode only

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-592

Because the matrix/list toggle is a `data-ng-show` for d3 reasons, the boxplot directive needs an extra `data-ng-if` so it knows not to draw when it's not there.


johncrossman:
Waiting on T...

PR #484, opened 2018-03-14
BOAC-578 Display expected graduation term

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-578


johncrossman:
How 'predictive' of us!

PR #487, opened 2018-03-15
BOAC-596, breadcrumbs are complicated and horrible

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-596

After a few weeks/months we can decide: resurrect a more sensible strategy or delete the logic.


paulkerschen:
Breadcrumbs are for waterfowl.

PR #488, opened 2018-03-16
BOAC-581, class page > filter canvas_sites per section id

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-581

Do not show canvas-course-code if section id maps to a single canvas course site (ccn). Such filtering is used when calculating 'Average Student' numbers – still a work in progress.

image


johncrossman:
Fyi, my plan is to add test coverage when the 'Average Student' impl is done.

PR #489, opened 2018-03-16
BOAC-600 Firmly insist that Angular update location from within d3 visualizations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-600


johncrossman:
"...`$apply` a kick..."  Thanks, @paulkerschen

PR #491, opened 2018-03-19
BOAC-594, a login page that is truly in the clouds

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-594

Includes dev-auth login.

image


boac/static/app/splash/splash.css
```diff
@@ -59,3 +81,99 @@
...
+  background-color: #0275d8;
+  height: 10px;
+  min-width: 320px;
+  max-width: 320px;
+}
+
+.splash-contact-us {
+  padding: 20px 40px 10px 40px;
+  max-width: 320px;
+  min-width: 320px;
```
johncrossman: I going to amend this use of max/min-width.  I think I can get away with `width=...` alone.

paulkerschen:
The UC Regents copyright notice has never flown so high. Thanks @johncrossman

PR #495, opened 2018-03-20
BOAC-607 Fix unit counts in level definitions

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-607


johncrossman:
Six of one, half dozen the other...

PR #496, opened 2018-03-20
BOAC-611, includeAverage=false for now

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-611


paulkerschen:
For now, everyone is free to imagine themselves above average.

PR #497, opened 2018-03-20
BOAC-603, no 'Expected graduation' term if Grad student

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-603


paulkerschen:
It's the only humane thing.

PR #498, opened 2018-03-20
BOAC-601, student name, in list view, cannot collide with adjacent column

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-601


johncrossman:
See latest comment in Jira for screenshot.

paulkerschen:
Thanks, @johncrossman, for making BOAC safe for the amply named.

PR #500, opened 2018-03-21
BOAC-620, autofocus on dev-auth UID and ng-submit on form

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-620

Use `?w=1` to review


paulkerschen:
Thanks for formalizing, @johncrossman.

PR #503, opened 2018-03-21
[QA] BOAC-581, only primary-section related Canvas course sites are shown on /course page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-581

For the sake of clarity: `/api/sections/ids_per_term` ->  `/api/sections/cached`

### DIFF
--- a/boac/api/course_controller.py
+++ b/boac/api/course_controller.py
@@ -82,9 +82,9 @@ def get_section(term_id, section_id):
     return tolerant_jsonify(section)
 
 
-@app.route('/api/sections/ids_per_term')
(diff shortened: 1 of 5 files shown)


johncrossman:
As discussed with @paulkerschen and @pfarestveit, BOAC feeds do not give the client-side code reliable info on sectionId <-> canvasCourseSiteId associations. The fix is non-trivial. So, with the v1.3 deployment, we will live with a slightly "noisy" course page: canvasCourseSites of primary *AND* secondary sections will be shown on BOAC's course page. The proper fix will come with the v1.4 release.

PR #507, opened 2018-03-21
NOJIRA, default.py gets DEMO_MODE = False and a bit o' documentation

johncrossman:
Let it be known...


paulkerschen:
And when BOAC is in Aries, it is an auspicious time to launch new initiatives.

PR #508, opened 2018-03-21
BOAC-626 Add library support for non-GET HTTP methods

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-626

Sharing the generic part of ets-berkeley-edu/nessie#6 back to BOAC.


johncrossman:
Aww... Nessie shared. Thanks @paulkerschen

PR #512, opened 2018-03-22
BOAC-623, cohort search filter preset > param is either string or array

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-623


paulkerschen:
This bug is cornered.

PR #515, opened 2018-03-27
[QA] BOAC-644 Better handling for race conditions in cache

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-644

Hat tip to http://skien.cc/blog/2014/01/15/sqlalchemy-and-race-conditions-implementing/


johncrossman:
I see that `std_commit()` has `rollback()` covered so... looks good. Thx @paulkerschen

PR #519, opened 2018-03-28
BOAC-653, student-groups backend must account for empty watchlists during transition

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-653

Bonus: `/api/group/<group_id>` added and test coverage on reported bug.


paulkerschen:
Supergroup!

PR #520, opened 2018-03-28
BOAC-652 Replace Canvas enrollment API analytics with loch

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-652


johncrossman:
Move over Canvas, there's a new Loch in town.

PR #522, opened 2018-03-29
BOAC-569, front-end directives and templates move to new studentGroup naming

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-568

Before we rework the UI, let's completely get rid of the `watchlist` scheme.


paulkerschen:
Watchlist? What watchlist?

PR #525, opened 2018-04-04
BOAC-669, home gets proper flex treatment w.r.t sortableAlertsTable

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-669

The UX goal is BOAC-401


raydavis:
If that CSS got any more flexible, it would be behind itself. Thanks, @johncrossman .

PR #532, opened 2018-04-06
BOAC-693, add/remove students and groups should refresh sidebar

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-693

It is no longer sufficient to pass around student SIDs. The listeners need whole student objects.


sandeepmjay:
Students are more than just a number !  LGTM @johncrossman.

PR #534, opened 2018-04-09
BOAC-699, sid is None if no such student

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-699

To verify the fix, I took Oliver off the Football team. No hard feelings.


raydavis:
So long as he stays on the baseball starting line-up, I think this is fine.

PR #535, opened 2018-04-09
BOAC-692, another pillar of the BOAC metaphor 

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-692

The boarding-pass img is in Jira above.


paulkerschen:
https://www.air-port-codes.com/airport-info/BRK/

No wonder we're lost!

PR #549, opened 2018-04-10
BOAC-698, min-height 100% for sidebar div

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-698

### DIFF
--- a/boac/static/app/main.css
+++ b/boac/static/app/main.css
@@ -108,10 +108,15 @@
   color: #cf1715;
 }
 
+.index-container {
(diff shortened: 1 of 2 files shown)


boac/static/app/main.css
```diff
@@ -151,11 +155,6 @@
   z-index: 999;
 }
 
-.main {
-  padding-top: 20px;
-  min-height: 370px;
-}
```
johncrossman: The above is unused.

PR #556, opened 2018-04-10
BOAC-742 CSS handling for text overflow in sidebar

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-742

Using CSS lets us specify pixel width rather than character count, catching out even the fattest of characters.


johncrossman:
Nifty!

PR #557, opened 2018-04-10
BOAC-746, section meeting info is ignored if there is no 'there' there

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-741

From debugger, here is what we ignore:
image

### DIFF
--- a/boac/api/util.py
+++ b/boac/api/util.py
@@ -148,11 +148,12 @@ def _get_meetings(section):
         }
         instructors = meeting.get('assignedInstructors', [])
         for entry in instructors:
-            instructor = entry['instructor']
-            for name in instructor['names']:
-                if name['type']['code'] == 'PRF':
+            instructor = entry.get('instructor', {})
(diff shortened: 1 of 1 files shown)


boac/api/util.py
```diff
@@ -148,11 +148,12 @@ def _get_meetings(section):
...
         for entry in instructors:
-            instructor = entry['instructor']
-            for name in instructor['names']:
-                if name['type']['code'] == 'PRF':
+            instructor = entry.get('instructor', {})
+            for name in instructor.get('names', []):
+                if name.get('type', {}).get('code') == 'PRF':
                     m['instructors'].append(name['formattedName'])
-        meetings.append(m)
+        if m['days'] or m['instructors'] or m['location']:
```
paulkerschen: @johncrossman Could we use `get` here for extra safety, or do we know these keys will be present even if the values are falsely?

johncrossman: Yes, a good idea. Amending...

PR #561, opened 2018-04-11
BOAC-747 Wide text was still too wide

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-747

We apologize for the earlier mechanical issue and look forward to serving you in the clouds.

download


johncrossman:
image

PR #565, opened 2018-04-11
BOAC-733, confirm 'added to group' on cohort and course pages

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-733
https://jira.ets.berkeley.edu/jira/browse/BOAC-754

**Extra:** No need to invoke `studentGroupFactory.getMyGroups` when `me` has what we need.


paulkerschen:
Add! Added. Another satisfied customer.

PR #567, opened 2018-04-11
BOAC-720, no splash before current user is in rootScope

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-720

image


paulkerschen:
In the shade...

PR #570, opened 2018-04-11
BOAC-699, when no student found, give a proper message and avoid failed use of sid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-699

image


paulkerschen:
That jersey number is wide open too.

PR #573, opened 2018-04-12
BOAC-720, angularStateName is accurate if set after stateChangeSuccess

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-720

On the coattails of @paulkerschen's `scrollTop` work, from yesterday.


paulkerschen:
Makes sense.

PR #584, opened 2018-04-13
BOAC-755, sidebar header indent depends on presence of href

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-755


paulkerschen:
Ahh, so subtle.

PR #586, opened 2018-04-16
BOAC-791, show error message on splash page if user is unauthorized

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-791

If/when BOAC has a broader audience, unauthorized users must know they are unauthorized. 

image


paulkerschen:
Informative! Thanks @johncrossman.

PR #589, opened 2018-04-19
BOAC-802 Teach cacheload JobProgress about the new tasks

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-802


johncrossman:
Teach the robots well.

PR #592, opened 2018-04-20
BOAC-806 BOAC-808 Simplify sidebar, remove special handling for primary group

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-806
https://jira.ets.berkeley.edu/jira/browse/BOAC-808

This is still rough and probably has lurking bugs, but given how much code it touches across the app, it seemed worth getting an early start to forestall merge conflicts.


raydavis:
Let's start delurking those bugs!

PR #595, opened 2018-04-23
BOAC-807, consistent naming fix: cohorts have students and depts have members

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-807

We have filtered cohorts and curated cohorts; let's have consistent naming given that they both contain students.  The word 'member' is reserved for `university_dept_members` usage.


paulkerschen:
Let's have consistent naming, indeed.

PR #600, opened 2018-04-23
BOAC-821 Follow-up fixes for page titles

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-821

Wow, that first round didn't come out very well. Let's try to do this right.


raydavis:
Let's!

PR #602, opened 2018-04-23
BOAC-792 Cohort filter criteria aren't searched, they are applied

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-792


johncrossman:
I wanted 'Feeling Lucky?' but this works.

PR #605, opened 2018-04-24
BOAC-627 Term enrollments feed: add primary status, site-to-section mappings

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-627


johncrossman:
Primacy rules.

PR #610, opened 2018-04-25
BOAC-835, users in authorized_users table are not always authorized

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-835

If user ain't authorized:
image


raydavis:
At least we said we're sorry.

PR #611, opened 2018-04-25
BOAC-822, no need for $scope.$destory when user hits esc key

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-822

I was hoping `finally(angular.noop);` would be good enough but no.


paulkerschen:
noop, Noop, NOOP!!!

PR #612, opened 2018-04-25
BOAC-817, ASC rules on front-end benefit from a simpler '/api/profile' feed

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-817

### DIFF
--- a/boac/static/app/cohort/cohortController.js
+++ b/boac/static/app/cohort/cohortController.js
@@ -28,10 +28,10 @@
   'use strict';
 
   angular.module('boac').controller('CohortController', function(
+    authService,
     cohortFactory,
     config,
     googleAnalyticsService,
-    me,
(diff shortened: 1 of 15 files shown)


boac/static/app/cohort/cohortController.js
```diff
@@ -137,50 +139,6 @@
...
-        majors,
-        unitRanges,
-        opts.intensive,
-        opts.inactive,
-        orderBy,
-        offset,
-        limit
-      );
-    };
-
```
johncrossman: `lint-js` became unhappy with the complexity of `cohortController.js` thus we delegate to studentSearchService.

paulkerschen:
Makes sense. We don't want the EHEEC getting mixed in with the UWASC.

PR #613, opened 2018-04-25
BOAC-837, cut red tape and regulation in cohort naming

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-837


raydavis:
Drain that swamp, hoist that seismically unsound highrise...

PR #616, opened 2018-04-26
BOAC-799, gulp rev third party js

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-799

I've tested this locally with Apache. Not a guaranteed fix but the odds are decent.


paulkerschen:
Ah ha! That seems like it would make a difference.

PR #618, opened 2018-04-27
BOAC-842, Flask user_loader does not benefit from AuthorizedUser lazy=True

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-842

The SQLAlchemy traceback includes ref to http://sqlalche.me/e/bhk3

```
DetachedInstanceError: Instance AuthorizedUser is not bound to a Session; attribute refresh operation cannot proceed
```


paulkerschen:
In these parts, son, lazy gets you killed.

PR #619, opened 2018-04-27
BOAC-841, if user is ASC then search excludes 'inactive' students

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-840
https://jira.ets.berkeley.edu/jira/browse/BOAC-841

The other "fix" here is to remove pagination in search results. If the first 50 results ain't got what you need then sorry, Charlie.


raydavis:
The "fix" is "in".

PR #621, opened 2018-04-27
BOAC-844, fontawesome 5

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-844

fa fa fas


paulkerschen:
fa sync fa spin fa plane

PR #622, opened 2018-04-27
BOAC-836 Add API support for cachejob/refresh?term=all

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-836

### DIFF
--- a/boac/models/job_progress.py
+++ b/boac/models/job_progress.py
@@ -39,13 +39,17 @@ class JobProgress:
 
     def delete(self):
         row = JsonCache.query.filter_by(key=self.key()).first()
+        job_state = row.json if row else None
         if row:
             db.session.query(JsonCache).filter(JsonCache.key == self.key()).delete()
-        return (row and row.json)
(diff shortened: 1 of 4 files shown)


boac/models/job_progress.py
```diff
@@ -39,13 +39,17 @@ def __init__(self, key=REFRESH_TERM):
 
     def delete(self):
         row = JsonCache.query.filter_by(key=self.key()).first()
+        job_state = row.json if row else None
         if row:
             db.session.query(JsonCache).filter(JsonCache.key == self.key()).delete()
-        return (row and row.json)
```
raydavis: After a long deadlock hunt, I learned that referring to a fetched row's data for the first time will start a new transaction. Thus, fetch before commit.

raydavis:
`cachejob/load?term=all` and `cachejob/continue` should work as well.

johncrossman:
Nice on the `fetch before commit`. Thx @raydavis

PR #623, opened 2018-04-28
BOAC-785, student search gets sortable-alerts UX w/ curated-cohort selector

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-785

image


raydavis:
We always like the combo of new feature and less code.

PR #625, opened 2018-04-30
BOAC-845 SQL scripts to copy external cache data in release

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-845

Not ideal, but it should give us more to go on after the next release.


johncrossman:
"Deploying data" is very 21st century. Thx @raydavis

PR #628, opened 2018-05-01
BOAC-842, lazy=False is not the default; Flask-Login needs eager loading

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-842

See http://docs.sqlalchemy.org/en/latest/orm/loading_relationships.html#joined-eager-loading

The default relationship.lazy is 'select', which indicates lazy loading. Didn't catch that in initial fix: PR #618

### DIFF
--- a/boac/models/authorized_user.py
+++ b/boac/models/authorized_user.py
@@ -24,7 +24,6 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """
 
 
-"""This package integrates with Flask-Login. Determine who can use the app and which privileges they have."""
 from boac import db, std_commit
 from boac.models.base import Base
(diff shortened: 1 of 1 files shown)


boac/models/authorized_user.py
```diff
@@ -69,7 +71,10 @@ def get_id(self):
 
     @classmethod
     def find_by_uid(cls, uid):
-        return AuthorizedUser.query.filter_by(uid=uid).first()
+        """Supports Flask-Login via user_loader in routes.py."""
+        user = AuthorizedUser.query.filter_by(uid=uid).first()
+        std_commit()
```
johncrossman: Explicitly commit db session for extra measure.

PR #635, opened 2018-05-03
BOAC-861, curated-cohort-selector must behave across browsers, result-sets

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-861

Some added insurance to this feature:
* do not render `curated-cohort-selector` until students have loaded
* `event.stopPropagation()` in case of cross-browser differences
* put the `toggle` function on each student instance, to guarantee its presence

### DIFF
--- a/boac/static/app/group/groupsSelectorDirective.js
+++ b/boac/static/app/group/groupsSelectorDirective.js
@@ -52,29 +52,49 @@
 
         scope.selector = {
           selectAllCheckbox: false,
-          showGroupsMenu: false
+          showCuratedCohortMenu: false
         };
-
-        var init = function() {
-          var me = authService.getMe();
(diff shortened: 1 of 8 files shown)


boac/static/app/group/groupsSelectorDirective.js
```diff
@@ -123,35 +143,14 @@
...
-                // Break out of loop.
-                return false;
-              }
-            });
-            scope.selector.selectAllCheckbox = allStudentsSelected;
-          } else {
-            scope.selector.selectAllCheckbox = false;
-          }
-          updateShowGroupsMenu();
-        };
```
johncrossman: Function above is now attached to each student.

paulkerschen:
I'm persuaded!

PR #638, opened 2018-05-07
BOAC-872 Use shared conflict-check code for JSON cache inserts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-872


johncrossman:
A tolerant cache is a happy cache.

PR #639, opened 2018-05-08
BOAC-871 Falsey check for student term in course controller

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-871

Subtle and annoying gotcha: if a dict key exists but has `None` value, Python `get` will return `None` rather than using the provided fallback.


raydavis:
That'd do it, all right.

PR #642, opened 2018-05-10
BOAC-793, swap out page-views on the front-end; prep for days-since-course-site-visited metric

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-793

I expect we'll revisit the matrix view design. Next step: plug in real "days since" data.


paulkerschen:
No turning back!

PR #644, opened 2018-05-10
BOAC-881 Clarify which json_cache table is missing an entry

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-881

In general, we expect the runtime table to find all items while the staging table fills itself. At some point, we may even want to boost the logging level to WARN for runtime cache.

The PR also standardizes string formatting.


johncrossman:
Clearly clarified. Thanks @raydavis

PR #656, opened 2018-05-29
BOAC-866, reusable search directive; faster transition to isLoading

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-866

* The Jira has UX of reusable search box.
* The use of rootScope in `page.loading(...)` gives illusion of a more responsive tool. Such responsiveness is especially important when kicking off search.


raydavis:
Thanks for wrapping the new rootScope manipulation in a service, @johncrossman

PR #658, opened 2018-05-30
BOAC-866, page.loading(..) in one more spot

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-866


raydavis:
Out, darn spot!

PR #665, opened 2018-06-04
BOAC-917, skip pandas.DataFrame if DataLoch has no such enrollment

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-917

BOAC should not analyze bad/missing data. For the record, user described in BOAC-917 is, in fact, a member of course 1471428. Why no entry in `boac_analytics.course_enrollments` (nessie_qa)? Student was added recently, perhaps. Investigation to follow.

### DIFF
--- a/boac/lib/analytics.py
+++ b/boac/lib/analytics.py
@@ -93,6 +93,9 @@ def loch_student_analytics(canvas_user_id, canvas_course_id, term_id):
     if enrollments is None:
         _error = {'error': 'Unable to retrieve from Data Loch'}
         return {'currentScore': _error, 'lastActivity': _error}
+    if not next((e for e in enrollments if e.get('canvas_user_id') == canvas_user_id), None):
+        _error = {'error': f'According to Data Loch, Canvas user {canvas_user_id} is not enrolled in {canvas_course_id}'}
+        return {'currentScore': _error, 'lastActivity': _error}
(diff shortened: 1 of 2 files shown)


boac/lib/analytics.py
```diff
@@ -93,6 +93,9 @@ def loch_student_analytics(canvas_user_id, canvas_course_id, term_id):
     if enrollments is None:
         _error = {'error': 'Unable to retrieve from Data Loch'}
         return {'currentScore': _error, 'lastActivity': _error}
+    if not next((e for e in enrollments if e.get('canvas_user_id') == canvas_user_id), None):
+        _error = {'error': f'According to Data Loch, Canvas user {canvas_user_id} is not enrolled in {canvas_course_id}'}
```
raydavis: This is a misleading error message, probably due to the misleading variable name being used. Presumably Data Loch _does_ think the Canvas user is enrolled in the Canvas site, or we wouldn't even be in this loop. The unexpected bit is that it's not in the Loch's Canvas Course Scores table. See the error message in loch_assignments_submitted for a similar case.

raydavis: Also, if this short circuit goes in, it doesn't seem like we need the `if enrollments and student_row.empty` logic. (Or maybe you could replace that error response with your new one?)

johncrossman:
I will resubmit after views are restored and we agree on expected UX in such scenarios.

PR #667, opened 2018-06-05
BOAC-917 Null last_activity_at should be falsey

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-917

An incomplete test gave us too much confidence in that reasonable-looking code.


johncrossman:
Zero ain't None. Thanks @raydavis

PR #678, opened 2018-06-12
BOAC-943 Disentangle external cache load from application-specific table updates

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-943

A lot of files are touched here, but I hope the principle is clear enough: normalized cache load and alert load move up into `boac.api.cache_utils`, and tests come along for the ride.

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -37,6 +37,9 @@ from boac.models import json_cache
 from boac.models.alert import Alert
 from boac.models.job_progress import JobProgress
 from boac.models.json_cache import JsonCache
+from boac.models.normalized_cache_enrollment import NormalizedCacheEnrollment
+from boac.models.normalized_cache_student import NormalizedCacheStudent
+from boac.models.normalized_cache_student_major import NormalizedCacheStudentMajor
 from boac.models.student import Student
 from flask import current_app as app
 from sqlalchemy import and_, or_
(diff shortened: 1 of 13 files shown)


tests/test_api/test_cache_utils.py
```diff
@@ -0,0 +1,92 @@
...
+from boac.merged.sis_enrollments import merge_sis_enrollments_for_term
+from boac.models import json_cache
+from boac.models.alert import Alert
+from boac.models.normalized_cache_student import NormalizedCacheStudent
+from boac.models.normalized_cache_student_major import NormalizedCacheStudentMajor
+import pytest
+
+
+@pytest.mark.usefixtures('db_session')
+class TestCacheUtils:
```
raydavis: I know these are just relocations, but it still feels good to have a `test_cache_utils.py`.

paulkerschen:
Hm, there's a DB setup problem on Travis that I'm not getting locally. Let me try moving some import statements around.

boac/api/cache_utils.py
```diff
@@ -341,16 +349,50 @@ def load_analytics_for_sites(sites):
...
             success_count += 1
         else:
-            failures.append(f'merge_sis_profile failed for SID {sid}')
+            failures.append(f'merge_sis_profile failed for SID {csid}')
     return success_count, failures
+
+
+def load_normalized_cache(term_id, ids=None):
+    if not ids:
+        ids = get_all_student_ids()
```
paulkerschen: Travis is now at peace. Having `get_all_student_ids()` as a default right in the function signature was the problem, hence the change to a `None` check here and below.

raydavis: I've had issues with dynamic defaults in other ways, as well, so I'm just as glad Travis complained.

Looks good, and I'm eager to get my first merge conflict.

PR #679, opened 2018-06-12
NS-100 Move CalNet load up into cache_utils for easier relocation

raydavis:
Step one of https://jira.ets.berkeley.edu/jira/browse/NS-100

When the code first deploys, the CalNet merge will usually be redundant effort since it's also done in the Import ASC Students task. But because the entire long chain of student updates and refreshes will be refactored in this sprint, I don't expect it to stay redundant for long. If you prefer that I eliminate the duplication now, just let me know.

### DIFF
--- a/boac/merged/calnet.py
+++ b/boac/merged/calnet.py
@@ -23,9 +23,11 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-
+from boac import db, std_commit
 from boac.externals import calnet
 from boac.models.json_cache import stow
+from boac.models.student import Student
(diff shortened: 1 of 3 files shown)


boac/merged/calnet.py
```diff
@@ -37,3 +39,41 @@ def get_calnet_user_for_uid(app, uid):
...
+                m.first_name = new_first_name
+                m.last_name = new_last_name
+    return students
+
+
+def merge_student_calnet_data():
+    students = Student.query.all()
+    update_student_attributes(students)
+    app.logger.info(f'Modified {len(db.session.dirty)} student records from calnet')
+    std_commit()
```
paulkerschen: Thanks @raydavis - I don't mind the redundancy either while we're taking small steps. It seems Github would like an end-of-file line break, though.

raydavis: D'oh! I forgot to push my amended commit.

PR #687, opened 2018-06-18
NS-105 Move some payload tweaks up to controller packages

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-105

Aside from the improved tests, this isn't so much a clean-up as a getting-out-of-the-way. The top-level code is de-DRYed a bit, but I think it'll make it simpler to shuffle the middleware around.


paulkerschen:
The moving crew thanks you, @raydavis.

PR #688, opened 2018-06-19
BOAC-942 Remove unneeded python-path; overwrite duplicate WSGI config

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-942

BOAC version of https://github.com/ets-berkeley-edu/nessie/pull/88.


raydavis:
If it was in Nessie, it's gotta be good!

PR #689, opened 2018-06-20
NS-124, flake8-builtins to avoid builtin/variable name collision

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-124

A good rule to follow, methinks. Use `noqa: A003` to ignore.

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -114,8 +114,8 @@ def user_analytics(uid):
         if term_id == current_term_id:
             has_enrollments_in_current_term = True
         for enrollment in term['enrollments']:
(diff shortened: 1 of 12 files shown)


boac/api/user_controller.py
```diff
@@ -114,8 +114,8 @@ def user_analytics(uid):
         if term_id == current_term_id:
             has_enrollments_in_current_term = True
         for enrollment in term['enrollments']:
-            merge_analytics_for_user(enrollment['canvasSites'], uid, student.sid, canvas_id, term_id)
-        merge_analytics_for_user(term['unmatchedCanvasSites'], uid, student.sid, canvas_id, term_id)
```
johncrossman: Unused params removed.

PR #691, opened 2018-06-25
BOAC-982 Work around ASC API URL transition

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-982


johncrossman:
**kwite nice. Waiting on T

PR #695, opened 2018-06-26
BOAC-965, SQL script creates authorized_user then maps UID to College of Engineering

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-965


paulkerschen:
plpgsql, the language of love!

PR #697, opened 2018-06-27
BOAC-965, rename post-deploy SQL script

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-136 (was BOAC-965)

Let's not confuse Ops with poorly named scripts.


paulkerschen:
Inarguable!

PR #698, opened 2018-06-28
BOAC-987, coe_advisor_uid added to cohort_filter options; decorate in cf model moved to controller level

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-987

* Decoration in cohort_filter model puts undue work on simple operations (eg, delete) and tests. Ultimately, only controllers need students, alerts, etc. added to a cohort_filter instance.
* We will give COE advisors access to their students with read-only cohort_filters configured with criteria `coe_advisor_uid=12345`

### DIFF
--- a/boac/api/cohort_controller.py
+++ b/boac/api/cohort_controller.py
@@ -23,9 +23,8 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-
 from boac.api.errors import BadRequestError, ForbiddenRequestError, ResourceNotFoundError
-from boac.api.util import strip_analytics
+from boac.api.util import decorate_cohort, strip_analytics
 from boac.lib import util
 from boac.lib.berkeley import is_department_member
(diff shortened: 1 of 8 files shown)


boac/lib/analytics.py
```diff
@@ -213,3 +219,62 @@ def zscore(dataframe, value):
         return None
     else:
         return (value - dataframe.mean()) / dataframe.std(ddof=0)
+
+
+def decorate_cohort(cohort, order_by=None, offset=0, limit=50, include_students=True, include_alerts_for_uid=None):
```
paulkerschen: Makes sense, @johncrossman, but I wonder if `lib.analytics` is the right place for this method. The rest of this file is feed-building that we expect Nessie to take over, and I'd been assuming that the file would go away entirely post-Nessie cutover. Since it's decoration at the controller level, perhaps `api.util` would be a place for it?

johncrossman: That was my instinct, too, but dropping it in `api.util` creates a circular dependency: `student` depends on `api.util` depends on `student`.  I'll make this right by taking the `api.util` dependency out of student. Amendment coming soon.

johncrossman:
Amendment is in

PR #699, opened 2018-06-28
BOAC-990, cohort defined by coeAdvisorUid alone is read-only

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-986

And, only COE advisors can leverage the `coeAdvisorUid` criteria.


paulkerschen:
Thanks @johncrossman. Seems like these fiddly advisor-type-dependent rules are likely to keep proliferating, but at least we have them corralled at the controller level for now.

PR #700, opened 2018-06-28
BOAC-988, create read-only cohort when COE advisor first logs in

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-988


raydavis:
I'm gonna miss that UID....

PR #701, opened 2018-06-29
BOAC-993, COE advisor gets 404 for /cohorts/all

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-993


johncrossman:
@paulkerschen – I've just learned of some differing opinions (within PDG) w.r.t who can see what. I'm going to amend this PR with a TODO: figure it out.

PR #702, opened 2018-06-29
BOAC-993, advisor can view cohorts of dept colleagues; admin sees all

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-993

* `can_view_cohort` util used to filter results in controller
* Variable renaming touched many files and tests. With COE in the mix, readability is important.


raydavis:
The wires are tangled but color-coding them helps.

PR #705, opened 2018-07-03
BOAC-983 Allow a site to be associated with more than one class

raydavis:
... but only once per class.

https://jira.ets.berkeley.edu/jira/browse/BOAC-983

The modified code will soon be on the chopping block, but at least it verifies what needs to be changed on the Nessie side.


paulkerschen:
History records it.

PR #706, opened 2018-07-05
BOAC-995 Switch student data structures from database to loch-backed merged model

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-995

Here goes nothin'.

The core of the story is the switch from boac.models.athletics and boac.models.student to boac.merged.athletics and boac.merged.student. Most everything else follows from there, but just about everything on the back end turns out to be a collaborator in some way.

The hope is for changes to be undetectable on the front end except in the matter of performance, which is definitely in "this will get worse before it gets better" mode since we've lost just about all of our extant BOAC-side caching. The one added bit of caching covers membership counts for filtered cohorts: with students now in the loch, re-querying cohort membership on every request is unworkable.


raydavis:
@paulkerschen , this PR makes sense to me, caveats and all. But I don't think Nessie-Dev is currently scheduling jobs to maintain the new source tables. If that's so, then wouldn't deploying the new client-side code freeze BOAC-Dev?

paulkerschen:
@raydavis Yes, quite. Given size and scope I wanted to get this out under the public eye (and am also curious how it behaves on dev) but I’m fine with holding off merge until the new Nessie jobs are running.

paulkerschen:
https://github.com/ets-berkeley-edu/nessie/pull/114 contains what I hope is the minimum to set things in order on nessie-dev.

raydavis:
Look out below!

PR #709, opened 2018-07-13
BOAC-1011 Cascade delete for curated cohort memberships

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1011


raydavis:
That was gonna be my guess too.

PR #711, opened 2018-07-16
[BOAC-979] Custom VPC security groups and SSH restriction

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-979

Previously tested off of `boac-etsla-ops` branch.

### DIFF
--- a/.ebextensions/00_ami.config
+++ b/.ebextensions/00_ami.config
@@ -10,7 +10,7 @@ packages:
 
 option_settings:
   aws:autoscaling:launchconfiguration:
-    SSHSourceRestriction: tcp, 22, 22, ETS-SSH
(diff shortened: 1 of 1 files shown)


buildspec.yml
```diff
@@ -1,7 +1,7 @@
 version: 0.2
 
 eb_codebuild_settings:
-  CodeBuildServiceRole: arn:aws:iam::697877139013:role/service-role/code-build-BOAC-service-role
+  CodeBuildServiceRole: arn:aws:iam::697877139013:role/service-role/code-build-boac-etsla-service-role
```
johncrossman: Ops does not want role name change. I'll amend.

PR #715, opened 2018-07-17
BOAC-1021 Add course-specific matrix to class page

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1021

Did some cleanup along the way...


paulkerschen:
Thanks for catching this missed boat, @raydavis.

PR #718, opened 2018-07-18
BOAC-1001 Debug logging for Redshift queries

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1001

This ought to give log watchers something to look at.


johncrossman:
When it comes to logs, I like to watch.

PR #720, opened 2018-07-18
BOAC-1031 Null-first sort ordering

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1031


johncrossman:
Never not nulls last.

PR #722, opened 2018-07-18
BOAC-1036, admin users in non-prod get link to 'Admin' in top right menu

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1036

A link that is quick.


paulkerschen:
Zippy.

PR #724, opened 2018-07-19
BOAC-989, advisorLdapUid checkbox label will include name if uid is not me.uid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-989


paulkerschen:
Thanks, @johncrossman. Waiting for Judge T to bang his gavel...

PR #729, opened 2018-07-19
BOAC-1017, /profile/my feed should not exclude my_groups.students info

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1017

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -47,7 +47,7 @@ def my_profile():
         # All BOAC views require group and cohort lists
         authorized_user_id = current_user.id
         groups = StudentGroup.get_groups_by_owner_id(authorized_user_id)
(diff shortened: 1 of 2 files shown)


boac/api/user_controller.py
```diff
@@ -47,7 +47,7 @@ def my_profile():
         # All BOAC views require group and cohort lists
         authorized_user_id = current_user.id
         groups = StudentGroup.get_groups_by_owner_id(authorized_user_id)
-        groups = [g.to_api_json(include_students=False) for g in groups]
+        groups = [g.to_api_json() for g in groups]
```
paulkerschen: @johncrossman I think this needs the `sids_only=True` kwarg: otherwise it will call out to the loch for student profiles on every page load.

paulkerschen:
Merging from seat 17B pre-takeoff.

PR #733, opened 2018-07-19
BOAC-1010, unit totals for non-ASC; Degree Progress 'no data' when no data

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1010


paulkerschen:
Thanks @johncrossman - amazing how much code we’re still turning up that assumes everyone’s an athlete. Mens sana in corpore sano.

PR #735, opened 2018-07-20
BOAC-1027 "Naturalize" ordering of strings in SQL queries

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1027

Ignore case differences and punctuation: "MCB" before "Mechanical" before "ME/NE" before "Me - You = Lonely".


johncrossman:
Naturally.

PR #738, opened 2018-07-24
BOAC-1053, consistent pattern for managing checkboxes on /cohort search

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1053

`cohortController.js` is too complex and the problem worsened with `advisorLdapUid` logic. The code is more manageable when all checkboxes follow a similar pattern.


paulkerschen:
Thanks for providing some order, @johncrossman.

PR #740, opened 2018-07-24
BOAC-1058, only ASC and admin can reach /teams

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1058


paulkerschen:
Unauthorized users can go check the official calbears.com roster like everyone else, thank you very much.

PR #743, opened 2018-07-25
BOAC-1058, /api/team_groups/all returns empty list for unauthorized users

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1058

The one and only call to the `/api/team_groups/all` feed is baked into the `cohortController.js` logic. Although team data does not surface to COE advisors, they are being impacted by the 404. The safest fix, given our release schedule, is to return an empty list to unauthorized users and keep the response.status at 200.


paulkerschen:
It's an empty box, but it's wrapped just like a real Christmas present.

PR #744, opened 2018-07-25
BOAC-1022 Content area height fix for Firefox

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1022

As far as I can tell this fixes the issue without breaking anything else, but you know, CSS.


johncrossman:
It's just browsers being browsers.

PR #746, opened 2018-07-25
BOAC-1048 SID is ordering field of last resort

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1048


johncrossman:
Christmas came early for QA.

PR #748, opened 2018-07-25
BOAC-1061, ASC users see inactives on class page and in custom cohort only

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1061

None is sometimes None and sometimes False. Confusing requirements is True.


paulkerschen:
Three-valued logic is a rock on which the finest minds have foundered, but this gets us over the hump. Thanks @johncrossman.

PR #753, opened 2018-07-26
BOAC-1064, allow admin to create cohort with non-None advisor_ldap_uid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1064


paulkerschen:
To administrate you must create.

PR #756, opened 2018-07-26
[release] BOAC-1065, SSHSourceRestriction change also needed for prod VPC move

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1065

Another necessary config tweak. Gives me the change to bump version.


johncrossman:
Expect this `00_config` file to be an exact match of what we have in qa.

PR #760, opened 2018-07-27
BOAC-1079, git mv front-end cohort-related files to prep for new filtered view

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1079

No functional changes here. Front-end dir structure, routes and naming should match the conceptual structure of cohorts. More to come...


raydavis:
But what will we do with all that useless historical knowledge?

boac/static/app/routes.js
```diff
@@ -181,7 +187,7 @@
         resolve: resolveSplash
       })
       .state('teams', {
-        url: '/cohorts/teams',
+        url: '/teams',
```
johncrossman: @pfarestveit Another point of interest for Teena: URL paths have changed (see above) on master branch. Eg, `/groups/manage` is now `/cohort/curated/manage`.

PR #762, opened 2018-07-30
BOAC-1069 Include no-units-yet students in the 'Units 0 - 29' filter

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1069

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -459,18 +459,25 @@ def numrange_to_sql(column, numrange):
     numrange_match = numrange_syntax.match(numrange)
     if numrange_match:
         bounds = []
+        include_null = False
         if numrange_match[1] != 'NULL':
             lower_bound_condition = '>'
             # Square brackets in numrange syntax indicate an inclusive range.
-            if numrange_match[3] == '[' and numrange_match[1] != '0':
+            if numrange_match[3] == '[':
                 lower_bound_condition += '='
             bounds.append(f'{column} {lower_bound_condition} {numrange_match[1]}')
+        else:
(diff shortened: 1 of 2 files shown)


boac/externals/data_loch.py
```diff
@@ -459,18 +459,25 @@ def numrange_to_sql(column, numrange):
     numrange_match = numrange_syntax.match(numrange)
     if numrange_match:
         bounds = []
+        include_null = False
         if numrange_match[1] != 'NULL':
             lower_bound_condition = '>'
             # Square brackets in numrange syntax indicate an inclusive range.
-            if numrange_match[3] == '[' and numrange_match[1] != '0':
```
raydavis: So far as I can tell from Git history, this odd exception crept in via a misinterpretation of earlier code. At any rate, it's easy enough _not_ to do a zero-inclusive filter: just specify '(' instead of '['.

johncrossman:
Nicely finessed, @raydavis

boac/externals/data_loch.py
```diff
@@ -459,18 +459,25 @@ def numrange_to_sql(column, numrange):
...
+            if numrange_match[3] == '[':
                 lower_bound_condition += '='
             bounds.append(f'{column} {lower_bound_condition} {numrange_match[1]}')
+        else:
+            # An inclusive lower bound of NULL is interpreted as including null values.
+            include_null = (numrange_match[3] == '[')
```
raydavis: This no longer matches the Postgres interpretation of `numrange`, so feel free to request amendment. My alternative would be to pass a different parsable range-specification-format in the request parameter. E.g., `RANGE(NULL, 30, '[)')`

PR #770, opened 2018-08-01
[QA] BOAC-1091, disable Matrix view if student count exceeds threshold

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1091

Details in the Jira.


paulkerschen:
It's not _that_ kind of all-swallowing matrix.

PR #779, opened 2018-08-02
BOAC-1089 Missing photos should not generate 404s

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1089

The background image will display instead.


paulkerschen:
Everyone's browser console thanks you, @raydavis.

PR #782, opened 2018-08-03
BOAC-1123, BOAC-1122, eslint-angular rules; many style fixes; rename() in new cohort controller

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1122
https://jira.ets.berkeley.edu/jira/browse/BOAC-1123

Many files touched here. The eslint/angular rules are useful but notice the `angular/di-order` rule: correct dependency-injection order is, for example, `($http, utilService)`  and not `(utilService, $http)`. Acceptable?

### DIFF
--- a/boac/static/app/bootstrap.js
+++ b/boac/static/app/bootstrap.js
@@ -26,8 +26,8 @@
 (function(angular) {
 
   var boac = angular.module('boac', [
-    'base64',
     'ngAnimate',
+    'base64',
     'ui.bootstrap',
     'ui.router'
   ]);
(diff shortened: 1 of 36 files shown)


boac/static/app/bootstrap.js
```diff
@@ -26,8 +26,8 @@
 (function(angular) {
 
   var boac = angular.module('boac', [
-    'base64',
     'ngAnimate',
+    'base64',
```
johncrossman: The `angular/module-dependency-order` rule complained: "ngAnimate is a standard module and should be sorted before base64".

raydavis:
Sometimes these ordering rules drive me crazy, I admit, but we might as well give 'em a shot!

PR #785, opened 2018-08-08
BOAC-1140, expose getMajors in filterCriteriaService

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1140


paulkerschen:
Majors: exposed.

PR #787, opened 2018-08-08
BOAC-1072, add unsaved filter to working set of filterCriteria

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1072

More work on adding unsaved-filter to criteria. Clean up some naming and enrich filter definition object.


paulkerschen:
Filter that fish tank!

PR #788, opened 2018-08-09
[QA] BOAC-114, create_my_students_cohort_per_uid script

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1147

For example:
```
python3 scripts/db/migrate/20180809-BOAC-1147/create_my_students_cohort_per_uid.py 12345 Jane
```

Much safer/easier than composing raw SQL.

### DIFF
--- a/scripts/db/migrate/20180809-BOAC-1147/create_my_students_cohort_per_uid.py
+++ b/scripts/db/migrate/20180809-BOAC-1147/create_my_students_cohort_per_uid.py
@@ -1,0 +1,55 @@
+"""
+Copyright ©2018. The Regents of the University of California (Regents). All Rights Reserved.
+
+Permission to use, copy, modify, and distribute this software and its documentation
+for educational, research, and not-for-profit purposes, without fee and without a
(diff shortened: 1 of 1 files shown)


scripts/db/migrate/20180809-BOAC-1147/create_my_students_cohort_per_uid.py
```diff
@@ -0,0 +1,55 @@
...
+    if uid:
+        label = f'{first_name}\'s Students' if first_name else 'My Students'
+        cohort = CohortFilter.create(
+            uid=uid,
+            label=label,
+            advisor_ldap_uid=uid,
+        )
+        print(f'Cohort created: {cohort}')
+    else:
+        print(f'ERROR: Failed to create cohort')
```
paulkerschen: @johncrossman f-string with no interpolation isn't a merge blocker (I do it all the time) but I wonder if there's a linter rule to catch it.

paulkerschen:
Looks right to me!

PR #789, opened 2018-08-09
BOAC-1146, populate 'Advisor' filter with data of /api/profiles/dept/{dept_code}

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1146


paulkerschen:
Look forward to seeing this in action, @johncrossman... looks good to me once lint-js is tamed.

PR #791, opened 2018-08-10
BOAC-1147, create_my_students_cohort_per_uid script

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1147

QA PRs: #788 and #790


paulkerschen:
@johncrossman Since this is going into master, should we remove [QA] from the commit messages to avoid confusion in the history?

johncrossman:
@paulkerschen Yes, good idea. Edit coming soon.

PR #794, opened 2018-08-13
BOAC-1124, execute search, disable options and query-args per filters

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1124

* Wire 'Apply' button to exec search function.
* Disable available options when added to criteria.
* Event listeners


paulkerschen:
Thanks @johncrossman.. that's a whole lotta front end.

PR #796, opened 2018-08-14
BOAC-1078, No phone? No icon.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1078


paulkerschen:
What if they have a fax number?

PR #797, opened 2018-08-14
BOAC-1157 Enforce current term cutoff in per-student enrollments query

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1157


johncrossman:
You've been a bad boy, BOAC.

PR #795, opened 2018-08-14
BOAC-1124, student photo and sis data in new cohort view

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1124
https://jira.ets.berkeley.edu/jira/browse/BOAC-1155

...and tabs and pagination.

PR #798, opened 2018-08-14
BOAC-1149, legacy select-majors-filter belongs in legacy controller

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1149

The BOAC-1149 fix was in PR #795 and yet while verifying I discovered `$scope` usage in `filterCriteriaService`. This puts that bit of legacy code back in legacy controller.


paulkerschen:
Sneaky.

PR #799, opened 2018-08-14
BOAC-1145, do not omit ASC team-groups with zero active students

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1145
https://jira.ets.berkeley.edu/jira/browse/BOAC-1158

No impact on ASC's exclude-inactives-in-search requirement but it does bring Big (Junk) Data to the "Teams" dropdown.


paulkerschen:
Batteries now included.

PR #800, opened 2018-08-14
BOAC-1155, two more sidebar links get null args in filteredCohort() ref

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1155

### DIFF
--- a/boac/static/app/home/home.html
+++ b/boac/static/app/home/home.html
@@ -38,7 +38,7 @@
         <h1 class="page-section-header">Filtered Cohorts</h1>
         <div class="home-cohort-actions">
           <a id="home-filtered-cohorts-create-link"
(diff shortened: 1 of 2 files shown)


boac/static/app/home/home.html
```diff
@@ -38,7 +38,7 @@ <h2 class="page-section-header-sub">
         <h1 class="page-section-header">Filtered Cohorts</h1>
         <div class="home-cohort-actions">
           <a id="home-filtered-cohorts-create-link"
-             data-ui-sref="filteredCohort({c: 'new', i: null, v: null})"
+             data-ui-sref="filteredCohort({c: 'new'})"
```
johncrossman: The reset-args-to-null is not need when linking from home.

PR #808, opened 2018-08-16
BOAC-1161, exclude ASC/COE filters based on user role and then sort

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1161

Wrap `_.map()` in `_.reject()` so unavailable filters do not surface as `undefined`.


paulkerschen:
Wrap it and map it.

PR #809, opened 2018-08-20
BOAC-1163, tab to matrix view with proper location.args

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1163
https://jira.ets.berkeley.edu/jira/browse/BOAC-1116

* `ng-show` allows matrix view render while `isLoading`
* `anchorSchool` (to top) when paging
* the odd `studentCountExceedsMatrixThreshold`


paulkerschen:
I can dig it.

PR #813, opened 2018-08-22
BOAC-1122, save cohort without reloading search results

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1122

Passing a set of "callbacks" to `filterCriteriaComponent` seems cleanest.


paulkerschen:
+1 on those callbacks.

PR #816, opened 2018-08-22
BOAC-1177, toss out the COE-read-only-cohort feature

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1177

Today is all about feature deletion.


paulkerschen:
Dump the ballast!

PR #817, opened 2018-08-22
BOAC-1029, What's better than pills? Bigger pills.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1029


paulkerschen:
No longer 2 small 4 numbers!

PR #822, opened 2018-08-23
BOAC-1166, add COE PREP as search option in new filtered-cohort

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1166

Involved taming the complexity in `data_loch.get_students_query()`

### DIFF
--- a/boac/api/student_controller.py
+++ b/boac/api/student_controller.py
@@ -40,6 +40,7 @@ from flask_login import current_user, login_required
 def get_students():
     params = request.get_json()
     advisor_ldap_uids = util.get(params, 'advisorLdapUids')
+    coe_prep_statuses = util.get(params, 'coePrepStatuses')
     gpa_ranges = util.get(params, 'gpaRanges')
     group_codes = util.get(params, 'groupCodes')
     levels = util.get(params, 'levels')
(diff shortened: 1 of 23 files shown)


boac/externals/data_loch.py
```diff
@@ -199,6 +199,14 @@ def get_athletics_profiles(sids):
     return safe_execute_redshift(sql, sids=sids)
 
 
+def get_coe_profiles(sids):
+    sql = f"""SELECT sid, gender, ethnicity, minority, did_prep, prep_eligible, did_tprep, tprep_eligible
+        FROM {coe_schema()}.students
+        WHERE sid = ANY(:sids)
+        """
+    return safe_execute_redshift(sql, sids=sids)
```
paulkerschen: @johncrossman Not that it matters as much for this schema, but the pattern we've been using to date for these profile fetches is to try and minimize Redshift and compute burden by fetching a pre-built JSON feed containing all these attributes, as in `get_asc_profiles` above. The loch has a `boac_advising_coe.student_profiles` table ready to serve.

boac/api/student_controller.py
```diff
@@ -68,6 +70,8 @@ def get_students():
         offset=offset,
         limit=limit,
     )
+    if results is None:
```
raydavis: Could I get a pointer to the `results = None` path, @johncrossman ? I'm having a hard time plucking that thread from the tangle.

johncrossman: _data_loch._safe_execute_ has:
```
    try:
        ...execute query...
    except sqlalchemy.exc.SQLAlchemyError as err:
        app.logger.error(f'SQL {s} threw {err}')
        return None
```
which then bubbles up through `student.query_students`. This could result from, for example, a COE user requesting ASC attributes. We could add validation at controller level, which I think is a good amendment for this PR.

johncrossman: @raydavis @paulkerschen Amendment is in. And I added validation in '/api/students'.

raydavis:
Very nice amendments, @johncrossman -- thanks!

boac/api/student_controller.py
paulkerschen: @raydavis @johncrossman If I'm remembering right, the way this works is that at the top of `boac.merged.student.query_students` there's a call to `narrow_scope_by_criteria`, which constrains the result set in the case of ASC or COE criteria being specified, and will return an empty array for `scope` in the case of someone requesting criteria they're not allowed to see. The empty `scope` array will then cause `results` to come back `None`.

We could make it clearer by adding an explicit short-circuit to `boac.merged.student.query_students` and `boac.merged.student.search_for_students` in the case of empty scope, but I agree, no reason to hold up this PR.

PR #823, opened 2018-08-23
BOAC-1180, step one in delete legacy filtered-cohort code

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1180

Move the valuable goods out of `boac/static/app/cohort/filtered` (legacy feature).


paulkerschen:
I'll underscore that.

PR #824, opened 2018-08-23
BOAC-1094 Alerts for no course site activity, configurable cutoffs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1094

All the important numbers are in configs because we have no idea how the third week will look until it hits us.


johncrossman:
Very nice. Let's take it for a test drive.

PR #825, opened 2018-08-23
BOAC-1180, delete unused /api/teams; consolidate student-related api code

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1180


paulkerschen:
Satisfying! And glad to lose `user` as a term for people who don't use the app.

PR #832, opened 2018-08-24
BOAC-1095 Infrequent bCourses activity alerts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1095


johncrossman:
Son, you are frequently infrequent.

PR #836, opened 2018-08-27
BOAC-1189, fix pagination; sort ethnicities and advisors

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1189


boac/static/app/cohort/filtered/filterCriteria.html
```diff
@@ -47,7 +47,6 @@
             <li role="menuitem"
                 data-ng-click="onDraftFilterClick(option)"
                 data-ng-class="{'divider': option === null, 'disabled': option.disabled}"
-                data-ng-disabled="option.disabled"
```
johncrossman: Does nothing. The action is in anchor tag below.

PR #838, opened 2018-08-27
BOAC-1066 Model/controller refactor for filtered cohorts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1066

`current_user` logic stays in the controller, everything else moves to a `to_api_json` method in the model.


johncrossman:
Good to see `test_api_util.py` retired. Thx @paulkerschen

PR #843, opened 2018-08-28
BOAC-1201, simplify redrawButtons in filterCriteriaComponent

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1201
https://jira.ets.berkeley.edu/jira/browse/BOAC-1188

Relying on listeners to redraw all buttons risks strange and unwanted show/hide behavior.

This includes removal of 'Manage Cohorts' from home and flex-width tweak in added-filter rows.

### DIFF
--- a/boac/static/app/cohort/cohort.css
+++ b/boac/static/app/cohort/cohort.css
@@ -116,7 +116,7 @@
 }
 
 .cohort-filter-draft-column-02 {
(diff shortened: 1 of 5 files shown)


paulkerschen:
Listeners aren't always good listeners. Thanks @johncrossman.

PR #844, opened 2018-08-28
BOAC-1190, fix Teams page and remove its inaccurate totalStudentCount

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1190

`totalStudentCount` will be inaccurate for Football because a SID might be double counted if student is on two squads. IMO, fixing this is not worth the effort.


paulkerschen:
Pragmatic.

PR #845, opened 2018-08-28
BOAC-1199, remove stray and unnecessary uib-pagination

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1199

Ironically, the `uib-pagination` element intended for only list-view had a faulty `ng-if` and never rendered. The second `uib-pagination` (how'd we get two?!) was showing up on both list and matrix view. I fixed the former and tossed out the latter.


paulkerschen:
Whoops! That's got to help.

PR #847, opened 2018-08-28
BOAC-1139, voice pagination; better show/hide for save button; various accessibility

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1203
https://jira.ets.berkeley.edu/jira/browse/BOAC-1139

Enhance `uib-pagination` accessibility-wise. IMO, customizing bootstrap template is safe because that project is "feature-complete and no longer being maintained". They've moved on to Angular 2.

Plus, aria attributes here and there.

### DIFF
--- a/boac/static/app/cohort/filtered/filterCriteriaComponent.js
+++ b/boac/static/app/cohort/filtered/filterCriteriaComponent.js
@@ -177,7 +177,7 @@
           // We set 'show' to true when user clicks Apply. We set 'show' to false if filters are altered.
           if (isInitPhase && _.size($scope.filters.added)) {
             $scope.buttons.saveButton.show = true;
-          } else if (_.isEmpty($scope.filters.added)) {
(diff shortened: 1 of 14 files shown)


boac/static/app/cohort/filtered/filterCriteriaComponent.js
```diff
@@ -177,7 +177,7 @@
           // We set 'show' to true when user clicks Apply. We set 'show' to false if filters are altered.
           if (isInitPhase && _.size($scope.filters.added)) {
             $scope.buttons.saveButton.show = true;
-          } else if (_.isEmpty($scope.filters.added)) {
+          } else if (_.isEmpty($scope.filters.added) || _.get($scope.filters.draft, 'key')) {
             $scope.buttons.saveButton.show = false;
```
johncrossman: If user is in process of adding a new filter then we hide Save button.

paulkerschen:
Nice customization, @johncrossman. Angular 1 was good enough for Grampa.

PR #849, opened 2018-08-28
BOAC-1204, fix links to inactive and intensive from sidebar

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1204

Params that are explicit in `routes.js` are readable and matched in sidebar.


raydavis:
I like words.

PR #850, opened 2018-08-28
BOAC-1207 Fix Alerts code for sites with no activity

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1207


paulkerschen:
Thanks for the variant text in the fixture, @raydavis.

PR #851, opened 2018-08-28
BOAC-1040 Collapsible cohorts on homepage

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1040

Accordions.


johncrossman:
Accordions it is.

PR #859, opened 2018-08-30
BOAC-1209, order_by (eg, ASC team) can influence query scope

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1209

See latest comment in the Jira.

### DIFF
--- a/boac/merged/student.py
+++ b/boac/merged/student.py
@@ -198,16 +198,19 @@ def query_students(
         sids_only=False,
         unit_ranges=None,
 ):
-    scope = narrow_scope_by_criteria(
-        get_student_query_scope(),
-        advisor_ldap_uids=advisor_ldap_uids,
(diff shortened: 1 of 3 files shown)


boac/merged/student.py
```diff
@@ -198,16 +198,19 @@ def query_students(
...
+        'ethnicities': ethnicities,
+        'genders': genders,
+        'group_codes': group_codes,
+        'in_intensive_cohort': in_intensive_cohort,
+        'is_active_asc': is_active_asc,
+    }
+    if order_by:
+        # 'order_by' value might influence query scope
+        criteria_.update({order_by: True})
+    scope = narrow_scope_by_criteria(get_student_query_scope(), **criteria_)
```
paulkerschen: @johncrossman Ah, clever solution. But why `criteria_` with an underscore? I don't see another variable it's colliding with.

PR #860, opened 2018-08-30
BOAC-1213, Safari and Firefox need explicit button background-color

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1213


raydavis:
Pfff

PR #863, opened 2018-09-04
BOAC-1045 More sorting on front page

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1045


raydavis:
If it moves, sort it.

PR #870, opened 2018-09-10
BOAC-1230, filter cohort by range of last_name initials

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1230

Some CoE advisors are assigned students according to last initial. This filter is our first of type 'range' and the corresponding UX is new. Getting it done early in the sprint gives Teena time to catch up.


johncrossman:
Travis passed on this PR (see https://travis-ci.org/ets-berkeley-edu/boac/builds/426869813) but Github has no idea. Our Github/Travis integration seems to be busted. I'll Jira.

PR #874, opened 2018-09-12
BOAC-1051 An empty matrix is not a satisfying UX

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1051


raydavis:
"No student data" is good student data.

PR #876, opened 2018-09-13
BOAC-797 Display alerts for SIS holds

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-797


johncrossman:
Looks good. Thanks @paulkerschen

PR #883, opened 2018-09-17
BOAC-1256, /admin gets demo-mode toggle and users organized by dept

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1256


paulkerschen:
`admin-accordion` is the boss accordion.

PR #884, opened 2018-09-18
BOAC-1263, delete-cohort modal fix; save-cohort post op fix

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1263
https://jira.ets.berkeley.edu/jira/browse/BOAC-1251

Also, removed pointless listener in homeController.


paulkerschen:
Where would we be without Teena?

PR #887, opened 2018-09-19
BOAC-1253 Disable zoom on matrix view

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1253


raydavis:
It was fun while it lasted.

PR #892, opened 2018-09-20
BOAC-1255, alert-count per student on course page; test coverage

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1255

Reusability is the word that you heard.


paulkerschen:
They might be cool kids, but they still have to turn in their homework.

PR #895, opened 2018-09-20
BOAC-1269, 'Cancel' button must show after initial filter category is selected

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1269

As Teena expects.


paulkerschen:
You can't outsmart Teena.

PR #897, opened 2018-09-21
BOAC-1237 Today becomes yesterday at midnight browser time

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1237


johncrossman:
Life according to browser time.

PR #899, opened 2018-09-21
BOAC-1271, click cancel then not isSaving; ids for Teena

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1271

An `ets-qa` PR will follow.


paulkerschen:
Settle down, spinner.

PR #901, opened 2018-09-21
BOAC-1271, noop callback for rename cohort; create cohort callback fix

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1271

The previous PR done broke the create cohort callback.


paulkerschen:
Give it a noop!

PR #902, opened 2018-09-23
BOAC-1275, other people's cohorts must be read-only; no option to Save-As

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1275

...and simplified `acknowledgeSave` on button.

### DIFF
--- a/boac/static/app/cohort/filtered/filterCriteriaComponent.js
+++ b/boac/static/app/cohort/filtered/filterCriteriaComponent.js
@@ -210,7 +210,6 @@
     };
 
     var onSuccessfulCohortSave = function(cohort) {
-      $scope.isSaving = false;
       // If user clicked 'Cancel' then do nothing.
       if (cohort) {
         $scope.search.cohort = cohort;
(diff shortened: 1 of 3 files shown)


boac/static/app/cohort/filtered/filterCriteriaComponent.js
```diff
@@ -275,8 +273,9 @@
     this.$onInit = function() {
       var filterCategories = _.clone(this.filterCategories);
       var filterCriteria = _.clone(this.search.filterCriteria);
+      var isOwnedByCurrentUser = _.get(this.search.cohort, 'isOwnedByCurrentUser');
 
-      $scope.allowEdits = _.get(_.pick(this.search.cohort, _.identity), 'isOwnedByCurrentUser', true);
+      $scope.allowEdits = _.isNil(isOwnedByCurrentUser) || cohortUtils.lenientBoolean(isOwnedByCurrentUser);
```
raydavis: This line is counterintuitive enough in isolation, and far enough away from knowledge that the filter is currently in a "draft" state and the user will be presented with a "save as" rather than "save" UX, that a comment might save future devs some detective work.

Alternatively, I'm just obtuse. :)

raydavis:
Looks good to me, @johncrossman -- just had the documentation suggestion, which is fine to disagree with.

boac/static/app/cohort/filtered/filterCriteriaComponent.js
johncrossman: I agree, @raydavis. Amending.

raydavis:
Thanks, @johncrossman !

PR #904, opened 2018-09-24
BOAC-1280, register 'minority' as COENG db column

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1280


paulkerschen:
Thanks for untangling, @johncrossman.

PR #908, opened 2018-09-24
BOAC-481, user-id with each googleAnalyticsService.track

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-481

I'm still working to understand Google Analytics dashboard. However, this seems to be the recommended pattern (see link in Jira) for session tracking.


paulkerschen:
Analytics about the analytics are the absolutely astounding analytics.

PR #909, opened 2018-09-25
BOAC-1286, no 'Score' label for assignments submitted

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1286


paulkerschen:
No score, Junior.

PR #911, opened 2018-09-25
BOAC-529 Per-term unit totals on student profile

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-529


johncrossman:
Those darn accordions!

PR #912, opened 2018-09-26
BOAC-1290 Imaginary entities should not be clickable in matrix

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1290

QA PR will follow a successful review...


paulkerschen:
Thanks, @raydavis!

PR #916, opened 2018-09-26
BOAC-1297, popup error message if invalid edit on an added cohort filter

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1297

image


paulkerschen:
To change the range, you must be well arranged.

PR #922, opened 2018-09-27
BOAC-1299 Zero-percentiles are still legitimate data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1299

To venture onto BOAC-1272's turf, most of the examples I've found have greater-than-zero rounded-up percentiles.


paulkerschen:
Yes indeed.

PR #927, opened 2018-09-28
BOAC-1303, one students is wrongs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1303


raydavis:
Ah, look at all the lonely student.

PR #928, opened 2018-09-28
BOAC-1309, self-signed SSL certificate for BOAC

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1309


raydavis:
Riveting!

PR #929, opened 2018-10-01
BOAC-1310, no googleAnalytics without 'ga'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1310

Notice the `asycn` in `index.html`:
```
<script async src='https://www.google-analytics.com/analytics.js'></script>
```
Do not use 'ga' until the async is done.


paulkerschen:
I do nothing without `id` and `ga`.

PR #934, opened 2018-10-02
BOAC-1309, SSL private key location (S3 URL) changed

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1309


johncrossman:
Do not merge yet.  I need to verify that I have the right URL here.

johncrossman:
Okay, the S3 URL is correct. We can merge when green.

PR #937, opened 2018-10-10
BOAC-529 Add breathing room between semesters

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-529


johncrossman:
What a diff 10px makes.

PR #938, opened 2018-10-10
BOAC-1324, no alerts for DeCal courses (directed/group study)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1324

A query of `intermediate.course_sections` (Redshift) view for 'GRP' sections and catalog_id pattern (below) returns only 'group study' and 'directed study' sections. I believe this fits with the requirement. A second pair of eyes is welcome.

Use `?w=1` for an easier diff read.

### DIFF
--- a/boac/lib/berkeley.py
+++ b/boac/lib/berkeley.py
@@ -271,3 +271,12 @@ def can_view_cohort(user, cohort):
     if cohort.owners:
         cohort_dept_codes = np.concatenate([get_dept_codes(o) for o in cohort.owners])
     return np.in1d(my_dept_codes, cohort_dept_codes)
+
+
+def section_is_eligible_for_alerts(enrollment, section):
+    if section.get('component') == 'LEC':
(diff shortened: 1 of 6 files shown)


paulkerschen:
@johncrossman Makes sense! BOAC also has some alerts for 98BC and 198BC courses (Berkeley Connect), which I would guess are likewise noise for advisors; should we screen them out too?

paulkerschen:
Also, on the GRP filter: it looks to me like 99 and 199 sections are never GRP, but usually IND and occasionally TUT. I'm guessing we should screen those out too. The 98 and 198 sections have a wider variety of instruction formats, but the courses titles all seem to be something on the order "Group Study". I wonder if we're better off ignoring instruction format and just taking catalog id as our guide?

johncrossman:
Nice sleuthing, @paulkerschen  PR amended; better catalog_id pattern match is in.

PR #941, opened 2018-10-12
BOAC-1335 Styling fixes for GPA

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1335

Charts that size to the window! Three count them three decimal places!


raydavis:
Four would definitely be _de trop._

PR #944, opened 2018-10-15
BOAC-1338 Tolerate missing graduation term in front end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1338


raydavis:
Anything's likely to go missing sometime.

PR #946, opened 2018-10-15
BOAC-1056: re-'names' filtered cohort label field

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1056

Will submit a follow-up PR with front-end changes.


johncrossman:
Nicely done, @lyttam  On to the front-end!

PR #947, opened 2018-10-15
BOAC-1152 Deal with placeholder units in EAP classes

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1152


paulkerschen:
Thanks for handling those gargantuan unit values, @raydavis.

PR #948, opened 2018-10-15
BOAC-1056: Renames filtered cohort element on front-end

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1056

Follow-up to #946


johncrossman:
Thank you @lyttam  Call that cohort by its name!

PR #949, opened 2018-10-16
BOAC-846 Adjust units/GPA alignment, decimal precision

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-846


johncrossman:
Precision may vary. Thanks @paulkerschen

PR #951, opened 2018-10-18
BOAC-134, alert_count per filtered cohort, like student_count, is stored in db

johncrossman:

Better performance on the front-end relies on caching `alert_count`


boac/api/cache_utils.py
```diff
@@ -225,8 +225,10 @@ def load_filtered_cohort_counts():
     for cohort in CohortFilter.query.all():
         # Remove!
         cohort.update_student_count(None)
+        cohort.update_alert_count(None)
         # Reload!
-        cohort.to_api_json(include_students=False)
+        owner_id = cohort.owners[0].id if len(cohort.owners) else None
```
johncrossman: The db schema supports multiple cohort owners but in the real world it is one owner per cohort.

boac/api/filtered_cohort_controller.py
```diff
@@ -104,7 +104,7 @@ def get_cohort(cohort_id):
             order_by=order_by,
             offset=int(offset),
             limit=int(limit),
-            include_alerts_for_uid=current_user.uid,
+            include_alerts_for_user_id=current_user.id,
```
johncrossman: Switching from `uid` to `user_id` simplifies the code a bit and means less work for `cache_utils` refresh.

boac/api/cache_utils.py
paulkerschen: Makes sense. Is that wrinkle maybe worth a comment for our future selves?

johncrossman: Future me agrees. Adding comment now.

PR #953, opened 2018-10-20
BOAC-1340, improve homepage performance by reducing queries

johncrossman:

My-cohorts in sidebar and on homepage are served from the same lightweight user profile data. Student lists are lazy loaded. Homepage _should_ more quickly load.


raydavis:
Reads much better from here. Thanks, @johncrossman

PR #954, opened 2018-10-21
BOAC-1340, front-end refactor to leverage bootstrap performance improvements

johncrossman:

A semi-daunting refactor should make for a snappier front-end. Includes:
* use of Angular components for header, footer (required minor Angular version bump)
* user 'status' in bootstrap saves effort downstream
* replace `me` with `profile`
* rely less on broadcasting events; each component/view has a handle on `profile` so cohort updates are reflected immediately and across the board
* no page hard-refresh between login and home
 
Reviewing with `?w=1` might help.


paulkerschen:
I'm daunted, but since I'm not catching anything the linter didn't, let's get it on dev and see what happens!

PR #962, opened 2018-10-23
BOAC-1330: displays searched student at top of class list

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1330

This satisfies the first requirement; will amend or submit a follow-up for the second.

### DIFF
--- a/boac/static/app/course/courseController.js
+++ b/boac/static/app/course/courseController.js
@@ -78,9 +78,18 @@
       }
     });
 
+    var featureSearchedStudent = function(data) {
+      var section = _.clone(data);
+      var subject = _.remove(section.students, function(student) {
+        return student.uid === _.get($location.search(), 'u');
+      });
+      section.students = _.union(subject, section.students);
+      return section;
+    };
+
     var updateCourseData = function(response) {
       $rootScope.pageTitle = response.data.displayName;
-      $scope.section = response.data;
+      $scope.section = featureSearchedStudent(response.data);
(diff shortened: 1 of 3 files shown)


boac/static/app/course/courseController.js
```diff
@@ -78,9 +78,18 @@
       }
     });
 
+    var processResponse = function(data) {
```
johncrossman: All looks good, @lyttam  My only request: a different name for `processResponse` function. Something that is self-documenting?

lyttam: @johncrossman I was on the fence about the name. I can change it to something more specific like `reorderStudents` or `featureSearchedStudent`

johncrossman: +1 on `featureSearchedStudent`

boac/static/app/course/courseController.js
```diff
@@ -78,9 +78,18 @@
...
+        return student.uid === _.get($location.search(), 'u');
+      });
+      section.students = _.union(subject, section.students);
+      return section;
+    };
+
     var updateCourseData = function(response) {
       $rootScope.pageTitle = response.data.displayName;
-      $scope.section = response.data;
+      $scope.section = featureSearchedStudent(response.data);
```
lyttam: Amended, @johncrossman.

johncrossman:
Travis failure is due to an updated "flake8" dependency –– Python linting is suddenly stricter.  The offending py code can be fixed in a separate PR.

PR #964, opened 2018-10-25
BOAC-1330: adds control to change class list page size

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1330

### DIFF
--- a/boac/static/app/course/courseController.js
+++ b/boac/static/app/course/courseController.js
@@ -48,7 +48,12 @@
     $scope.lastActivityDays = utilService.lastActivityDays;
     $scope.pagination = {
       currentPage: 1,
-      itemsPerPage: 50
+      defaultItemsPerPage: 50,
+      itemsPerPage: 50,
+      selected: {
+        itemsPerPage: 50
+      },
+      options: [{label: 'Show 50', value: 50}, {label: 'Show 100', value: 100}]
     };
     $scope.tab = 'list';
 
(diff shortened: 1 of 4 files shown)


boac/static/app/course/courseController.js
```diff
@@ -169,6 +174,17 @@
       });
     };
 
+    $scope.resizePage = function(selectedItemsPerPage) {
+      var currentItemsPerPage = $scope.pagination.itemsPerPage;
+      $scope.pagination.currentPage = Math.round($scope.pagination.currentPage * (currentItemsPerPage / selectedItemsPerPage));
+      $scope.pagination.itemsPerPage = selectedItemsPerPage;
```
johncrossman: Nicely done here. The one thing I'd add is:
```
$location.search('page', $scope.pagination.currentPage);
```
That'll keep browser URL current.

boac/static/app/course/courseController.js
```diff
@@ -48,7 +48,12 @@
     $scope.lastActivityDays = utilService.lastActivityDays;
     $scope.pagination = {
       currentPage: 1,
-      itemsPerPage: 50
+      defaultItemsPerPage: 50,
+      itemsPerPage: 50,
+      selected: {
+        itemsPerPage: 50
+      },
+      options: [{label: 'Show 50', value: 50}, {label: 'Show 100', value: 100}]
```
johncrossman: @lyttam  Do we expect this value to persist if, for example, I reload the page in my browser?

johncrossman: I checked in with the UX dept and the answer is "No", the value does not need to persist. We'll keep the impl simple!

lyttam: Sounds good, thanks for checking.

PR #965, opened 2018-10-25
BOAC-1352 Query interface for student/course name search

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1352


johncrossman:
It's all about them queries.

PR #970, opened 2018-10-31
BOAC-1395 Don't blow up on non-alphanumeric search phrases

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1395


johncrossman:
BOAC is blowin' up?! Isn't that a good thing?

PR #971, opened 2018-10-31
BOAC-1293: Returns 500 status when cache refresh fails

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1293

### DIFF
--- a/boac/api/admin_controller.py
+++ b/boac/api/admin_controller.py
@@ -85,7 +85,8 @@ def start_load_only():
 @app.route('/api/admin/cachejob/refresh')
 @admin_required
 def start_refresh():
-    return tolerant_jsonify(cache_utils.refresh_request_handler(term()))
+    response = cache_utils.refresh_request_handler(term())
+    return tolerant_jsonify(response, status=500 if 'error' in response else 200)
 
 
 @app.route('/api/admin/cachejob/import_refresh')
--- a/tests/test_api/test_admin_controller.py
+++ b/tests/test_api/test_admin_controller.py
@@ -23,7 +23,6 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-
 from flask import current_app as app
 import pytest
 import simplejson as json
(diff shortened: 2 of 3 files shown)


tests/test_api/test_admin_controller.py
```diff
@@ -77,6 +76,13 @@ def test_api_key_disabled(self, app, client):
         response = client.get('/api/admin/cachejob', headers=headers)
         assert response.status_code == 401
 
+class TestCachejobRefresh:
+    """Cachejob Refresh API."""
+    def test_cachejob_refresh_success(self, client, admin_session):
+        """Returns a well-formed sucess response."""
+        response = client.get('/api/admin/cachejob/refresh')
+        assert response.status_code == 200
+        assert response.headers.get('Content-Type') == 'application/json'
```
lyttam: I wanted to write a test for the error response, but I don't think it's possible with our current mocking capabilities.

boac/api/admin_controller.py
```diff
@@ -85,7 +85,11 @@ def start_load_only():
 @app.route('/api/admin/cachejob/refresh')
 @admin_required
 def start_refresh():
-    return tolerant_jsonify(cache_utils.refresh_request_handler(term()))
+    response = cache_utils.refresh_request_handler(term())
+    if response.get('error'):
+        return tolerant_jsonify(response, status=500)
+
+    return tolerant_jsonify(response)
```
johncrossman: Looks good @lyttam. Sort out Travis woes and I'll merge. Fyi, the above could be written with a one-liner:
```
return tolerant_jsonify(response, status=500 if 'error' in response else 200)
```

paulkerschen:
@lyttam Travis's lint-py complaints should be trivial to fix. Unfortunately, I suspect the one pytest failure is caused by the new cache job refresh test. Since we aren't able to mock out that endpoint, hitting it really does trigger a cache job refresh in a background thread, which in turn creates a bunch of alerts that conflict with later code trying to set up alerts of its own.

Nessie deals with a similar problem by suppressing background job kickoff in test mode.
https://github.com/ets-berkeley-edu/nessie/blob/master/nessie/jobs/background_job.py#L88
Alternatively, we could just wait for a functional mocking library before testing that endpoint.

boac/api/admin_controller.py
lyttam: Nice! Will amend.

lyttam:
@johncrossman Amended.

johncrossman:
Thanks @lyttam

PR #972, opened 2018-11-01
BOAC-1406, onCreateCuratedCohort callback adds students; event-listeners begone

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1406


paulkerschen:
How devilishly asynchronous. Thanks, @johncrossman, and let's all list the things we won't miss about Angular 1.x.

PR #974, opened 2018-11-01
BOAC-1396 Include instructor names in class search results

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1396


raydavis:
The best of all possible PRs.

PR #975, opened 2018-11-01
BOAC-1344, perk up the washed out splash page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1344

The perkier UX is in screenshot attached to BOAC-1344.


paulkerschen:
See, kids? No need to be scared - the Regents were there all along.

PR #977, opened 2018-11-01
BOAC-1398, Save Cohort button styled as primary

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1398


raydavis:
Can't think of a better choice off-hand...

PR #978, opened 2018-11-02
BOAC-1412, if callback is nil when cohort-create then _.noop

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1412

I discovered a bit of a bug in our error-handling, too.


lyttam:
Looks good to me.

PR #979, opened 2018-11-02
BOAC-1415, first step in prototyping Vue.js version of BOAC Admin page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1415

The dir structure is similar to Nessie. Operational stuff (eg, tox) are meant to operate side-by-side without sharing configs (eslint.yml).

An interesting diff is in `boac/routes.py`.  The idea is that `https://boac.berkeley.edu/admin` goes to the Admin page _as we know it_ and `https://harvey.berkeley.edu/admin` (future URL scheme depends on product name) will go to the new Vue version of that page.


johncrossman:
I have prefixed the PR summary with "[DO NOT MERGE]"

johncrossman:
One amendment coming soon to this PR. I'll update summary when it's ready for review.  cc: @paulkerschen @lyttam

johncrossman:
@paulkerschen @lyttam  This PR is ready for review.

boac/routes.py
```diff
@@ -66,7 +66,10 @@ def handle_unmatched_api_route(**kwargs):
     @app.route('/', defaults={'path': ''})
     @app.route('/<path:path>')
     def front_end_route(**kwargs):
-        return make_response(open(app.config['INDEX_HTML']).read())
+        host_url = request.host_url
+        is_legacy = 'boac' in host_url or 'boa' not in host_url
```
paulkerschen: @johncrossman So the distinction is `boa.berkeley.edu` versus `boac.berkeley.edu`? I'm with you so far, but what's the effect while the alias hostnames are not yet in place (or if we're running the server locally)?

johncrossman: Locally, I have `boa-local.berkeley.edu` in _/etc/hosts_ which might be feasible for all devs. As for "alias hostnames not yet in place", we'll need to see where we are in a week. If zero progress is made on the new name then we might fall back to using a request arg: `?v=1`. I'd prefer not to rely on path pattern matching (eg, `/v/admin`)

scripts/production-build-boac-vuejs.sh
```diff
@@ -0,0 +1,13 @@
...
+set -e
+
+npm run build-vue
+
+cp dist/static/vue/css/* dist/static/css/
+cp dist/static/vue/js/* dist/static/js/
+cp dist/static/vue/img/* dist/static/img/
+cp dist/static/vue/index.html dist/static/
+
+exit 0
```
johncrossman: The script above is run by `buildspec` (CodePipeline). The old and new (vue.js) static files will live side-by-side in `dist` directory; the workaround above – ie, copy files after `npm build-vue` – avoids inadvertent removal of `dist`.

paulkerschen:
Thanks for working through all the tricky corners in this problem, @johncrossman. I dub you, "VueMaster"

PR #983, opened 2018-11-06
BOAC-1416, scope.section check before googleAnalyticsService.track

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1416

A Jira with "busted" in the summary is worth fixing.


raydavis:
Thanks, @johncrossman

PR #987, opened 2018-11-09
BOAC-1436, structure App.vue to match legacy index; menu options in top-right header

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1436

### DIFF
--- a/src/components/Sidebar.vue
+++ b/src/components/Sidebar.vue
@@ -1,12 +1,12 @@
 <template>
-  <div class="main">
-    <slot/>
+  <div>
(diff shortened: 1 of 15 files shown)


src/components/Sidebar.vue
```diff
@@ -1,12 +1,12 @@
 <template>
-  <div class="main">
-    <slot/>
+  <div>
+    I am a sidebar.
```
paulkerschen: Yes, you are. Yes, you are!!!!

PR #989, opened 2018-11-13
BOAC-1308 Simplify cohort sidebar navigation

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1308

Designs: https://jira.ets.berkeley.edu/jira/browse/BOAC-1238

User-facing text is changed to match the new "cohort"/"curated group" distinction, but I didn't go down the road of changing paths, filenames, variables, etc., since we expect Vue to have its own opinion on such things before long.


johncrossman:
Parts is parts and groups is groups. Thx @paulkerschen

PR #990, opened 2018-11-14
BOAC-1436, use Vuetify for layout; wire up demoMode.blur on Admin page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1436

* [Vuetify](https://vuetifyjs.com) allows us to avoid our own flex CSS.
* I moved custom configs (eg, Jest) out of `package.json`
* the Admin page makes progress, too


paulkerschen:
Vuetiful!

PR #992, opened 2018-11-14
BOAC-1430: blurs instructor names in demo mode

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1430

I added blurring on the search results page as well as the course page.


paulkerschen:
Thanks @lyttam!

PR #998, opened 2018-11-15
BOAC-1411: adds page size param to course page URL

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1411


raydavis:
But what if I really _need_ a page size of 45000?

PR #1006, opened 2018-11-20
BOAC-1419: exposes read-only admin page in prod

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1419


paulkerschen:
Thanks @lyttam, looks good.

PR #1014, opened 2018-11-26
BOAC-1447 Remove Angular-side 'Manage Curated Groups' page

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1447


johncrossman:
Less is more.

PR #1019, opened 2018-11-26
BOAC-1413 Search for courses in appropriate contexts only

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1413


johncrossman:
Everything in its right place.

PR #1021, opened 2018-11-27
BOAC-1455, sensible strategy for mapping new Vue paths to legacy Angular paths

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1455


raydavis:
Thanks for the improvements, @johncrossman !

PR #1022, opened 2018-11-27
BOAC-1445 Add COE probation filter

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1445


johncrossman:
That's what we call end-to-end. Thx @paulkerschen

PR #1023, opened 2018-11-27
BOAC-1343 More UX tweaks for students with minimal data

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1343

### DIFF
--- a/boac/static/app/home/sortableAlertsTable.html
+++ b/boac/static/app/home/sortableAlertsTable.html
@@ -112,7 +112,8 @@
       </td>
       <td class="group-summary-column group-summary-column-major">
         <span class="sr-only" data-ng-bind="srText['majors[0]']"></span>
-        <div data-ng-bind="major" data-ng-repeat="major in student.majors"></div>
+        <div data-ng-if="student.majors.length == 0">--<span class="sr-only">No data</span></div>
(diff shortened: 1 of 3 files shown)


paulkerschen:
We've got nothing on them.

boac/static/app/home/sortableAlertsTable.html
```diff
@@ -112,7 +112,8 @@
       </td>
       <td class="group-summary-column group-summary-column-major">
         <span class="sr-only" data-ng-bind="srText['majors[0]']"></span>
-        <div data-ng-bind="major" data-ng-repeat="major in student.majors"></div>
+        <div data-ng-if="student.majors.length == 0">--<span class="sr-only">No data</span></div>
```
johncrossman: @raydavis I suggest ` === ` above.

raydavis: Oogh, sorry about that, @johncrossman . I'm surprised that tox didn't catch that.

johncrossman: Under the hood, tox uses eslint to lint js files and HTML is ignored. We do not use `eslint-plugin-html` because it does very little – it would not have reported this `===` concern.

PR #1024, opened 2018-11-27
BOAC-1453 Fix add-to-curated-group checkbox in search results

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1453


johncrossman:
'change' is good.

PR #1027, opened 2018-11-28
BOAC-1474, remove vueMigratedLink component; Vue redirect logic is now in Angular routes.js

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1474


paulkerschen:
Seems easier to manage for sure.

PR #1030, opened 2018-11-29
BOAC-1483, capture path arg (eg, /student/:uid) when redirect to Vue

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1483

Enter regex.


lyttam:
Thanks for solving this, @johncrossman !

PR #1031, opened 2018-11-29
BOAC-1483, append query_string to redirect; if incoming path is Vue then do the right thing

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1483

Follow up to PR #1030


boac/routes.py
```diff
@@ -92,22 +92,32 @@ def after_api_request(response):
...
-            return make_response(open(app.config['INDEX_HTML_VUE']).read())
+    return vue_redirect
+
+
+def _vue_response(app, uri_path):
+    vue_base_url = app.config['VUE_LOCALHOST_BASE_URL']
+    if vue_base_url:
+        return redirect(vue_base_url + uri_path)
+    else:
+        return make_response(open(app.config['INDEX_HTML_VUE']).read())
```
johncrossman: Append `?w=1` to github URL to see simpler diff. Bringing the above logic out of `def front_end_route(...)` reduced complexity – thus the deceptive diff.

paulkerschen:
The app with two brains.

PR #1036, opened 2018-11-30
BOAC-1486, use BootstrapVue (for now) and remove Vuetify and vue-material

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1486

With Vuetify out we need to fix the layout – not a big deal. See the Jira for "why BootstrapVue?".


paulkerschen:
Thanks for the extra round of research, @johncrossman. I'm persuaded on the choice.

PR #1041, opened 2018-12-03
BOAC-1468: Rewrites curated group page header in Vue

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1477

I added the [Flask-Cors](https://flask-cors.readthedocs.io/en/latest/) library to handle cross-origin requests from Vue components (localhost:8080) to APIs (localhost:5000).  GET, POST, and OPTIONS can be handled without it, but DELETE and PUT require the `Content-Type` header to be explicitly set.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -51,6 +51,12 @@ const mutations = {
     if (indexOf > -1) {
       state.errors.splice(indexOf, 1);
     }
+  },
+  updateCuratedGroup: (state: any, data: any) => {
+    var group = state.user.myCuratedCohorts.find(
+      group => group.id === +data.id
+    );
+    group.name = data.name;
   }
 };
 
--- a/src/views/group/CuratedGroup.vue
+++ b/src/views/group/CuratedGroup.vue
@@ -1,125 +1,179 @@
 <template>
-  <div class="container-manage-cohorts">
+  <div class="cohort-container">
     <Spinner/>
-    <div v-if="!loading">
-      <h1 class="page-section-header">Manage Curated Groups</h1>
-      <div v-if="!curatedGroups.length" data-ng-controller="CreateCuratedCohortController">
-        You have no curated groups.
-        <a id="curated-cohort-create"
-           href
-           data-ng-click="openCreateCuratedCohortModal()"><i class="fas fa-plus"></i> Create a new curated group</a>
+    <div class="cohort-header-container" v-if="!loading">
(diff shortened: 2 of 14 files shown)


src/store.ts
```diff
@@ -51,6 +51,12 @@ const mutations = {
     if (indexOf > -1) {
       state.errors.splice(indexOf, 1);
     }
+  },
+  updateCuratedGroup: (state: any, data: any) => {
+    var group = state.user.myCuratedCohorts.find(
+      group => group.id === +data.id
```
johncrossman: @lyttam Is that a stray '+'?

src/views/group/CuratedGroup.vue
```diff
@@ -1,125 +1,190 @@
...
   computed: {
-    curatedGroups() {
+    pluralizeStudents() {
+      if (this.curatedGroup.students.length === 0) {
+        return '0 students';
+      } else if (this.curatedGroup.students.length === 1) {
+        return 'One student';
+      } else {
+        return this.curatedGroup.students.length + ' students';
+      }
```
johncrossman: @lyttam Does it make sense to use the `pluralize` filter defined in `main.ts`? You might enhance it such that `length === 1` gives the nice "One" wording.

src/views/group/CuratedGroup.vue
```diff
@@ -1,125 +1,190 @@
...
+                  class="btn-link cohort-manage-btn-link"
+                  v-on:click="enterRenameMode">
+            Rename
+          </button>
+          <span>
+            <span class="faint-text">|</span>
+            <b-btn v-b-modal="'myModal'"
+                    id="delete-cohort-button"
+                    aria-label="Delete this cohort"
+                    class="btn-link cohort-manage-btn-link">
```
johncrossman: Minor: indentation fix needed above

johncrossman:
Looks good, @lyttam Thank you for leading the way on `v-b-modal`.  A few minor questions posted.

src/views/group/CuratedGroup.vue
lyttam: Absolutely, thanks for pointing that out. Will amend.

lyttam:
@johncrossman, amended with your suggestions - thanks!

johncrossman:
Thank you @lyttam  Nicely done.

PR #1042, opened 2018-12-04
BOAC-1489 Student profile: bio, photo, academic status

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1489

### DIFF
--- a/boac/static/app/main.css
+++ b/boac/static/app/main.css
@@ -27,6 +27,33 @@
   margin: 0;
 }
 
+.avatar {
+  background-image: url(/static/app/shared/avatar-50.png);
+  background-size: cover;
+  border: 5px solid #ccc;
+  border-radius: 30px;
+  height: 60px;
+  object-fit: cover;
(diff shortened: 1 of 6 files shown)


paulkerschen:
In moving the CSS around, my goal was to eliminate reliance on the `student.css` currently being used by the Angular front end.

Styles specific to the student profile page are copied into `student.vue`, modifying and simplifying where appropriate.

Styles that we expect to be more globally applicable are moved into `main.css`, under the assumption that file will persist even when the rest of the Angular front end is gone.

boac/static/app/main.css
```diff
@@ -342,6 +373,20 @@ body {
   height: 100%;
 }
```
paulkerschen: Global styles defined below override some Bootstrap defaults to make fonts and colors more consistent between Angular and Vue.

johncrossman:
Looks good @paulkerschen  Thank you for the global styles.

PR #1044, opened 2018-12-04
BOAC-1439, accessibility of demo-mode toggle checkbox

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1439


paulkerschen:
@johncrossman I'm having a bit of trouble figuring out the function of `demoMode` versus `inDemoMode`. Is one a global config and one a model specific to this component? Is there any way to make the difference clearer?

johncrossman:
@paulkerschen Thank you – I agree that it was unclear. I amended for clarity and put comment in UserMetadata: "If you must bind a user attribute to a form input field then declare it in data block".  Plus, new mixins allow for simpler component code.

PR #1045, opened 2018-12-04
BOAC-1497 Inactive status in student profile

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1497


johncrossman:
Spin a `mixin` in the mix

PR #1048, opened 2018-12-05
BOAC-1494 BOAC-1470, add students to curated group from search results

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1470

Noteworthy:
* `beforeRouteUpdate` allows in-place route refresh (eg, search from search view)
* eslint rule 'arrow-parens' relaxed cuz parens in arrow function impact scope of `this.` modifier(?!)
* work remains on modal
* I moved simple `data` properties out of `mixin`


src/views/Search.vue
```diff
@@ -82,7 +96,7 @@ export default {
           0,
           this.limit
         )
-          .then(data => {
+          .then((data) => {
```
johncrossman: More info on parens in arrow functions: https://dzone.com/articles/learning-es6-arrow-functions-and-the-visibility-of   I'm going to follow article's advice with `var self = this;` prior to Promise block. Amending now.

src/views/Search.vue
```diff
@@ -74,6 +88,8 @@ export default {
     performSearch() {
       this.phrase = _.trim(this.phrase);
       if (this.phrase) {
+        // In arrow function below, 'this' does NOT reference Search component. Using 'self' as an alias is sufficient.
+        const self = this;
```
paulkerschen: Fickleness, thy name is JavaScript `this`.

PR #1049, opened 2018-12-05
BOAC-1494, wired up api call to addStudents() to group

johncrossman:

The modal will be next.

### DIFF
--- a/src/components/curated/CuratedGroupSelector.vue
+++ b/src/components/curated/CuratedGroupSelector.vue
@@ -5,28 +5,27 @@
              class="sr-only">Select all students to add to a curated group</label>
       <b-form-checkbox plain
                        class="p-2 mr-0"
-                       v-model="checked"
+                       :disabled="isSaving"
+                       v-model="isSelectAllChecked"
                        :indeterminate="indeterminate"
                        aria-describedby="students"
                        aria-controls="students"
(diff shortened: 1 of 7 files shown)


lyttam:
@johncrossman looks great, just one suggestion:  name it CreateCuratedGroupModal.vue, or perhaps CuratedGroupModalCreate.vue (see https://vuejs.org/v2/style-guide/#Order-of-words-in-component-names-strongly-recommended)

johncrossman:
Thanks for the tip, @lyttam  Amending.

src/components/curated/CuratedGroupSelector.vue
```diff
@@ -50,33 +50,43 @@
...
+                 v-b-modal="'modal'"
+                 aria-label="Create a new curated group">
             <i class="fas fa-plus"></i> Create New Curated Group
-          </span>
+          </b-btn>
         </b-dropdown-item>
       </b-dropdown>
+      <b-modal id="modal"
+               hide-footer
+               @ok="createNewCuratedGroup"
```
lyttam: @johncrossman If you add `.stop` here, I think you won't need to call `preventDefault()`

`@ok="createNewCuratedGroup.stop"`

https://vuejs.org/v2/guide/events.html#Event-Modifiers

PR #1051, opened 2018-12-06
BOAC-1494, cohort/group name validation for create new/rename

johncrossman:

`src/mixins/Validator.vue` is for reuse (eg, when renaming cohort)


paulkerschen:
Behold the Validator!

PR #1056, opened 2018-12-07
BOAC-1519, user auth in Vue-land: avoid browser reload and report errors on /login

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1519

* auth-related endpoints get `/api/auth/` prefix
*  `store` tracks `isUserAuthenticated` and user object separately for speed in rendering
* error messaging in `/login`
* basic page title mgmt
* router.ts tidying

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -21,51 +21,52 @@ const safeNext = (to: any, next: any) => {
   }
 };
 
-const beforeEach = (to: any, from: any, next: any) => {
-  let legacyPathRedirect = _.get(to, 'meta.legacyPathRedirect');
-  if (legacyPathRedirect) {
-    for (const key in to.params) {
-      legacyPathRedirect = _.replace(
-        legacyPathRedirect,
-        ':' + key,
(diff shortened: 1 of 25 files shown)


src/router.ts
```diff
@@ -82,38 +83,57 @@ const router = new Router({
...
+      props: true,
+      meta: {
+        title: 'Curated Group'
+      }
     },
     {
       path: '/student_:uid',
       beforeEnter: requiresAuth,
-      component: Student,
-      meta: { legacyPathRedirect: '/student/:uid' }
```
johncrossman: @paulkerschen do we want the `legacyPathRedirect` above?

paulkerschen: @johncrossman I think we're okay without it. The student page isn't done, but I think enough of it is done to be worth showing whenever Vue gets enabled on dev.

paulkerschen:
Looks good to me! Thanks for extra tidying on the auth paths.

PR #1058, opened 2018-12-07
BOAC-1535, improved /login setup; lazy-init per reqts of route; better use of /status

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1535

* A more readable and performant `router.ts`
* Simpler mgmt of login vs. standard wireframe
* Load user profile after template has loaded
* threw in NotFound view w/o UX

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -1,30 +1,16 @@
 <template>
   <div id="app" class="fill-viewport">
-    <component :is="layout"
-               v-if="layout"
-               class="fill-viewport"></component>
+    <router-view></router-view>
   </div>
 </template>
 
 <script>
-import Login from './layouts/Login.vue';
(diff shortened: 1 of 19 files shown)


src/App.vue
```diff
@@ -1,30 +1,16 @@
...
 export default {
-  name: 'App',
-  computed: {
-    layout: () =>
-      store.getters.isUserAuthenticated === null
-        ? null
-        : store.getters.isUserAuthenticated
-          ? StandardLayout
-          : Login
-  }
```
johncrossman: The when-to-use-StandardLayout logic is now managed in semi-nested routes of `router.ts`

paulkerschen:
Looks good to me, but it won't be BOAC until the 404 page is a ticket from Berkeley to nowhere.

PR #1060, opened 2018-12-10
BOAC-1538, 'legacyRedirectsEnabled' config (Vue) is false locally and true in all other envs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1538

Jira above has link to doc with configs explained.


paulkerschen:
That's how you swap out an engine in midflight.

PR #1065, opened 2018-12-11
BOAC-1531, fix styling and aria alerts on curated group selector

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1532


lyttam:
Looks good.

PR #1066, opened 2018-12-11
BOAC-1525, separate 'curated' and 'cohort' in front-end /api structure

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1525

The `/cohort` view is coming soon and we'll need a dedicate client-side file for cohort api calls.


raydavis:
As a shepherd separates the sheep from the goats.

PR #1068, opened 2018-12-12
BOAC-1539, 'excludeCohorts' option in /api/profile/my if front-end wants data modularity

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1539

The Vue front-end will load user profile and cohorts/groups separately. The key benefit is a more manageable front-end store. If we plan to add/remove/rename cohorts then I'd like to detach them from the user object. Also, a faster `/profile/my` causes certain page elements (eg, header-menu) to load more quickly.  This has no impact on Angular side of things.


paulkerschen:
Thanks for the reassuring test coverage, @johncrossman.

PR #1070, opened 2018-12-12
BOAC-1539, vuex store restructured for scalability, better performance

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1539

My hope is that this new structure is intuitive and scalable. There will be refinements (e.g., single, centralized call to `store.getters['context/apiBaseUrl']` and further use of Vuex `mapGetters`). I'm pleased with `router.ts` simplification, the decoupling of user, myCohorts, myCurated objects (see PR #1068), and faster page load with `async` actions in the store.


src/api/auth.ts
```diff
@@ -2,31 +2,32 @@ import axios from 'axios';
 import store from '@/store';
 
 export function devAuthLogIn(uid: string, password: string) {
+  let apiBaseUrl = store.getters['context/apiBaseUrl'];
```
johncrossman: The var declaration above _cannot_ be global within this file. We can revisit the minor concern of multiple `apiBaseUrl` assignments in the `/api` module. For now, I want to restrain the code disruption of this PR.

src/mixins/AppConfig.vue
```diff
@@ -1,16 +1,15 @@
...
-    baseUrl: () => store.state.apiBaseUrl,
-    devAuthEnabled: () => _.get(store.getters.config, 'devAuthEnabled'),
-    supportEmailAddress: () =>
-      _.get(store.getters.config, 'supportEmailAddress')
+    ...mapGetters('context', [
+      'apiBaseUrl',
+      'currentEnrollmentTermId',
+      'devAuthEnabled',
+      'supportEmailAddress'
+    ])
```
johncrossman: This mixin can be refactored away. Components relying on AppConfig can just use `mapGetters`, as above.

src/store/modules/context.ts
```diff
@@ -0,0 +1,88 @@
...
+const actions = {
+  dismissError: ({ commit }, id) => {
+    commit('dismissError', id);
+  },
+  loadingComplete: ({ commit }) => {
+    commit('loadingComplete');
+  },
+  loadingStart: ({ commit }) => {
+    commit('loadingStart');
+  },
```
johncrossman: Not every "mutation" must have an accompanying "action". That said, `state` cannot be modified in actions. As we better understand Vuex we can simplify some of this.

paulkerschen:
Seems like a sensible setup to me, to the extent I can follow - I'll let @lyttam take a pass too.

lyttam:
Thanks, this seems logical and clean. Nice work @johncrossman!

PR #1071, opened 2018-12-12
BOAC-1540 Collapsible term GPA table

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1540


sandeepmjay:
Looks good @paulkerschen.

PR #1072, opened 2018-12-13
BOAC-1533, search results > classes; plus, a few mixin fixes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1533


paulkerschen:
Thanks for the extra mixin fixin.

PR #1074, opened 2018-12-13
BOAC-1529, move our custom Vue.filters to ./filters file; a cleaner main.ts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1529

A little tidying before I finish `variablePrecisionNumber` filter.


paulkerschen:
Thanks, @johncrossman - the filters are definitely grown-up enough to move into their own place, but should we squash these two commits to keep the history clean?

johncrossman:
Oh, whoops! That was not intentional. Closing.

PR #1077, opened 2018-12-13
BOAC-1533, courses section of search results (sort, style, /course route)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1533


paulkerschen:
Sorting!

PR #1080, opened 2018-12-14
BOAC-1493 Consolidate duplicate lastActivityDays methods

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1493

The trouble with mixins! We also have a `lastActivityDays` method over in https://github.com/ets-berkeley-edu/boac/blob/master/src/mixins/StudentAnalytics.vue, which is distinguished from StudentMetadata.vue by focusing specifically on our bCourses analytics (and is about to get some more methods to help with boxplots and the matrix).

### DIFF
--- a/src/mixins/StudentMetadata.vue
+++ b/src/mixins/StudentMetadata.vue
@@ -1,5 +1,4 @@
 <script>
-import _ from 'lodash';
 import store from '@/store';
 
 export default {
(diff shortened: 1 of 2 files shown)


src/mixins/StudentMetadata.vue
```diff
@@ -18,28 +17,6 @@ export default {
...
-        _.get(analytics, 'lastActivity.student.raw'),
-        10
-      );
-      if (!timestamp || isNaN(timestamp)) {
-        return 'Never';
-      }
-      // Days tick over at midnight according to the user's browser.
-      let daysSince = Math.round(
-        new Date().setHours(0, 0, 0, 0) -
-          new Date(timestamp * 1000).setHours(0, 0, 0, 0) / 86400000
```
paulkerschen: At some point we may want to examine the linter, which I think is leading to counterintuitive layout on complex expressions. It doesn't like "unnecessary" parentheses, but here, stripping parentheses from the Angular version of the method yields the result "XYZ visited the course site 1544687982126 days ago."

johncrossman:
Let the growing pains begin. Thx @paulkerschen

PR #1083, opened 2018-12-14
BOAC-1546 Don't show course properties that ain't there

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1546


raydavis:
They were not there again today. I wish those course properties would go away.

PR #1084, opened 2018-12-17
BOAC-1521: Adds scroll-to-anchor behavior

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1521


paulkerschen:
Thanks, @lyttam, looks good. I gather `$nextTick` is important for doing fancy things.

lyttam:
@paulkerschen Yes, `$nextTick` is what you do when you need to interact with a loaded DOM.

PR #1090, opened 2018-12-19
BOAC-1562 Leverage Lodash utils in Vue templates

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1562

Since Vue templates don't have the same implicit short-circuit evaluation as Angular 1, we need a better guard against errors of the `Cannot read property 'length' of null` type. After some hemming and hawing, I decided that the least obtrusive way of handling it was to explicitly add a couple of Lodash utils to our arsenal.

### DIFF
--- a/src/mixins/Util.vue
+++ b/src/mixins/Util.vue
@@ -1,0 +1,11 @@
+<script>
+import _ from 'lodash';
+
+export default {
+  name: 'Util',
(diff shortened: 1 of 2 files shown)


raydavis:
Oh man, I'm going to miss those short-circuits...

PR #1091, opened 2018-12-19
BOAC-1559: Improves rename curated group screen reader experience

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1559


johncrossman:
Screenreaders thank you.

PR #1092, opened 2018-12-19
BOAC-1562 Fix vue-analytics console error

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1562

Switching around the initialization logic quiets `vue-analytics.js:1 t.replace is not a function` errors in local testing.


raydavis:
But if it makes you happy to feed squirrels to the nuts, who am I to say "Nuts to the squirrels"?

PR #1094, opened 2018-12-20
BOAC-1569, router fixes (py, js and vue) to support split-brain testing

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1569

The most delicate fix in this PR is found in `router.ts` where we must parse VUE_PATHS config and perform "legacyPathRedirect" if the URL path is not yet supported in Vue-land. This split-brain code will be removed in late Jan.

### DIFF
--- a/boac/static/app/routes.js
+++ b/boac/static/app/routes.js
@@ -118,7 +118,7 @@
         reloadOnSearch: false
       })
       .state('filteredCohortsAll', {
-        url: '/cohort/filtered/all',
+        url: '/cohorts/all',
         views: standardLayout('AllFilteredCohortsController', '/static/app/cohort/filtered/all.html'),
         resolve: resolvePrivate
       })
@@ -166,7 +166,7 @@
         reloadOnSearch: false
       });
 
-  }).run(function($rootScope, $state, $transitions, authFactory, config, status) {
+  }).run(function($location, $rootScope, $state, $transitions, authFactory, config, status) {
 
     $state.defaultErrorHandler(function(error) {
       var message = _.get(error, 'detail.message');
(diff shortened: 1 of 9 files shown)


boac/static/app/routes.js
```diff
@@ -225,6 +213,15 @@
     });
 
     $transitions.onSuccess({}, function() {
+      if (config.vueEnabled) {
+        var path = $location.url();
+        for (var pattern in config.vuePaths) {
+          if (new RegExp(pattern).exec(path)) {
+            // Force page refresh. The server-side routing will redirect user to Vue-enabled index.html and proper path.
+            window.location = $location.absUrl();
```
johncrossman: The logic above, when run during the `$transitions.onBefore()` hook, was unreliable. Using `$location` in the `$transitions.onSuccess()` hook seems to work.

src/router.ts
```diff
@@ -107,13 +117,17 @@ const router = new Router({
...
           }
         },
         {
           path: '/student/:uid',
           beforeEnter: requiresAuth,
-          component: Student
+          component: Student,
+          meta: {
+            legacyPathRedirect: '/student/:uid'
+          }
```
johncrossman: Until the cutover is done, _each_ Vue route must define `meta.legacyPathRedirect`.

boac/static/app/routes.js
```diff
@@ -118,7 +118,7 @@
         reloadOnSearch: false
       })
       .state('filteredCohortsAll', {
-        url: '/cohort/filtered/all',
+        url: '/cohorts/all',
```
johncrossman: Changing this route path made regex parsing easier in `routes.py`

raydavis:
This does indeed look like a delicate surgery.

PR #1097, opened 2018-12-20
BOAC-1586, correct the naming convention in API routes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1586

I will be adding a few more `/api/cohort/` endpoints and I'd like a proper API namespace.


paulkerschen:
There was never any kind of cohort other than a filtered cohort! Anyone who says otherwise is a liar!

PR #1099, opened 2018-12-21
BOAC-1578: fixes sorting on curated group

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1578


paulkerschen:
Thinking bout this and that.

PR #1100, opened 2018-12-21
BOAC-1586, show filters section of /cohort view with edit/remove links

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1586

Many TODOs ahead. The operations of this page are complex and I think dedicated `store` module is reasonable. The `CohortEditSession` mixin allows all sub-components to easily share getters and store actions.


paulkerschen:
Well on the way.

PR #1103, opened 2018-12-21
BOAC-1587, document all VUE_PATHS in config/default.py; add /teams route

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1587

cc: @pfarestveit


paulkerschen:
That's a helpful list, that is.

pfarestveit:
Thanks @johncrossman !

PR #1106, opened 2018-12-21
BOAC-1568 Vue: add to curated groups from course page

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1568

Component magic!

### DIFF
--- a/src/components/curated/CuratedGroupSelector.vue
+++ b/src/components/curated/CuratedGroupSelector.vue
@@ -16,6 +16,7 @@
     </div>
     <div>
       <b-dropdown id="curated-group-dropdown-select"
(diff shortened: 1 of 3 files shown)


src/components/curated/CuratedGroupSelector.vue
```diff
@@ -113,7 +114,9 @@ export default {
     });
   },
   computed: {
-    showMenu: () => this.sids.length
+    showMenu() {
+      return this.sids.length;
+    }
```
paulkerschen: Note gotcha with `this` and arrow functions: https://stackoverflow.com/questions/43929650/vuejs-why-is-this-undefined

PR #1108, opened 2019-01-03
BOAC-1588, server-side API to populate and manage menu of filters for /cohort view

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1588

This code will evolve as new features are folded in; I just want to establish some sensible patterns. Separating the "menu API"  from `/api/cohort` helped with semantic separation. Relating everything to cohorts becomes confusing.


tests/test_api/test_menu_controller.py
```diff
@@ -0,0 +1,300 @@
...
+            'type': 'array',
+            'value': student_class_level,
+        }
+
+    def test_search_menu_not_authenticated(self, client):
+        """Custom cohort deletion requires authentication."""
+        assert self._post(client).status_code == 401
+
+    def test_search_menu_none_disabled(self, client, coe_advisor_session):
+        """Custom cohort deletion requires authentication."""
```
raydavis: Copy-and-paste glitch?

src/components/cohort/AddCohortFilterMenu.vue
```diff
@@ -0,0 +1,177 @@
...
+  <div>
+    <div class="sr-only" aria-live="polite">{{ filterUpdateStatus }}</div>
+    <div>
+      <div>
+        <div class="cohort-filter-draft-row">
+          <div class="cohort-filter-draft-column-01">
+            <b-form-select id="draft-filter"
+                           class="btn-filter"
+                           :disabled="disable"
+                           :aria-disabled="disable"
```
raydavis: Self-education: Is `aria-disabled` needed because this doesn't produce a standard "select" element?

src/components/cohort/CohortPageHeader.vue
```diff
@@ -114,12 +117,25 @@ export default {
...
   }),
+  created() {
+    this.name = this.cohortName;
+  },
+  computed: {
+    renameMode() {
+      return this.pageMode === 'rename';
+    },
+    disableButtons() {
+      return !this.isOwnedByCurrentUser && false;
```
paulkerschen: @johncrossman Won't this expression always return `false`, no matter what the value of `this.isOwnedByCurrentUser`?

tests/test_api/test_menu_controller.py
johncrossman: Copy-and-paste secrets revealed! Fixing.

src/components/cohort/AddCohortFilterMenu.vue
johncrossman: Good catch. `aria-disabled` was necessary when we rendered this menu with the `bootstrap-dropdown` component because, as explained in https://www.w3.org/TR/html-aria, we use `aria-disabled` if the HTML element does not support standard 'disabled' attribute. With a `select` we can drop the aria-disabled. Amending...

src/components/cohort/CohortPageHeader.vue
johncrossman: Good find, @paulkerschen My debugging expression found its way into the commit.  Fixing.

johncrossman:
@raydavis @paulkerschen Amendments in, Travis green.

PR #1110, opened 2019-01-03
BOAC-1541, rename/delete cohort and keep 'store' in sync

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1541

Rename or delete cohort from `/cohort` from view.


johncrossman:
Amended this PR to add 'delete' support.  cc: @lyttam @raydavis

PR #1113, opened 2019-01-04
BOAC-1583 Fix search results sorting in Vue

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1583


johncrossman:
We are case-blind. Thanks @raydavis

PR #1114, opened 2019-01-04
BOAC-1582, use numFormat filter to format cumulativeUnits in search column

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1582

Perhaps custom `variablePrecisionNumber` filter is not needed in Vue-land.


raydavis:
More zeros for less work.

PR #1115, opened 2019-01-04
BOAC-1577 Vue: add carets for student search result sorts

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1577

Also removes the no-op `dropup`/`caret` classes which were copied from the AngularJS implementation.

While testing this, I noticed a couple of more minor visual glitches relating to caret spacing:
* The first sort on "GPA" bounces the column header text to the left (because it's currently right-aligned).
* The first sort on a wrapped column header (e.g., "Term Units" in my testing) bounces all rows down (because the caret appears under the header text).

This is bug-for-bug compatible with the Angular page, though, so I'll leave it as is for the nonce.

### DIFF
--- a/src/components/course/SortableCourseList.vue
+++ b/src/components/course/SortableCourseList.vue
@@ -17,14 +17,14 @@
     </div>
     <table class="table-full-width" v-if="totalCourseCount">
       <tr role="row">
-        <th :class="{ dropup: !sort.reverse.section }">
+        <th>
           <button id="column-sort-button-section"
                   class="btn btn-link table-header-text group-summary-column-header group-summary-header-sortable search-results-cell"
                   :aria-label="`Sort by section ${ sort.by === 'section' ? describeReverse(sort.reverse.section) : ''}`"
                   tabindex="0"
                   @click="courseSort('section')">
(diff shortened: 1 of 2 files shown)


src/components/course/SortableCourseList.vue
```diff
@@ -17,29 +17,29 @@
...
+        <th>
           <button id="column-sort-button-section"
                   class="btn btn-link table-header-text group-summary-column-header group-summary-header-sortable search-results-cell"
                   :aria-label="`Sort by section ${ sort.by === 'section' ? describeReverse(sort.reverse.section) : ''}`"
                   tabindex="0"
                   @click="courseSort('section')">
             Section
-            <span class="caret" v-if="sort.by === 'section'">
+            <span v-if="sort.by === 'section'">
               <i :class="{
```
raydavis: Am I right to think that this up-or-down-caret could become a shared component?

paulkerschen: Surely it could.

paulkerschen:
With carets like these, who needs sticks?

PR #1116, opened 2019-01-05
BOAC-1608/BOAC-1609, Cohort view > proper results when user clicks 'Apply'; create cohort

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1608
https://jira.ets.berkeley.edu/jira/browse/BOAC-1609

* Include `student_count` when `create_cohort`
* `/api/cohort/create` accepts an array of filters, as seen in UI
* click 'Apply' and only the students section of the page gets a spinner

### DIFF
--- a/boac/api/cohort_controller.py
+++ b/boac/api/cohort_controller.py
@@ -85,11 +85,12 @@ def students_with_alerts(cohort_id):
 @app.route('/api/cohort/<cohort_id>')
 @login_required
 def get_cohort(cohort_id):
-    if is_unauthorized_search(request.args):
+    filter_keys = list(request.args.keys())
+    order_by = get_param(request.args, 'orderBy', None)
+    if is_unauthorized_search(filter_keys, order_by):
         raise ForbiddenRequestError('You are unauthorized to access student data managed by other departments')
     include_students = to_bool(get_param(request.args, 'includeStudents'))
     include_students = True if include_students is None else include_students
-    order_by = get_param(request.args, 'orderBy', None)
     offset = get_param(request.args, 'offset', 0)
     limit = get_param(request.args, 'limit', 50)
     cohort = CohortFilter.find_by_id(int(cohort_id))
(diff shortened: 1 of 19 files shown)


boac/api/cohort_controller.py
```diff
@@ -107,35 +109,75 @@ def get_cohort(cohort_id):
...
+            'groupCodes',
+            'lastNameRange',
+            'levels',
+            'majors',
+            'underrepresented',
+            'unitRanges',
+        ]:
+            filter_criteria[key] = get_param(params, key)
+        for key in ['inIntensiveCohort', 'isInactiveAsc', 'isInactiveCoe']:
+            filter_criteria[key] = to_bool(params.get(key))
```
johncrossman: The `else` block above will be removed when Vue cutover is done. Going forward, the front-end only deals with a list of 'filters' and is not asked to deal with `filter_criteria` dict as stored in db.

boac/models/cohort_filter.py
```diff
@@ -65,15 +65,11 @@ def __repr__(self):
...
     @classmethod
-    def create(cls, uid, name, **kwargs):
-        at_least_one_is_defined = False
-        filter_criteria = {}
-        for k, v in kwargs.items():
-            at_least_one_is_defined = at_least_one_is_defined or (len(v) if isinstance(v, list) else v is not None)
-            filter_criteria[util.camelize(k)] = v
-        if not at_least_one_is_defined:
+    def create(cls, uid, name, filter_criteria, student_count=None):
+        if all(not value for value in filter_criteria.values()):
```
paulkerschen: @johncrossman Will this logic get tripped up if searching on criteria with value `False`?

johncrossman: Yes, good catch. I'll amend

paulkerschen:
Thanks for the array-tidying on the filters, @johncrossman.

PR #1117, opened 2019-01-07
BOAC-1613, use predictable ng-change on curated checkbox (student profile)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1613

https://docs.angularjs.org/api/ng/directive/ngChange says, "ngChange is only evaluated when ... new value to be committed to the model". I verified on Chrome and Safari.


raydavis:
It must've hurt to do an Angular fix at this point...

PR #1121, opened 2019-01-09
BOAC-1611, FilterRow.vue for existing and new filters

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1611

As long as we keep `FilterRow` readable, this is a more intuitive arrangement.


package-lock.json
```diff
@@ -2248,7 +2248,7 @@
     },
     "array-equal": {
       "version": "1.0.0",
-      "resolved": "http://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
+      "resolved": "https://registry.npmjs.org/array-equal/-/array-equal-1.0.0.tgz",
```
paulkerschen: More fun with https://npm.community/t/some-packages-have-dist-tarball-as-http-and-not-https/285/49 ... I believe my machine was the most recent offender on this.

paulkerschen:
Makes sense to me!

PR #1123, opened 2019-01-09
BOAC-1350 Purple holds

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1350

Note front-end implementation in both Vue and Angular.

screen shot 2019-01-09 at 1 13 39 pm


johncrossman:
A delicious UI.
 
image

PR #1137, opened 2019-01-10
BOAC-1640, on /cohort page, menu option 'Level' can be wrongly disabled

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1640


boac/lib/cohort_filter_definition.py
```diff
@@ -284,9 +284,12 @@ def _keys_of_type_boolean(rows):
 def _selections_of_type_array(rows):
     existing_array_rows = list(filter(lambda row: row['type'] in ['array'], rows))
     unique_keys = set(map(lambda row: row['key'], existing_array_rows))
-    selections = dict.fromkeys(unique_keys, [])
+    selections = dict.fromkeys(unique_keys)
```
johncrossman: The cause o' bug was in line above – all keys pointed to same array instance. See test coverage to understand details.

paulkerschen:
Sneaky!

PR #1139, opened 2019-01-10
BOAC-1639: Restricts home page filtered cohort view to only students with alerts

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1639


paulkerschen:
We are alerted.

PR #1145, opened 2019-01-11
BOAC-1598 Drop teams page from Vue

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1598


johncrossman:
Teams drop players and developers drop teams.

PR #1157, opened 2019-01-13
BOAC-1646, sortBy user preference shared by /cohort and /curated

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1646

The `user` store now has `preferences`, where we keep user's last sortBy choice. I got rid of `Students` component cuz `/cohort` and `/curated` pages had different UI needs.


lyttam:
It's nice to see the components sharing.

PR #1158, opened 2019-01-14
BOAC-1461: improves home page screen reader subtext

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1461


paulkerschen:
The screenreader swears to tell nothing but the truth.

PR #1162, opened 2019-01-14
BOAC-1638, 'undefined.length' never ends well, use _.size

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1638


paulkerschen:
It's, like, length isn't even a _property_ here, man.

PR #1176, opened 2019-01-16
BOAC-1654: Updates copyright to 2019

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1654


johncrossman:
Is 2019 BOAC's year? It is now.

PR #1178, opened 2019-01-16
BOAC-1612 Featured student appears on any page in class list

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1612

Many of the code changes are just to ensure that you can't type random UIDs into the address bar and have those students show up.

Proper test coverage for this kind of pagination-related logic might require some easier way to generate fake students in quantity.


johncrossman:
Thank you for protecting against URL hacks, @paulkerschen

PR #1182, opened 2019-01-17
BOAC-1664, create_cohort from sidebar must reload /cohort

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1664

Sidebar router-links that want to force reload now have a way.


paulkerschen:
It's like angular-ui-router except it works!

PR #1196, opened 2019-01-18
BOAC-1675, do not remove user from Vuex store just before /logout

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1675

Includes a little cleanup in code style.

### DIFF
--- a/boac/api/auth_controller.py
+++ b/boac/api/auth_controller.py
@@ -23,7 +23,6 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-
 from urllib.parse import urlencode
 
 from boac.api.errors import ForbiddenRequestError, ResourceNotFoundError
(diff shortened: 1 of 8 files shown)


boac/api/auth_controller.py
```diff
@@ -87,7 +86,9 @@ def become():
 @login_required
 def logout():
     logout_user()
-    cas_logout_url = _cas_client().get_logout_url(redirect_url=request.url_root)
+    vue_base_url = app.config['VUE_LOCALHOST_BASE_URL']
+    redirect_url = vue_base_url if app.config['VUE_ENABLED'] and vue_base_url else request.url_root
+    cas_logout_url = _cas_client().get_logout_url(redirect_url=redirect_url)
```
johncrossman: In local env, this redirect was sending you to port 5000 which would then redirect you to 8080 and so on. Such wackiness compounds the problem. The console has more time to complain about null user object.

src/store/modules/user.ts
```diff
@@ -97,7 +82,12 @@ const actions = {
         });
       }
     });
-  }
+  },
+  logout: ({ commit }) => commit('logout'),
+  setDemoMode: ({ commit }, demoMode) => commit('setDemoMode', demoMode),
+  setUserPreference: ({ commit }, { key, value }) =>
+    commit('setUserPreference', { key, value }),
+  userAuthenticated: ({ commit }) => commit('userAuthenticated')
```
johncrossman: Alpha order and arrow functions. No change in logic.

paulkerschen:
No need to terminate the user; redirect does our dirty work for us.

PR #1198, opened 2019-01-18
BOAC-1671, key=route.fullPath reloads view; sidebar paths are unique per current time (ms)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1671

Sidebar links to `/create`, `/cohort` and `/curated` will always cause `router-view` reload. I also verified pagination on course and cohort pages.

### DIFF
--- a/src/store/modules/user.ts
+++ b/src/store/modules/user.ts
@@ -31,10 +31,6 @@ const getters = {
 };
 
 const mutations = {
-  logout: (state: any) => {
-    state.isUserAuthenticated = false;
-    state.user = null;
-  },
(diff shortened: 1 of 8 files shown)


src/store/modules/user.ts
```diff
@@ -31,10 +31,6 @@ const getters = {
 };
 
 const mutations = {
-  logout: (state: any) => {
-    state.isUserAuthenticated = false;
-    state.user = null;
-  },
```
johncrossman: This is unused, should have been removed with PR #1196

PR #1199, opened 2019-01-18
BOAC-1219: prevents home page total issues count from wrapping

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1219

### DIFF
--- a/boac/static/app/home/home.css
+++ b/boac/static/app/home/home.css
@@ -37,6 +37,7 @@
   align-items: center;
   display: flex;
   margin: 10px 15px;
+  min-width: 130px;
 }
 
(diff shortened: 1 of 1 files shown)


boac/static/app/home/home.css
```diff
@@ -37,6 +37,7 @@
   align-items: center;
   display: flex;
   margin: 10px 15px;
+  min-width: 130px;
```
johncrossman: @lyttam Fyi, the bootstrap [utility css classes](https://bootstrap-vue.js.org/docs/reference/utility-classes/), which we have access to via vue-bootstrap, include "text-nowrap".  But this fix works, too.

PR #1204, opened 2019-01-19
BOAC-1320, if user edits filter then auto-cancel other filter row edit

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1320

I'm tossing out the use of `this.$eventHub.$on` when it comes to state management in filter rows – each row allows add/edit/remove of filter – in favor of a more intuitive use of `EDIT_MODE_TYPES`.  (We keep `editMode` value in Vuex store.)

### DIFF
--- a/src/components/cohort/FilterRow.vue
+++ b/src/components/cohort/FilterRow.vue
@@ -3,7 +3,7 @@
        :class="{'pt-2': !isExistingFilter}"
        v-if="showRow">
     <div :id="`existing-name-${index}`"
-         class="dropdown-width p-2"
+         class="existing-filter-name p-2"
          v-if="isExistingFilter">
       <span class="sr-only">Filter name:</span> {{ filter.name }}
     </div>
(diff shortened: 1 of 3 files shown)


src/components/cohort/FilterRow.vue
```diff
@@ -203,11 +203,6 @@ export default {
   created() {
     this.reset();
     this.valueOriginal = this.filter && this.filter.value;
-    this.$eventHub.$on('cohort-filter-row-edit', editIndex => {
-      if (this.index && this.index !== editIndex) {
-        this.cancelEditExisting();
-      }
-    });
```
johncrossman: IMO, getting rid of the bit above is important to the maintainability of this FilterRow code. The event-listening is now solely kept in the `watch` block below, where I've added a little inline documentation.

paulkerschen:
Makes sense; thanks for the inline comments.

PR #1210, opened 2019-01-22
BOAC-1691, render view only after not loading and simplify Search view/component

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1691

`Search.vue` was one of the first views cut to Vue and it shows.  Use of `beforeRouteUpdate()` is not necessary if we force unique path from the sidebar. 

Use `?w=1` to get simpler diff.

### DIFF
--- a/src/components/curated/CuratedGroupHeader.vue
+++ b/src/components/curated/CuratedGroupHeader.vue
@@ -77,6 +77,7 @@
           </div>
           <div slot="modal-footer">
             <b-btn id="delete-confirm"
+                   class="btn-primary-color-override"
                    variant="primary"
                    @click="deleteGroup">
               Delete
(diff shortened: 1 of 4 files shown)


src/components/curated/CuratedGroupHeader.vue
```diff
@@ -77,6 +77,7 @@
           </div>
           <div slot="modal-footer">
             <b-btn id="delete-confirm"
+                   class="btn-primary-color-override"
```
johncrossman: Fix for BOAC-1685 which should have been in PR #1206

src/mixins/Loading.vue
```diff
@@ -9,7 +9,7 @@ export default {
     ...mapGetters('context', ['loading'])
   },
   methods: {
-    startLoading: () => store.dispatch('context/loadingStart'),
+    ...mapActions('context', ['loadingStart']),
```
johncrossman: We had only one usage of `startLoading` (above), which has been refactored to `loadingStart`

PR #1212, opened 2019-01-23
[QA] BOAC-1696, /create_cohort is wrong; /cohort/new is right

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1696


lyttam:
Right is best.

PR #1217, opened 2019-01-23
BOAC-1695, goodbye v-focus; BOAC-1575, accessibility on /search and /course

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1695
https://jira.ets.berkeley.edu/jira/browse/BOAC-1686
https://jira.ets.berkeley.edu/jira/browse/BOAC-1575

Use `putFocusNextTick()` instead of `v-focus`. Plus, heading structure and element ids.


src/views/Search.vue
```diff
@@ -13,10 +13,9 @@
         <li>Abbreviations of section titles may not return results; <strong>COMPSCI 161</strong> instead of <strong>CS 161</strong>.</li>
       </ul>
     </div>
-    <div id="page-header"
-         tabindex="0"
+    <div tabindex="0"
          v-if="!loading && results.totalStudentCount">
-      <h1 ref="pageHeader">{{ 'student' | pluralize(results.totalStudentCount) }} matching '{{ phrase }}'</h1>
```
paulkerschen: @johncrossman Not that it's easy to tell from the name, but the `pageHeader` ref, here and in a few other pages, has the effect of setting focus after load, courtesy of the `Loading` mixin. Your newer code probably gives us a clearer way to accomplish this, but if so, we might need to add some explicit focus-setting to replace.

johncrossman: Good catch, @paulkerschen  Actually, the explicit focus-setting logic is there, at the bottom of Search.vue:
```
          this.loaded();
          const focusId =
            this.results.totalCourseCount || this.results.totalStudentCount
              ? 'page-header'
              : 'page-header-no-results';
          this.putFocusNextTick(focusId);
```

paulkerschen: Thanks for the pointer!

PR #1224, opened 2019-01-24
BOAC-1704, set page=1 before applying a new set of cohort filters

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1704


paulkerschen:
Begin the begin.

PR #1229, opened 2019-01-24
BOAC-1703: enables pagination arrows and scopes styles

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1703

The Bootstrap-Vue pagination widget works differently than Angular's - the ellipsis is not meant to be clickable.  Instead, pages 11-_n_ can be accessed via the arrow buttons (first, previous, next, and last).

I also realized that unless `scoped` is specified, anything in the `<style>...</style>` section of a component applies to the entire app.

### DIFF
--- a/src/components/curated/CuratedStudentCheckbox.vue
+++ b/src/components/curated/CuratedStudentCheckbox.vue
@@ -42,9 +42,3 @@ export default {
   }
 };
 </script>
-
-<style>
-.form-check-inline {
-  margin: 0;
-}
-</style>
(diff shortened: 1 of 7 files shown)


paulkerschen:
@lyttam Does custom styling d3 and Highcharts elements (matrix view and tooltips, GPA/units charts and tooltips) still work with the styles scoped? When I was setting up the integrations, I was finding that, since the styles affect elements outside the components created by third-party integrations, they only seemed to work when unscoped.

lyttam:
Good call, @paulkerschen - I was overzealous in applying `scoped`.  I've removed it from the components containing styles for third-party elements or elements that live in another component.

paulkerschen:
Thanks, @lyttam!

PR #1233, opened 2019-01-24
BOAC-1719, Admin page avatar to verify demoMode blur

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1719
image

image


paulkerschen:
There's no other way.

PR #1234, opened 2019-01-24
BOAC-1721, set cohort.alert_count on cohort create/update

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1721


paulkerschen:
Persistence is hard!

PR #1238, opened 2019-01-25
BOAC-1603: redirects to requested route after login

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1603

Redirect to login when an unauthenticated user requests a protected route was already working, but now the user will be re-redirected to the protected route after login.


paulkerschen:
@lyttam Code change looks good but did an empty "oops" commit sneak in there?

lyttam:
@paulkerschen yet, forgot to "fixup". Amended.

PR #1239, opened 2019-01-28
BOAC-1643 Add physics dept to advising groups

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1643

I thought this was going to be more complicated, but apparently we designed it right the first time. Creation of department and user records in the database will be handled via scripts under release instructions.


raydavis:
And a free delinting to boot.

PR #1241, opened 2019-01-28
BOAC-1681 Erase traces of Angular in routes, configs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1681


johncrossman:
Such satisfying deletions.

PR #1242, opened 2019-01-29
BOAC-1723, structural change to /api/alerts json: 'dismissed' prop on all alert objects

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1723

Suitable to new UX paradigm.


paulkerschen:
Dismissed, but never dismal.

PR #1245, opened 2019-01-29
BOAC-1732 Do not seek to plot the unplottable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1732


johncrossman:
It is plottable, except when it's not.

PR #1250, opened 2019-01-30
BOAC-1683 Remove legacy.css, parceling out as appropriate

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1683

The global grab bag `legacy.css` is distributed among many smaller bags. Note the consequence that more components now include `<style scoped>` for their own styling plus `<style>` for styles that need to be shared. I expect our conventions will continue to evolve.


johncrossman:
Well parcelled. Waiting on T.

PR #1258, opened 2019-02-01
BOAC-1751, move global styles to boac-global.css and bootstrap-overrides.css

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1751


paulkerschen:
I feel better already!

PR #1261, opened 2019-02-01
BOAC-1758 Fix error before state.config initialized

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1758


johncrossman:
Thanks @paulkerschen ;)

PR #1268, opened 2019-02-01
BOAC-1729: makes home and skip links keyboard-friendly

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1729


johncrossman:
Hey keyboard, let's be friends.

PR #1271, opened 2019-02-04
BOAC-1742, strip anchor ref in router path with split, not trimEnd

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1742

PR #1269 was not a fix.

### DIFF
--- a/src/layouts/StandardLayout.vue
+++ b/src/layouts/StandardLayout.vue
@@ -9,8 +9,8 @@
         <div class="header-text">
           <h1>
             <router-link to="/home"
-                       id="home-header"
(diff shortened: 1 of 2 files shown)


src/layouts/StandardLayout.vue
```diff
@@ -51,7 +51,7 @@ export default {
   },
   methods: {
     stripAnchorRef(fullPath) {
-      return this.trimEnd(fullPath, '#');
+      return this.split(fullPath, '#', 1)[0];
```
johncrossman: Btw, this is null-safe. Eg, `_.split(null, '#', 1)[0]` returns empty string.

PR #1275, opened 2019-02-06
BOAC-1738: improves accessibility of cohort filters

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1738


johncrossman:
Nice work @lyttam

PR #1284, opened 2019-02-09
BOAC-1788, refactor to put alerts/holds/etc. under student['notifications']

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1788

A substantial refactor allowed for a lot of code removal and faster page load. Plus, having alerts/holds/etc. bundled in one spot is more intuitive.


raydavis:
Thanks, @johncrossman

PR #1288, opened 2019-02-11
BOAC-1798, message show/hide fix; order of messages; fall back to createdAt

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1798

Having `isOpen` on each message was thwarted by use of computed property: DOM does not re-render when props of objects in list are modified.


paulkerschen:
Thanks @johncrossman - that's some delicate Vue-ing.

PR #1292, opened 2019-02-12
BOAC-1774, add notes (read-only) to Academic Timeline

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1774

The loch query will evolve.


paulkerschen:
Looking tidy.

PR #1295, opened 2019-02-12
BOAC-1800 Let the ternary operator operate

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1800


johncrossman:
You be you, ternary

PR #1299, opened 2019-02-12
BOAC-1806, Academic Timeline: tabs always present; notes as html with min-height

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1796
https://jira.ets.berkeley.edu/jira/browse/BOAC-1806

Tabs with zero are disabled. Open a one-line message and min-height style forces row to expand a bit.


paulkerschen:
Tabulous.

PR #1305, opened 2019-02-14
BOAC-1812, Academic Timeline: proper padding in right column (dates)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1812


paulkerschen:
This pretty padding is perfectly proper.

PR #1311, opened 2019-02-14
BOAC-1705, vue and vuex upgrade; stricter linting per vuejs.org style-guide

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1705

Vue template linting as prescribed in https://vuejs.org/v2/style-guide/  I removed 'prettier' due to eslint rule conflicts.  Review with `?w=1`


src/components/HeaderMenu.vue
```diff
@@ -3,11 +3,12 @@
...
       <span v-if="user && !user.firstName">{{ user }}</span>
-      <b-dropdown id="header-dropdown-under-name"
-                  class="mr-3"
-                  variant="link"
-                  no-caret
-                  right>
+      <b-dropdown
+        id="header-dropdown-under-name"
+        class="mr-3"
+        variant="link"
```
johncrossman: The indentation rule above is the most significant change to our code style.

paulkerschen: That's definitely an easier indentation rule to deal with, especially when the element names get long.

.eslintrc.js
```diff
@@ -5,23 +5,27 @@ module.exports = {
...
   },
   extends: [
-    '@vue/prettier',
     '@vue/typescript',
-    'eslint:recommended',
-    'plugin:vue/essential'
+    'plugin:vue/recommended',
+    'eslint:recommended'
   ],
-  plugins: ['vue', 'prettier'],
```
paulkerschen: This is subjective, but I didn't find `prettier` to be all that pretty and don't think I'll miss it.

paulkerschen:
Thanks for this, @johncrossman - these seem like sensible rules. Given the size of the diff, I'll give someone else the chance to second.

johncrossman:
Thank you @paulkerschen.  (Maybe @lyttam is available as a second pair of eyes.)

lyttam:
Looks good, thanks @johncrossman!

PR #1315, opened 2019-02-15
BOAC-1705, fix linting 'warns' per stricter linting

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1705

Follow up to PR #1311


paulkerschen:
Oh good, I was feeling called out by those warns.

PR #1320, opened 2019-02-20
BOAC-1737, proper error messages on /login

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1737

Simplified some code. Improved wording in errors. Popover instead of modal. Hold on to error when `router.ts` does redirect.


paulkerschen:
I didn't even know we had a support email in the configs; guess we're a real project.

PR #1324, opened 2019-02-20
BOAC-1710 Report the state of assignment scores in the current term

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1710


paulkerschen:
Thanks, @raydavis, for handling this comparison of apples to oranges, guavas, and durian.

PR #1328, opened 2019-02-21
BOAC-1809: enables deeplinks for dev auth

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1809

If unauthenticated user hits a protected route (e.g. /student/999), lands on /login, and authenticates using dev auth, they will be conveniently redirected to /student/999 instead of /home.


johncrossman:
The devAuth'ers thank you @lyttam

raydavis:
Polished!

PR #1331, opened 2019-02-22
BOAC-1844, StudentRow id elements with sid<->rowIndex translation for Teena

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1844

Teena no like XPath.

### DIFF
--- a/src/components/student/StudentRow.vue
+++ b/src/components/student/StudentRow.vue
@@ -1,9 +1,11 @@
 <template>
   <div>
+    <span :id="`row-index-of-${student.sid}`" hidden aria-hidden="true">{{ rowIndex }}</span>
+    <span :id="`student-sid-of-row-${rowIndex}`" hidden aria-hidden="true">{{ student.sid }}</span>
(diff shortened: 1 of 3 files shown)


src/components/student/StudentRow.vue
```diff
@@ -138,12 +142,12 @@
         </tr>
         <tr v-for="(enrollment, index) in get(student.term, 'enrollments', [])" :key="index">
           <td class="cohort-course-activity-data cohort-course-activity-course-name">
-            <div>{{ enrollment.displayName }}</div>
+            <div :id="`row-${rowIndex}-student-enrollment-name`">{{ enrollment.displayName }}</div>
```
lyttam: @johncrossman if we don't use `index` as well as `rowIndex` here, will this result in multiple `<div>` elements with the same id?

johncrossman: Good catch, @lyttam I'll fix in subsequent PR

PR #1333, opened 2019-02-23
BOAC-1839, initial NewNoteModal with API support; db model fixes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1839

`AcademicTimeline.vue` changes are minor – review with `?w=1`. Not using Bootstrap modal – code is more manageable if we roll-our-own.


paulkerschen:
I said, that’s nothing, you should hear me play piano.

PR #1343, opened 2019-02-26
BOAC-1797 Keep cohort sort, pagination in sync on filter update

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1797


raydavis:
Metamixil!

PR #1344, opened 2019-02-27
BOAC-1768, CalNet queries: widen search_base; 'ou' restriction on filter

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1768

See Jira for details. Plus, code cleanup for readability.


paulkerschen:
A wider base is always more stable.

PR #1345, opened 2019-02-28
BOAC-1692 Buttons to reset filter edits to last apply or save

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1692

Tricky because the nature of "resetting" depends on context. When you've made changes to filters but not applied them, presumably "Reset" should take you back to the last apply. But when you've applied changes but not saved them, presumably "Reset" should take you back to the last save point.


johncrossman:
Indeed, delicate surgery. Thx @paulkerschen

PR #1346, opened 2019-02-28
BOAC-1880, notes table captures advisor's current name, role, and depts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1880

I will drop/create the `notes` table if/when merged. (QA approved the dropping of test data in dev.) The `/api/user/{id}` endpoint is not `get_by_uid`.

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -24,9 +24,9 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """
 
 from boac.api import errors
-from boac.api.util import admin_required, authorized_users_api_feed, get_current_user_status, get_my_cohorts, get_my_curated_groups
+from boac.api.util import admin_required, authorized_users_api_feed, current_user_profile, get_current_user_status
 from boac.lib import util
-from boac.lib.berkeley import BERKELEY_DEPT_NAME_TO_CODE, get_dept_codes
+from boac.lib.berkeley import BERKELEY_DEPT_CODE_TO_NAME
 from boac.lib.http import tolerant_jsonify
 from boac.merged import calnet
 from boac.models.authorized_user import AuthorizedUser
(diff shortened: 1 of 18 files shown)


boac/api/user_controller.py
```diff
@@ -96,17 +69,21 @@ def user_by_id(user_id):
...
-        depts[dept_code] = {
-            'code': dept_code,
-            'name': dept_name,
-            'users': [],
-        }
+
+    def _put(_dept_code, _user):
+        if _dept_code not in depts:
+            dept_name = 'Admins' if _dept_code == 'ADMIN' else BERKELEY_DEPT_CODE_TO_NAME.get(_dept_code)
+            depts[_dept_code] = {
```
johncrossman: This is used on Admin page; we only serve depts that have 1+ users. Some `dept_codes` (see the added 'QCADV' below) might map to zero users.

src/components/note/AdvisingNote.vue
```diff
@@ -1,32 +1,47 @@
...
+          target="_blank">{{ getName(author) }}</a>
+        <span v-if="author.role">
+          - <span :id="`note-${note.id}-author-role`" class="text-dark">{{ author.role }}</span>
+        </span>
+      </div>
+      <div v-if="size(author.depts)" class="text-secondary">
+        <span v-if="note.isLegacy">(currently </span><span v-for="(dept, index) in author.depts" :key="dept">
+          <span :id="`note-${note.id}-author-dept-${index}`">{{ dept }}</span>
+        </span><span v-if="note.isLegacy">)</span>
+      </div>
```
johncrossman: If BOAC recognizes the `dept_code` then legacy notes will display as below.
image

paulkerschen:
Thanks, @johncrossman. I was going back and forth about having advisor department and role right there in the notes table instead of somewhere else, but I came to the same conclusion you presumably did - that because advisors can change affiliation over time, this denormalized setup is the easiest one to deal with.

PR #1347, opened 2019-02-28
BOAC-1882 Move advising notes into their own merged model

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1882

Before changing behavior, this is a good moment to organize the garage.


johncrossman:
Yes and yes! Entropy reduced.

PR #1350, opened 2019-02-28
BOAC-1872, remove sid from gaNoteEvent and this.gaStudentAlert

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1872

On second thought, let's not push sid to GA.


lyttam:
Seems wise.

PR #1354, opened 2019-03-01
BOAC-1896, sorry Admin, no note creation for you

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1896

Plus, id tweaks for Teena


pfarestveit:
Thanks @johncrossman .  Hope this does the trick!

PR #1363, opened 2019-03-05
BOAC-1915 Blur search results in demo mode

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1915

Needs to be applied in a couple of different places to catch all the elements.


lyttam:
Looks good and blurred.

PR #1364, opened 2019-03-06
BOAC-1902: Tweaks enrollment term display rules

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1902


johncrossman:
"At BOA, we give you what's relevant."

PR #1365, opened 2019-03-06
BOAC-1909, if dev_auth=True then advisors can toggle demo_mode, too

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1909

A good enhancement to have in the upcoming release, specifically in _boac-demo_ env. We don't want to rely on `boac-dev`.

### DIFF
--- a/src/views/Admin.vue
+++ b/src/views/Admin.vue
@@ -7,7 +7,7 @@
         aria-live="passive"
         class="sr-only">Admin page loaded</span>
       <h1>BOAC Flight Deck</h1>
-      <div v-if="user.isAdmin && devAuthEnabled" class="d-flex flex-row mt-3 mb-3">
+      <div v-if="devAuthEnabled" class="d-flex flex-row mt-3 mb-3">
         <div class="mr-3">
           <img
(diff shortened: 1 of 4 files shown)


src/views/Admin.vue
```diff
@@ -7,7 +7,7 @@
         aria-live="passive"
         class="sr-only">Admin page loaded</span>
       <h1>BOAC Flight Deck</h1>
-      <div v-if="user.isAdmin && devAuthEnabled" class="d-flex flex-row mt-3 mb-3">
+      <div v-if="devAuthEnabled" class="d-flex flex-row mt-3 mb-3">
```
lyttam: @johncrossman shouldn't admin users be allowed to see this even when dev auth is not enabled?

johncrossman: @lyttam If `devAuth` enabled then all users, including admins, will have access to demo-mode. I've relaxed the condition.

lyttam: I see; the demo mode toggle is not currently available in prod.

PR #1370, opened 2019-03-06
BOAC-1914 Friendly filenames for CS-sourced attachments

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1914

The predictably formatted `attachment.sisFilename` is always used for API calls. We give display preference to `attachment.userFilename` when available (non-UCBCONVERSION attachments).


johncrossman:
At BOA, we do friendly.

PR #1376, opened 2019-03-08
BOAC-1706, update py and package.json deps

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1706

I kept `boto3` upgrade in the `1.7.*` range and no upgrade for `moto`.  Those two were not playing nice earlier.


paulkerschen:
We could always try replacing Moto with Motö.

PR #1379, opened 2019-03-09
BOAC-1916, step 1 in the 'are you sure?' in cancel Note

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1916

### DIFF
--- a/src/components/note/NewNoteModal.vue
+++ b/src/components/note/NewNoteModal.vue
@@ -1,5 +1,5 @@
 <template>
-  <div>
+  <div v-click-outside="clickOutside">
     <div id="sr-alert-new-note" class="sr-only" aria-live="polite">{{ screenReaderAlert }}</div>
     <div>
       <b-btn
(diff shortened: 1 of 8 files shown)


src/components/note/NewNoteModal.vue
```diff
@@ -100,7 +100,7 @@
           aria-label="Bring create-note form into view"
           variant="link"
           class="pr-2"
-          @click.prevent="maximize()">
+          @click.stop="maximize()">
```
paulkerschen: @johncrossman I see `maximize` is invoked with parentheses but `minimize` isn't. Intended?

src/mixins/Util.vue
```diff
@@ -49,6 +49,11 @@ export default {
     size: _.size,
     slice: _.slice,
     split: _.split,
+    stripHtmlAndTrim: (html) => {
+      let text = html && html.replace(/(<([^>]+)>)/ig,"");
```
paulkerschen: @johncrossman I think parentheses inside the angle brackets aren't needed here since we aren't doing anything with the capture group.

johncrossman:
Thanks @paulkerschen  Amendment in, parens added.

PR #1380, opened 2019-03-11
NS-429 Remove vestiges of term-wide analytics averaging

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-429

BOAC no longer attempts to show term-wide averages of analytics data, but the underlying numbers were still being supplied by Nessie until https://github.com/ets-berkeley-edu/nessie/pull/371.


johncrossman:
Less is more.

PR #1382, opened 2019-03-11
BOAC-1934, if feature flag false then turn off create and edit notes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1934

See comment in https://jira.ets.berkeley.edu/jira/browse/BOAC-1949


paulkerschen:
Emergency brakes are important.

PR #1385, opened 2019-03-13
BOAC-1734: Fixes screen reader alerts for table sorting

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1734

### DIFF
--- a/src/components/course/SortableCourseList.vue
+++ b/src/components/course/SortableCourseList.vue
@@ -17,19 +17,13 @@
         Showing the first {{ courses.length }} classes.
       </div>
     </div>
-    <div
-      v-if="totalCourseCount && resorted"
-      class="sr-only"
-      role="alert">
-      Courses sorted by {{ sort.by === 'section' ? 'section' : 'course name' }} {{ describeReverse(sort.reverse.section) }}
-    </div>
     <table v-if="totalCourseCount" class="table-full-width">
       <tr role="row">
         <th>
           <button
             id="column-sort-button-section"
             class="btn btn-link table-header-text sortable-table-header cursor-pointer table-cell"
-            :aria-label="`Sort by section ${ sort.by === 'section' ? describeReverse(sort.reverse.section) : ''}`"
+            :aria-label="`Sort by section ${describeReverse(sort.reverse.section)}`"
             tabindex="0"
             @click="courseSort('section')">
             Section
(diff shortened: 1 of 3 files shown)


src/components/course/SortableCourseList.vue
```diff
@@ -160,7 +165,7 @@ export default {
       let split = course.courseName.split(' ');
       return [split.slice(0, -1).join(' '), split[split.length - 1]];
     },
-    describeReverse: reverse => (reverse ? 'descending' : 'ascending')
+    describeReverse: reverse => (reverse ? 'descending' : '')
```
lyttam: I decided to remove 'ascending' because it sounds too similar to 'descending' when NVDA pronounces it, and to match how the SortableStudents component does it.

src/components/course/SortableCourseList.vue
```diff
@@ -20,24 +20,24 @@
     <div
       v-if="totalCourseCount && resorted"
       class="sr-only"
-      role="alert">
-      Courses sorted by {{ sort.by === 'section' ? 'section' : 'course name' }} {{ describeReverse(sort.reverse.section) }}
+      :role="srAlert ? 'alert' : ''">
+      Courses sorted by {{ sort.by === 'section' ? 'section' : 'course name' }} {{ describeReverse(sort.reverse[sort.by]) }}
```
johncrossman: @lyttam I'm curious, why did you switch to brackets?

src/components/search/SortableStudents.vue
```diff
@@ -243,6 +243,7 @@ export default {
   },
   data: () => ({
     currentSortDescription: null,
+    srAlert: false,
```
johncrossman: @lyttam The `Context` mixin has `srAlert` getter. I would expect an error in the console due to conflict w/ the prop above. Is there an easy/better way to leverage the mixin – which also has an `alertScreenReader` action method – in order to announce sort change?

lyttam: @johncrossman great idea, I'll take a look at that method.

src/components/course/SortableCourseList.vue
lyttam: Not necessary; I'll switch it back.

lyttam:
@johncrossman I incorporated your suggestions.

I'm curious what others think about adding `role="alert"` to the screen reader alert in `StandardLayout`.  I wasn't able to get NVDA on FireFox to voice the sorting alert without it.

`role="alert"` overrides `aria-live="polite"` and has the same effect as adding `aria-live="assertive"` and `aria-atomic="true"`. It appears to have good browser and screen reader support, but if some combo of browser+screen reader doesn't support it, we would fall back on `aria-live="polite"`.

Is there a good reason our screen reader alert is coded as polite rather than assertive?  Seems to me that if we limit to one alert message per page and use it sparingly, it should be ok to use `role="alert"`.

johncrossman:
@lyttam - Your summary is convincing. (Thank you for educating me!)  I say we add `role="alert"`.

PR #1386, opened 2019-03-13
BOAC-1911, delete note (available to admins), front-end support

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1911

* use `?w=1` for clearer diff
* hidden delete-button for accessibility
* element `id` changes per pending Teena PR


paulkerschen:
I'm glad to see Teena moving into the development space.

PR #1388, opened 2019-03-18
BOAC-1965, highcharts 6.2.0 to move past security vulnerability

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1965

Bonus: remove test on legacy index.html


paulkerschen:
Charts fly higher.

PR #1389, opened 2019-03-19
BOAC-1958, equivalent data structure in note author and user object from LDAP

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1958
https://jira.ets.berkeley.edu/jira/browse/NS-417

More metadata from LDAP for legacy notes. Standard user object structure in 'departments' – a precursor to what we'll store on Nessie side.


paulkerschen:
Looks good!

PR #1393, opened 2019-03-19
BOAC-1969, decouple dev-auth and demo-mode

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1969

* further use of `override_config`
* review with `?w=1` for a simpler diff


lyttam:
Nicely done.

PR #1395, opened 2019-03-20
BOAC-1978, notes table gets deleted_at column; no hard deletes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1978


paulkerschen:
No one wins when SQLAlchemy magic fights flake8 magic, but I approve this code.

PR #1396, opened 2019-03-21
BOAC-1816: Makes pages wait for user data before finishing loading

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1816

On pages that have a component or sub-component reliant on the `user` object, the spinner will spin until the `api/profile/my` response is received.

### DIFF
--- a/src/mixins/GoogleAnalytics.vue
+++ b/src/mixins/GoogleAnalytics.vue
@@ -1,15 +1,25 @@
 <script>
-import store from '@/store';
 import { event } from 'vue-analytics';
+import { mapGetters } from 'vuex';
 
 export default {
   name: 'GoogleAnalytics',
+  computed: {
+    ...mapGetters('user', ['user'])
+  },
   methods: {
     gaEvent(category, action, label, value) {
-      const user = store.getters['user/user'];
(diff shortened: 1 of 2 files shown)


src/views/Admin.vue
```diff
@@ -96,10 +96,10 @@ export default {
     if (this.userAuthStatus.isAdmin) {
       store.dispatch('user/loadUserGroups').then(data => {
         this.userGroups = data;
-        this.loaded();
+        this.$watch('user', this.loaded);
       });
     } else {
-      this.loaded();
+      this.$watch('user', this.loaded);
```
johncrossman: @lyttam  This makes sense – thank you for the investing time. I think we need one more tweak: this `$watch` should live in the mixin's `loaded()` method – it would be a direct watch on the 'user' store. If we can make that work then we won't need to remember the `$watch` in future code.

lyttam:
@johncrossman Amended. It turns out that if you put a watch on `user` when it's already populated, the callback will never get called (unless `user` changes again).

johncrossman:
Very cool, @lyttam  I've learned something today.

PR #1398, opened 2019-03-25
BOAC-1873 Include locally created notes in search results

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1873

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -351,7 +351,7 @@ def get_advising_note_attachments(sid):
     return safe_execute_redshift(sql, sid=sid)
 
 
-def search_advising_notes(search_phrase, sid_filter, limit=None):
+def search_advising_notes(search_phrase, sid_filter, offset=None, limit=None):
     sql = f"""SELECT
         an.sid, an.id, an.note_body, an.advisor_sid, an.created_by, an.created_at, an.updated_at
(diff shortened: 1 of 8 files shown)


boac/externals/data_loch.py
```diff
@@ -363,10 +363,12 @@ def search_advising_notes(search_phrase, sid_filter, limit=None):
         JOIN {advising_notes_schema()}.advising_notes an
           ON s.id = an.id
           AND an.sid = ANY(:sid_filter)
-        ORDER BY s.rank DESC"""
-    if limit and limit < 100:  # Sanity check large limits
+        ORDER BY s.rank DESC, an.id"""
+    if offset is not None and offset > 0:
+        sql += ' OFFSET :offset'
+    if limit is not None and limit < 150:  # Sanity check large limits
```
johncrossman: @paulkerschen adding `limit > 0` above would be consistent with the offset check – a minor concern

paulkerschen: @johncrossman I thought about that, but you could also see a case where the desired limit actually ends up being zero (that is, if the paged result set is already full of local BOAC notes and there's no room for additional loch notes on the page).

johncrossman:
Search gives something old and something new. Thx @paulkerschen

PR #1399, opened 2019-03-26
BOAC-1967, if create/edit note in progress then 'Are you sure?' if navigating away

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1967

Vuex store used to watch `newNoteMode` and `editingNoteId`.  Only the top-level `Student` can use `beforeRouteLeave()` (which captures the navigate-away event) because it's the component registered in `router.ts`. See docs: https://router.vuejs.org/guide/advanced/navigation-guards.html#in-component-guards

### DIFF
--- a/src/components/note/NewNoteModal.vue
+++ b/src/components/note/NewNoteModal.vue
@@ -5,7 +5,7 @@
         id="new-note-button"
         class="mt-1 mr-2 btn-primary-color-override"
         variant="primary"
-        :disabled="disable || includes(['minimized', 'open'], mode)"
+        :disabled="disable || includes(['minimized', 'open'], newNoteMode)"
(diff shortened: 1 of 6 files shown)


src/components/note/NewNoteModal.vue
```diff
@@ -202,16 +198,11 @@ export default {
       this.error = null;
       this.showErrorPopover = false;
     },
-    clickOutside() {
-      if (this.mode === 'open') {
-        this.cancel();
-      }
-    },
```
johncrossman: The block above is dead code, abandoned when we ditched `v-click-outside`.

paulkerschen:
Begin the begin, cancel the cancel.

PR #1400, opened 2019-03-26
BOAC-1912: Makes note search more forgiving, less powerful

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1912

Using `plainto_tsquery` instead of `to_tsquery` means that logical operators `|`, `&`, and `!` will be ignored - but the tradeoff is you can search for 'Tues 3/28 @ 1:15pm' and it will find the relevant note.  That tradeoff seems worth it to me, but if anyone disagrees we can look at alternatives. 

I changed the regex used to tokenize the search terms and the note result to better handle dates, times, email addresses, decimal numbers (e.g. 2.0) and the like.  It could use further refinement (for example it doesn't handle non-word characters in email addresses, e.g. `anne.lyttle@berkeley.edu`).


paulkerschen:
This looks like a solid improvement to me, @lyttam - thanks! Let's get it deployed and start banging away.

PR #1403, opened 2019-03-28
BOAC-1875: adds SR alert for open/close timeline messages

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1875


raydavis:
Toggling ain't easy.

PR #1408, opened 2019-04-02
BOAC-1877 Do not include ex-advisors in Act As list

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1877

The current logic assumes at a pretty low level that we get faked results from LDAP when real results aren't returned, and so I made the decision optional for now. In a follow-up, I'd like to leave the choice up to higher-level code.


johncrossman:
`force_feed` looks good – as long as foie gras is not involved. Thanks @raydavis

PR #1411, opened 2019-04-03
BOAC-2027 Allow refresh when Vue needs a loaded user profile

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2027

I also have a more focused, Search-only, version of the fix, but @johncrossman thought it might be nice to hook it into the router. What do you think, @lyttam & John C.?

### DIFF
--- a/src/mixins/Loading.vue
+++ b/src/mixins/Loading.vue
@@ -12,13 +12,6 @@ export default {
   methods: {
     ...mapActions('context', ['loadingStart']),
     loaded() {
-      if (this.user) {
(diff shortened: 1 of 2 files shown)


raydavis:
The Git diff for the router change is hard to follow -- probably best to look at the whole file.

johncrossman:
Looks good to me, @raydavis  Welcome to the realm of typescript/Vue.

lyttam:
I like this approach better than using a watch in Loading. Thanks @raydavis!

PR #1412, opened 2019-04-04
BOAC-1917, expand LDAP/CalNet query to include 'expired people'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1917

Broadening our calnet query allows us to put off https://jira.ets.berkeley.edu/jira/browse/NS-417  

Plus, we're evaluating advisor data from SIS. When we have a clearer picture we will reconsider a shadow db table in Nessie-land.

### DIFF
--- a/boac/merged/calnet.py
+++ b/boac/merged/calnet.py
@@ -30,7 +30,10 @@ from boac.models.json_cache import stow
 
 @stow('calnet_user_for_uid_{uid}')
 def get_calnet_user_for_uid(app, uid, force_feed=True):
-    persons = calnet.client(app).search_uids([uid])
+    for search_expired in (False, True):
+        persons = calnet.client(app).search_uids([uid], search_expired)
+        if persons:
+            break
     if not persons and not force_feed:
         return None
     return {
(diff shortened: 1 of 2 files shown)


raydavis:
@johncrossman For this number of IDs and a once-a-day run, an all-in-one LDAP query might work, but "expired" inclusive-searches have always been noticeably slower in my tests. That's why we don't standardly look at them in Junction. It might be worth treating the "expired" search separately as a fallback.

johncrossman:
Makes sense @raydavis I'll amend.

johncrossman:
@raydavis Amended

boac/merged/calnet.py
```diff
@@ -30,7 +30,10 @@
 
 @stow('calnet_user_for_uid_{uid}')
 def get_calnet_user_for_uid(app, uid, force_feed=True):
-    persons = calnet.client(app).search_uids([uid])
```
raydavis: Since our first loop failed to find the UID in ou "people" and "advcon people", do we need to specify them again the second time through?

johncrossman: Good catch. We do not. Please hold...

johncrossman:
Ready for another pass. Thank you for patient review, @raydavis

raydavis:
I have nothing to report.

PR #1413, opened 2019-04-04
BOAC-2013: UTCifies note timestamps from RDS to match Redshift's

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2013


johncrossman:
Looks good @lyttam  The front-end parses the format w/o a problem?  (I would expect so.)

raydavis:
@lyttam You might want to look at https://jira.ets.berkeley.edu/jira/browse/BOAC-1989 as well.

lyttam:
@johncrossman Yes, effectively this just makes sure the '+00:00' is included at the end of the date string so the front-end time zone conversion is done properly.

@raydavis might as well pick that one up while I have my hands dirty with dates.

PR #1416, opened 2019-04-05
BOAC-2005, in eyes of customer, BOAC is now BOA

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2005

Everything under the hood remains BOAC.


raydavis:
The end of an erac.

PR #1417, opened 2019-04-08
BOAC-2053, auth user_loader checks 'isExpiredPerLdap' at login time

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2053

Final piece in forbidding 'expired people' access to BOA. And, on `/admin` page, a small 'is expired' alert in place of the 'become' button.


lyttam:
image

PR #1418, opened 2019-04-08
BOAC-1971: Updates 'updatedAt' when note is edited

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1971


johncrossman:
Up-to-date updated.

PR #1419, opened 2019-04-09
BOAC-2039: Restores edit note form to its former glory

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2039


johncrossman:
O that will be glory!

PR #1421, opened 2019-04-09
BOAC-2043, '/api/notes/attachment/upload' API with tests; put id pkey on note_attachments table

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2043

Introduced `id` pkey on `note_attachments` table cuz the front-end cannot work with composite key `note_id` + `path_to_attachment`.  The pre-deploy script will make the adjustment if run again on BOA db.

### DIFF
--- a/tests/test_api/test_curated_group_controller.py
+++ b/tests/test_api/test_curated_group_controller.py
@@ -50,33 +50,31 @@ def admin_user_session(fake_auth):
 
 
 @pytest.fixture()
-def cohorts_of_asc_advisor():
+def groups_of_asc_advisor():
     advisor = AuthorizedUser.find_by_uid(asc_advisor_uid)
     return CuratedCohort.get_curated_cohorts_by_owner_id(advisor.id)
 
 
 @pytest.fixture()
-def cohorts_of_coe_advisor():
+def groups_of_coe_advisor():
     advisor = AuthorizedUser.find_by_uid(coe_advisor_uid)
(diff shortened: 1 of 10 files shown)


johncrossman:
Why build fails on `TestCuratedCohortDelete` is unclear – tests pass locally. I'll follow up tomorrow.

tests/test_api/test_curated_group_controller.py
```diff
@@ -200,73 +200,83 @@ def test_add_multiple_students_to_curated_cohort(self, asc_advisor, client):
...
-        response = client.get(f'/api/curated_group/{curated_cohort_id}')
-        assert response.status_code == 404
+        assert client.get(f'/api/curated_group/{group_id}').json['name'] == new_name
+
+    def test_delete_group(self, asc_advisor, client):
+        """Delete curated group."""
+        group = _create_group(client, 'Mellow Together')
+        group_id = group['id']
+        assert client.delete(f'/api/curated_group/delete/{group_id}').status_code == 200
+        assert client.get(f'/api/curated_group/{group_id}').status_code == 404
```
johncrossman: My fix to the mysteriously failing test: break the multifaceted `test_create_add_remove_and_delete` into a set of isolated tests, one per endpoint. I could not reproduce – I tried both recreating local test dbs and reinstalling third-party deps. For the record, I would expect such a test, with rapid fire hits + assertions on endpoints, is vulnerable to a race condition.

raydavis:
Plausible enough for me. Thanks, @johncrossman .

PR #1424, opened 2019-04-11
BOAC-1983: Marks updated or newly created note as read by user who did the deed

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1983

### DIFF
--- a/boac/api/notes_controller.py
+++ b/boac/api/notes_controller.py
@@ -37,7 +37,7 @@ from flask_login import current_user, login_required
 @app.route('/api/notes/<note_id>/mark_read', methods=['POST'])
 @login_required
 def mark_read(note_id):
-    if NoteRead.create(current_user.id, note_id):
+    if NoteRead.find_or_create(current_user.id, note_id):
         return tolerant_jsonify({'status': 'created'}, status=201)
     else:
         raise BadRequestError(f'Failed to mark note {note_id} as read by user {current_user.uid}')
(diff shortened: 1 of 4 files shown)


boac/merged/advising_note.py
```diff
@@ -208,7 +208,7 @@ def note_to_compatible_json(note, topics=None, attachments=None):
         'createdAt': _resolve_created_at(note),
         'updatedBy': note.get('updated_by'),
         'updatedAt': _resolve_updated_at(note),
-        'read': False,
+        'read': True if note_read else False,
```
johncrossman: @lyttam Can `True if note_read else False` be replaced with `note_read`?

boac/api/notes_controller.py
```diff
@@ -90,7 +91,10 @@ def update_note():
         subject=subject,
         body=body,
     )
-    note_json = note_to_compatible_json(note.to_api_json())
+    note_read = NoteRead.is_note_read_by_user(current_user.id, note.id)
+    if note and not note_read:
+        note_read = NoteRead.create(current_user.id, note.id)
+    note_json = note_to_compatible_json(note.to_api_json(), note_read=note_read)
```
johncrossman: @lyttam Minor concern: Perhaps `is_note_read_by_user()` is not needed? I.e., no harm in always invoking `NoteRead.create(...)`.  (Also, no null check on `note` is necessary.)

boac/merged/advising_note.py
lyttam: @johncrossman Tried that, but `note_read` is an object that's not JSON serializable. I could probably implement a method on `NoteRead` to make it serializable, if you think it's worth it.

boac/api/notes_controller.py
lyttam: @johncrossman you're right, I'll remove the null check.  But I found that calling `NoteRead.create` twice with the same arguments resulted in a database error (primary key already exists). Should I dispense with `NoteRead.is_note_read_by_user` and add some protection to `NoteRead.create` instead?

boac/merged/advising_note.py
johncrossman: Not necessary, @lyttam This works.

lyttam:
@johncrossman alright, ready for re-review.

PR #1426, opened 2019-04-11
BOAC-1999: Resets edit mode when user navigates away from note edit

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1999


johncrossman:
Looks right 2 me. Thanks @lyttam

PR #1429, opened 2019-04-12
BOAC-2065: Refreshes search index on note update and delete

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2065


johncrossman:
Freshness matters. Thx @lyttam

PR #1430, opened 2019-04-15
BOAC-2042, max attachments per note; BOAC-1937, keep focus in full view modal

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2042


raydavis:
Makes sense from here.

PR #1434, opened 2019-04-18
BOAC-2098/BOAC-2091, filename in s3; change 'Choose file' input to drag-and-drop-with-button

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2091
https://jira.ets.berkeley.edu/jira/browse/BOAC-2098

* The s3 key/path has date info (and so does the db) so we don't need it in filename
* see BOAC-2098 for screenshot of new 'choose file' UX

### DIFF
--- a/src/components/note/NewNoteModal.vue
+++ b/src/components/note/NewNoteModal.vue
@@ -79,23 +79,30 @@
               </span>
             </div>
           </div>
-          <div v-if="newNoteMode === 'fullScreen'" class="mt-2 mr-3 mb-1 ml-3 w-75">
-            <div>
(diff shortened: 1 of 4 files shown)


src/components/note/NewNoteModal.vue
```diff
@@ -315,6 +337,41 @@ export default {
...
+  top: 0;
+  opacity: 0;
+}
+.choose-attachment-file-wrapper:hover,
+.choose-attachment-file-wrapper:focus,
+.choose-attachment-file-wrapper:active
+{
+  color: #333;
+  background-color: #eee;
+}
```
johncrossman: Both new-note and edit-note have styles/logic above (ie, duplicate code). In a separate PR, I will merge into a new component.

PR #1437, opened 2019-04-19
BOAC-2001, update npm packages and certain py dev dependencies

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2001

Most of these are dev deps. A low risk update, methinks.


raydavis:
Guess we'll find out...

PR #1438, opened 2019-04-22
BOAC-2103 Show users their supplied attachment filenames

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2103

... but avoid S3 name collisions internally: BOAC-2091


paulkerschen:
image

PR #1441, opened 2019-04-23
BOAC-2112 Resize note attachment dropzone; no drops outside zone

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2112


raydavis:
Drops that stay dropped. Thanks, @paulkerschen

PR #1448, opened 2019-04-26
BOAC-2122 Not invocation of the function but the function itself

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2122


lyttam:
Important distinction.

PR #1453, opened 2019-04-29
BOAC-2071, refactor away 'curated_cohort' naming; remove cohorts/groups from /api/profile/my

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2071

* Remove `myCohorts` and `myCurated` from `/api/my/profile` because the front-end manages those lists separately. See https://github.com/ets-berkeley-edu/boac/pull/1195 
* This refactor is a nice start to enhancing the API – clarity in naming.

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -36,8 +36,7 @@ from flask_login import current_user, login_required
 
 @app.route('/api/profile/my')
 def my_profile():
-    exclude_cohorts = util.to_bool_or_none(request.args.get('excludeCohorts'))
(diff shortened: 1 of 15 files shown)


boac/api/user_controller.py
```diff
@@ -36,8 +36,7 @@
 
 @app.route('/api/profile/my')
 def my_profile():
-    exclude_cohorts = util.to_bool_or_none(request.args.get('excludeCohorts'))
```
johncrossman: Notice that the front-end always sends `excludeCohorts=true`

paulkerschen:
Thanks for untangling, @johncrossman.

PR #1455, opened 2019-04-30
BOAC-2072, begin front-end work on curated-group-bulk-add

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2072

Front-end route, api call and basic form.


raydavis:
Cheaper by the dozen.

PR #1457, opened 2019-04-30
BOAC-1847 BOAC-1848 Course search by partial subject area or catalog id alone

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1847
https://jira.ets.berkeley.edu/jira/browse/BOAC-1848


johncrossman:
Power to the search.

PR #1458, opened 2019-04-30
BOAC-2072, finishing touches on group creation with bulk-add-SIDs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2072

Using bulk-add with an existing group is next.

### DIFF
--- a/src/views/CuratedGroupBulkAdd.vue
+++ b/src/views/CuratedGroupBulkAdd.vue
@@ -7,19 +7,21 @@
     <div v-if="!curatedGroup">
       Create a curated group of students by adding their Student Identification (SID) numbers below.
     </div>
+    <h2 class="page-section-header-sub mt-3">Add SID numbers</h2>
+    <div>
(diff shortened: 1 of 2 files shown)


src/views/CuratedGroupBulkAdd.vue
```diff
@@ -7,28 +7,30 @@
     <div v-if="!curatedGroup">
       Create a curated group of students by adding their Student Identification (SID) numbers below.
     </div>
+    <h2 class="page-section-header-sub mt-3">Add SID numbers</h2>
+    <div>
+      Type or paste a list of SID numbers. Example: 3033223869, 3033112579
```
paulkerschen: @johncrossman Are these example SIDs for test users or otherwise guaranteed safe to use?

johncrossman: A fair question. I've amended with nine-9s SIDs.

paulkerschen:
Thanks @johncrossman!

PR #1463, opened 2019-05-01
BOAC-2127, accessibility enhancements for curated group: bulk-add SIDs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2127


paulkerschen:
Explicit!

PR #1467, opened 2019-05-01
BOAC-2130, omit the 'unavailable' SIDs when rendering curated-group view

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2130


paulkerschen:
Nice that you were able to leverage existing code to support this weird requirement.

PR #1469, opened 2019-05-01
BOAC-2049: Adds id to note_topics table

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2049

This is to lay the necessary groundwork for my next PR for BOAC-2051.


paulkerschen:
An id column never hurt no one.

PR #1470, opened 2019-05-01
BOAC-2128 Clear errors on note expand/collapse

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2128


lyttam:
Short and to the point.

PR #1471, opened 2019-05-01
BOAC-2132, curated group student-count should not include the group's 'unavailable' SIDs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2132

Includes more concise wording (eg, toss out "Bulk")

### DIFF
--- a/boac/api/util.py
+++ b/boac/api/util.py
@@ -260,15 +260,14 @@ def get_my_curated_groups():
     curated_groups = []
     user_id = current_user.id
     for curated_group in CuratedGroup.get_curated_groups_by_owner_id(user_id):
-        _curated_group_api_json(curated_group)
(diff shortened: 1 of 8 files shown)


boac/api/util.py
```diff
@@ -261,14 +261,14 @@ def get_my_curated_groups():
     user_id = current_user.id
     for curated_group in CuratedGroup.get_curated_groups_by_owner_id(user_id):
         _curated_group_api_json(curated_group)
```
paulkerschen: @johncrossman Not that this line is part of your PR, but I wonder how it got in here? It doesn't seem to be generating anything that subsequent lines use; looks like it's just building a data structure and throwing it away.

johncrossman: You're right and I think it should go. Amending.

PR #1472, opened 2019-05-02
BOAC-2067 Increase message lifespan on curated group edits

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2067

Subjectively, and depending on what else is happening in the browser, 1000ms feels a bit brief.


johncrossman:
An extra 1000ms gives us time to take a breath

PR #1474, opened 2019-05-02
BOAC-2133, empty curated group UX: Keep it simple st__id

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2133

See screenshots in the Jira. I did not touch 'context' prop in `SearchForm` – it might prove to be useful.


paulkerschen:
Works for me!

PR #1475, opened 2019-05-03
BOAC-2146, curated-group bulk-add SIDs > all student profile data

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2146

See Jira description.

### DIFF
--- a/boac/api/curated_group_controller.py
+++ b/boac/api/curated_group_controller.py
@@ -55,18 +55,6 @@ def create_curated_group():
     return tolerant_jsonify(curated_group.to_api_json())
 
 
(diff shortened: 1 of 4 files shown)


boac/api/curated_group_controller.py
```diff
@@ -55,18 +55,6 @@ def create_curated_group():
     return tolerant_jsonify(curated_group.to_api_json())
 
 
-@app.route('/api/curated_group/<curated_group_id>/add_student/<sid>')
```
johncrossman: It turns out that this API path is unused on the front-end.

PR #1476, opened 2019-05-03
NS-482 Begin adding optional support for L&S advisors

raydavis:
https://jira.ets.berkeley.edu/jira/browse/NS-482

The PR also simplifies the SQL scripts which add departments and advisors, because otherwise the release instructions I'd have to add would be too annoying.


raydavis:
Missed a `git push`; Travis should be happier now.

paulkerschen:
Progress is replacing opaque numeric ids with slightly less opaque five-letter codes. Thanks @raydavis!

PR #1479, opened 2019-05-06
BOAC-2079, tool_settings table and protected get-settings API

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2079
https://jira.ets.berkeley.edu/jira/browse/BOAC-2080

Server-side (incl. db) work to support get/set of tool_settings. On the front-end, `toolSettings` are folded in with the 'config' store. In next PR, the `Context` mixin will make serviceAlert available to templates and Admin view will post to '/api/tool_setting/upsert'.


johncrossman:
I will reopen this PR after a couple amendments are made.

PR #1480, opened 2019-05-06
BOAC-2079, tool_settings table and protected get/update service-announcement API

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2079
https://jira.ets.berkeley.edu/jira/browse/BOAC-2080

Server-side (incl. db) work to support get/set of service-announcement. On the front-end, toolSettings are folded in with the 'config' store. This PR includes front-end api and store logic. The textarea/checkbox on Admin page works – the styling on the announcement div and Admin component is not done.

### DIFF
--- a/src/components/admin/EditServiceAnnouncement.vue
+++ b/src/components/admin/EditServiceAnnouncement.vue
@@ -1,0 +1,62 @@
+<template>
+  <div v-if="user.isAdmin && announcement !== undefined">
+    <h2 class="page-section-header-sub">Service Alert</h2>
+    <div class="p-2">
+      <span
+        v-if="isSaving"
+        role="alert"
+        aria-live="passive"
+        class="sr-only">{{ isPublished ? 'Publishing' : 'Unpublishing' }} the service announcement</span>
+      <b-form-checkbox
+        id="publish-service-announcement"
+        v-model="isPublished"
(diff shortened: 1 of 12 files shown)


src/components/admin/EditServiceAnnouncement.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+  <div v-if="announcement !== undefined">
+    <h2 class="page-section-header-sub">Service Alert</h2>
+    <div class="p-2">
+      <span
+        v-if="isSaving"
+        role="alert"
+        aria-live="passive"
+        class="sr-only">{{ isPublished ? 'Publishing' : 'Unpublishing' }} the service announcement</span>
```
lyttam: @johncrossman Should this be the other way around? If `isSaving = true` then `isPublished` reflects the old value right?

johncrossman: @lyttam It's confusing but this is right. This `sr-alert` only happens if `isSaving` is true. So, during save, this logic makes sense. I retested and noticed that UserMetadata mixin is needed – that amendment is in.

lyttam: @johncrossman thanks for confirming!

PR #1481, opened 2019-05-06
BOAC-2152, self.isSaving is useless if 'self' is undefined

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2152

I assume that `self` was missed during a refactor.


raydavis:
Salvo ergo sum.

PR #1483, opened 2019-05-06
BOAC-2051: Implements add/remove topic feature for note editing

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2051

Will submit a subsequent PR for adding note topics during note creation.


johncrossman:
Thank you @lyttam  Looks right.

PR #1484, opened 2019-05-07
BOAC-2083, validation on service-alert input and aria on announcement text

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2081


paulkerschen:
I feel validated.

PR #1486, opened 2019-05-07
BOAC-2158, UX enhancements to update/publish service alert on Admin page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2158

* publish and update-text are now separate operations
* UX tweaked to reflect the point above


paulkerschen:
It's a stealth CMS, isn't it.

PR #1487, opened 2019-05-07
BOAC-2162, CuratedGroupBulkAdd component must know if create/save op was cancelled

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2162


paulkerschen:
Think of it this way: every time you're shopping, you're saving.

PR #1489, opened 2019-05-07
BOAC-2051: Implements add/remove topic feature for note creation

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2051

Also fixes some apparent regressions that should have been included in the previous PR.

### DIFF
--- a/boac/api/notes_controller.py
+++ b/boac/api/notes_controller.py
@@ -201,7 +201,7 @@ def _get_name(user):
 
 def _get_topics(params):
     topics = params.get('topics', ())
-    return topics if isinstance(topics, list) else str(topics).split(',')
+    return topics if isinstance(topics, list) else list(filter(None, str(topics).split(',')))
 
 
 def _get_attachments(request_files, tolerate_none=False):
--- a/src/api/notes.ts
+++ b/src/api/notes.ts
@@ -21,7 +21,7 @@ export function createNote(
     sid: sid.toString(),
     subject: subject,
     body: body,
-    topics: topics,
+    topics: topics
   };
   _.each(attachments || [], (attachment, index) => data[`attachment[${index}]`] = attachment);
   return apiUtils.postMultipartFormData('/api/notes/create', data);
--- a/src/api/notes.ts
+++ b/src/api/notes.ts
@@ -10,11 +10,18 @@ export function markRead(noteId) {
     .then(response => response.data, () => null);
 }
 
-export function createNote(sid: any, subject: string, body: string, attachments: any[]) {
+export function createNote(
+    sid: any,
+    subject: string,
+    body: string,
+    topics: string[],
+    attachments: any[]
+) {
   const data = {
     sid: sid.toString(),
     subject: subject,
-    body: body
+    body: body,
+    topics: topics,
   };
   _.each(attachments || [], (attachment, index) => data[`attachment[${index}]`] = attachment);
   return apiUtils.postMultipartFormData('/api/notes/create', data);
--- a/src/components/note/NewNoteModal.vue
+++ b/src/components/note/NewNoteModal.vue
@@ -80,6 +80,53 @@
             </div>
           </div>
           <div v-if="newNoteMode === 'fullScreen'" class="mt-2 mr-3 mb-1 ml-3">
+            <div>
+              <label id="add-note-topic-label" class="font-weight-bold mt-2" for="add-note-topic">
+                Topic Categories
+              </label>
+            </div>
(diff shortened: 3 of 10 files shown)


boac/api/notes_controller.py
```diff
@@ -201,7 +201,7 @@ def _get_name(user):
 
 def _get_topics(params):
     topics = params.get('topics', ())
-    return topics if isinstance(topics, list) else str(topics).split(',')
+    return topics if isinstance(topics, list) else list(filter(None, str(topics).split(',')))
```
johncrossman: @lyttam Why doesn't `split()` return a list as you need?

tests/test_api/test_notes_controller.py
```diff
@@ -120,6 +120,9 @@ def test_create_note_with_topics(self, app, client, fake_auth):
             topics=['collaborative synergies', 'integrated architectures', 'vertical solutions'],
         )
         assert len(note.get('topics')) == 3
+        assert note.get('topics')[0] == 'COLLABORATIVE SYNERGIES'
+        assert note.get('topics')[1] == 'INTEGRATED ARCHITECTURES'
+        assert note.get('topics')[2] == 'VERTICAL SOLUTIONS'
```
johncrossman: @lyttam Will we be displaying topics in all caps on front-end?

src/api/notes.ts
```diff
@@ -10,11 +10,18 @@ export function markRead(noteId) {
...
+    body: string,
+    topics: string[],
+    attachments: any[]
+) {
   const data = {
     sid: sid.toString(),
     subject: subject,
-    body: body
+    body: body,
+    topics: topics,
```
johncrossman: Minor: trailing comma above

tests/test_api/test_notes_controller.py
lyttam: Yes, according to the mockup. But I also made sure to store them in the database in all caps to facilitate future autocomplete functionality. I think I'd rather have them be in title case so I might change that later, but for now all caps was the easiest thing to do.

src/components/note/NewNoteModal.vue
```diff
@@ -79,6 +79,53 @@
...
+                        class="px-0 pt-1"
+                        @click.prevent="removeTopic(topic)">
+                        <i class="fas fa-times-circle has-error pl-2"></i>
+                      </b-btn>
+                    </span>
+                  </li>
+                </ul>
+              </div>
+            </b-container>
+          </div>
```
johncrossman: @lyttam the `b-container` above appears to be identical to what's in EditAdvisingNote. Would a shared component make sense?

boac/api/notes_controller.py
lyttam: It does, but if a blank topic is submitted the list will contain an empty string. The purpose of `filter` is to remove the empty strings. Is there a better way to do this?

tests/test_api/test_notes_controller.py
johncrossman: Thanks @lyttam  The mockups in BOAC-1924 have title case. Has there been a revision of the design? IMO, all caps in the UI would be a bit jarring.

boac/api/notes_controller.py
johncrossman: Ah, thank you. Makes sense.

johncrossman:
I posted a few questions. Nice work @lyttam

src/api/notes.ts
lyttam: Thanks; will amend.

tests/test_api/test_notes_controller.py
lyttam: @johncrossman the autocomplete suggestions are in title case, but the added topics are all caps. I'm going to work on the autocomplete Jira next and I think I'll wind up changing it so topics are stored in title case, but added topics are still displayed in all caps using CSS. It's just a matter of finding the right library to properly convert articles (of, and, or, but) and acronyms (PNP) to title case.

image

src/components/note/NewNoteModal.vue
lyttam: @johncrossman yes, I thought about it. I'll make it shared.

tests/test_api/test_notes_controller.py
johncrossman: Makes sense @lyttam  Ping me when amendments are in and I'll merge.

lyttam:
Ready for re-review, @johncrossman.

johncrossman:
Thanks @lyttam!

PR #1491, opened 2019-05-08
BOAC-2030, minor fix to curated group select on student page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2030


src/components/student/profile/StudentGroupSelector.vue
```diff
@@ -194,12 +192,6 @@ export default {
 };
 </script>
 
-<style>
-.cohort-checkbox-name {
-  padding: 10px;
-}
-</style>
-
```
johncrossman: Moved the above to `boac-global.css`

PR #1492, opened 2019-05-08
BOAC-2171 Cohort membership constrained by creator

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2171


johncrossman:
Cohorts respect the Creator.

PR #1493, opened 2019-05-08
BOAC-2050: Introduces suggestions for note topics

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2050


johncrossman:
Thanks @lyttam  Let's give topics a spin.

PR #1496, opened 2019-05-09
BOAC-2148, swap in BOA email address; responsive footer

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2148


paulkerschen:
When I was younger, so much younger than today, I never needed any BOA help in any way.

PR #1497, opened 2019-05-09
BOAC-2179, if user is not auth then /api/service_announcement returns None

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2179


paulkerschen:
Sensible.

PR #1505, opened 2019-05-13
BOAC-2186, b-container:fluid in StandardLayout for responsive template

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2186

In a narrow viewport (eg, mobile) the header, sidebar and main-content elements will fall into a single column. This PR also trims down the amount of CSS. 

Review with `?w=1`


paulkerschen:
Less CSS, more lean protein for responsive power.

PR #1508, opened 2019-05-13
BOAC-2191 Remove HTML tags with a parser like a grown-up

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2191


johncrossman:
`tag_stripper` is the variable name of the month

PR #1510, opened 2019-05-13
BOAC-2193: Autocompletes topic with match from suggested topics

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2193

I tried to make this as unobtrusive as possible. It would be great to have others test it and report their annoyance levels.

### DIFF
--- a/src/components/note/AdvisingNoteTopics.vue
+++ b/src/components/note/AdvisingNoteTopics.vue
@@ -52,8 +52,11 @@
 </template>
 
 <script>
+import Util from '@/mixins/Util';
+
 export default {
   name: 'AdvisingNoteTopics',
+  mixins: [Util],
   props: {
     functionAdd: {
       type: Function,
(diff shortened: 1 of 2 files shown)


src/components/note/AdvisingNoteTopics.vue
```diff
@@ -75,12 +78,36 @@ export default {
...
+  },
   methods: {
     addTopic() {
       if (this.topic && this.topic.trim()) {
         this.functionAdd(this.topic);
       }
       this.topic = undefined;
+    },
+    normalizeString(str) {
+      return str.toUpperCase().replace(/[^A-Za]/g, '');
```
paulkerschen: @lyttam Looks like this regex would preserve lowercase `a` along with uppercase letters - typo or quirk of the topics?

lyttam: @paulkerschen thank you, that is a typo. Will amend.

lyttam: @paulkerschen all fixed.

paulkerschen:
Let's give it a spin!

PR #1515, opened 2019-05-14
[QA] BOAC-2202, if text is truncated then Safari has full-text tooltip (a problem for demo-mode)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2202

See suggested solutions in https://zzz.buzz/2017/07/31/prevent-tooltip-over-truncated-text-in-safari

I chose "the most obvious solution ... add a title attribute"


lyttam:
@johncrossman this PR has more changes than the corresponding master PR, is that intentional?

johncrossman:
@lyttam You're right, the simpler fix is possible here. I'll amend.

lyttam:
👍

PR #1519, opened 2019-05-14
[QA] BOAC-2203: Prevents note/topics call when note editing is turned off

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2203

QA PR for #1518


lyttam:
Oops, wrong branch. Will submit a new PR.

PR #1521, opened 2019-05-14
NS-532 Update smart_open to 1.8.3

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-532

I'm unable to reproduce reported 1.8.3 breakages locally, but the dev environment is made to be broken, so let's try it there.


johncrossman:
Break, baby, break!

PR #1523, opened 2019-05-14
BOAC-855, paginated curated groups

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-855

I had to remove the SQLAlchemy join on students and load students (or sids) on demand, with offset and limit options. Remember that curated groups are special in that we filter SIDs of the group based on user's `dept_code`.

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -31,6 +31,7 @@ from boac.externals import data_loch
 from boac.lib import berkeley
 from boac.models import json_cache
 from boac.models.alert import Alert
+from boac.models.curated_group import CuratedGroupStudent
 from boac.models.job_progress import JobProgress
 from boac.models.json_cache import JsonCache
 from flask import current_app as app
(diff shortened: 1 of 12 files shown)


boac/api/cache_utils.py
```diff
@@ -236,10 +237,10 @@ def update_curated_group_lists():
...
     from boac.models.curated_group import CuratedGroup
     for curated_group in CuratedGroup.query.all():
-        member_sids = [s.sid for s in curated_group.students]
-        available_students = [s['sid'] for s in data_loch.get_student_profiles(member_sids)]
-        if len(member_sids) > len(available_students):
-            unavailable_sids = set(member_sids) - set(available_students)
+        all_sids = CuratedGroupStudent.get_sids(curated_group.id)
+        available_students = [s['sid'] for s in data_loch.get_student_profiles(all_sids)]
+        if len(all_sids) > len(available_students):
+            unavailable_sids = set(all_sids) - set(available_students)
```
paulkerschen: @johncrossman Is this controller-level filtering still necessary? If I'm reading the PR right, seems like it's now being done at the model level down in `query_students` so that the offset and limit come out correctly?

paulkerschen: @johncrossman Never mind; this isn't the controller at all, it's the cache. So many moving parts.

paulkerschen:
Well, let's paginate 'em.

PR #1524, opened 2019-05-15
BOAC-2199: Restores topics after cancel note edit

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2199


paulkerschen:
Always Be Cloning.

PR #1527, opened 2019-05-15
BOAC-2209 Switch profile and term queries to RDS

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2209

Testing out what effect this has on performance in dev.


johncrossman:
We're in PDOg Labs

PR #1530, opened 2019-05-15
BOAC-2208, SQL scripts: delete_authorized_user and add_admin_user;  moved such scripts to 'scripts/db/tools'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2208


paulkerschen:
Remember kids, always check with a grown-up before using the db/tools!

PR #1531, opened 2019-05-15
BOAC-2209 Refactor search for (perhaps) performance improvements

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2209


johncrossman:
"perhaps" is good enough for me!

PR #1536, opened 2019-05-16
BOAC-1898, recognize/tag URLs in body text of new/updated notes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1898
https://jira.ets.berkeley.edu/jira/browse/BOAC-1897

`linkify` does its job but we don't want it parsing properly tagged URLs, thus the custom pre-processing (see `placeholder`).

### DIFF
--- a/boac/lib/util.py
+++ b/boac/lib/util.py
@@ -23,11 +23,13 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-
 from datetime import datetime
 import inspect
+import re
 import string
+import time
 
+from autolink import linkify
 from flask import current_app as app
 import pytz
 from titlecase import titlecase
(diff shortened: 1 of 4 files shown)


boac/lib/util.py
```diff
@@ -64,6 +66,20 @@ def localize_datetime(dt):
     return dt.astimezone(pytz.timezone(app.config['TIMEZONE']))
 
 
+def process_input_from_rich_text_editor(rich_text):
+    parsed = rich_text.strip()
+    exclude_from_parse = {}
+    now = time.time()
+    for index, match in enumerate(re.findall('<a .*>.*</a>', parsed)):
```
paulkerschen: @johncrossman What happens if there's more than one link tag? In at least some implementations the `.*` operator is greedy, meaning it will match as much as possible in one go; so if you had text like

`<p>They were <a href="http://www.guzzle.com">www.guzzle.com</a> at <b>https://marsh.mallows.com</b> and <a href="http://www.foxnews.com">FOX news</a></p>`

it might miss the URL in between the two links.

If it does in fact behave that way, you might be able to tame the greediness by changing it to something like `<a [^>]*>.*?</a>`.

johncrossman: You are right. And... the suggested regex does the trick. Amending.

PR #1541, opened 2019-05-17
BOAC-2052: Makes topics feature more screen reader friendly

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2052

### DIFF
--- a/src/components/note/AdvisingNoteTopics.vue
+++ b/src/components/note/AdvisingNoteTopics.vue
@@ -13,52 +13,69 @@
               id="add-note-topic"
               v-model="topic"
               aria-labelledby="add-note-topic-label"
+              aria-describedby="add-note-topic-instructions"
+              aria-owns="add-topic-input-list"
               type="text"
               maxlength="255"
               list="add-topic-input-list"
               @keydown.enter="addTopic">
             </b-form-input>
-            <b-form-datalist id="add-topic-input-list" :options="suggestedTopics"></b-form-datalist>
+            <span id="add-note-topic-instructions" class="sr-only">When autocomplete results are available, use up and down arrows to review and enter to select.</span>
+            <b-form-datalist id="add-topic-input-list" :options="suggestedTopics" role="listbox"></b-form-datalist>
             <b-input-group-append>
               <b-button
                 id="add-topic-button"
                 slot="append"
                 class="btn-add-topic"
+                aria-controls="note-topics-list"
                 @click="addTopic">
-                <i class="fas fa-plus pr-1"></i>Add
+                <i class="fas fa-plus pr-1"></i>Add<span class="sr-only"> Topic</span>
               </b-button>
             </b-input-group-append>
           </b-input-group>
         </b-col>
       </b-form-row>
       <div>
-        <ul class="pill-list pl-0">
+        <ul
+          id="note-topics-list"
+          class="pill-list pl-0"
+          aria-labelledby="note-topics-label">
           <li
-            v-for="(topic, index) in topics"
+            v-for="(addedTopic, index) in topics"
+            id="`note-topic-${index}`"
             :key="index">
             <span class="pill pill-attachment text-uppercase text-nowrap">
-              {{ topic }}
+              {{ addedTopic }}
               <b-btn
-                :id="`remove-topic-${index}`"
+                :id="`remove-note-topic-${index}`"
                 variant="link"
                 class="px-0 pt-1"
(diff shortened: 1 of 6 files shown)


src/components/note/AdvisingNoteTopics.vue
```diff
@@ -13,52 +13,68 @@
...
+          id="note-topics-list"
+          class="pill-list pl-0"
+          aria-labelledby="note-topics-label">
           <li
             v-for="(topic, index) in topics"
+            id="`note-topic-${index}`"
             :key="index">
             <span class="pill pill-attachment text-uppercase text-nowrap">
-              {{ topic }}
+              <span tabindex="0">{{ topic }}</span>
```
johncrossman: @lyttam Is `tabindex` necessary here? Is it for screenreader voicing?

src/components/note/AdvisingNoteTopics.vue
```diff
@@ -92,14 +108,25 @@ export default {
   },
   methods: {
     addTopic() {
+      this.alertScreenReader();
```
johncrossman: @lyttam Is this empty `alertScreenReader()` call designed to silence screenreader?

src/components/note/AdvisingNoteTopics.vue
```diff
@@ -13,52 +13,68 @@
...
+            <span id="add-note-topic-instructions" class="sr-only">When autocomplete results are available, use up and down arrows to review and enter to select.</span>
             <b-form-datalist id="add-topic-input-list" :options="suggestedTopics"></b-form-datalist>
             <b-input-group-append>
               <b-button
                 id="add-topic-button"
                 slot="append"
                 class="btn-add-topic"
                 @click="addTopic">
-                <i class="fas fa-plus pr-1"></i>Add
+                <i class="fas fa-plus pr-1"></i>Add <span class="sr-only">Topic</span>
```
johncrossman: Minor: the space should live inside the sr-only span

src/views/Student.vue
```diff
@@ -2,6 +2,7 @@
   <div>
     <Spinner />
     <div v-if="!loading">
+      <div class="sr-only" aria-live="polite">{{ screenReaderAlert }}</div>
```
johncrossman: @lyttam I don't think this is necessary. See `srAlert` in `StandardLayout.vue` and how the `Context` mixin is wired.

src/views/Student.vue
```diff
@@ -62,6 +63,7 @@ export default {
   data: () => ({
     cancelTheCancel: undefined,
     cancelConfirmed: undefined,
+    screenReaderAlert: undefined,
```
johncrossman: Same: not necessary.

src/components/note/AdvisingNoteTopics.vue
lyttam: @johncrossman It was necessary to allow focus on the list item so the screen reader would voice the added topic, but now that the SR text for the remove button includes the topic, I suppose it's not needed. I'll remove it.

src/components/note/AdvisingNoteTopics.vue
lyttam: @johncrossman No, it's to clear the alert text before adding new text. I found this was needed in order to get NVDA to voice the entire alert in Firefox. Otherwise, it only voices the part of the alert that changed. For example if I added a topic it would say "topic Academic Progress added", but if I removed the topic it would only say "removed".

src/components/note/AdvisingNoteTopics.vue
lyttam: I'll amend this.

src/views/Student.vue
lyttam: @johncrossman You're right; I'll amend.

src/components/note/AdvisingNoteTopics.vue
johncrossman: Thanks @lyttam  - that makes sense. I'm concerned that our `alertScreenReader('Blah, blah, blah')` calls must be proceeded by `alertScreenReader()`.  Can we add something at the mixin level to trick NVDA into reading the entire `srAlert`?

lyttam: Yes good idea.

lyttam:
Thanks for the great feedback @johncrossman. I addressed all the issues so this should be ready to merge.

johncrossman:
Thanks @lyttam  Nice work!

PR #1542, opened 2019-05-17
NS-528, remove sudo from buildspec.yml

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/NS-528

`sudo` is not available on CodeBuild's transient Docker image – maybe because `root` is default user in new 0.2 spec.  So `sudo` is out. (We tested this change, from my working branch, on boac-dev pipeline and deploy succeeded.)

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -15,8 +15,8 @@ phases:
 
   build:
     commands:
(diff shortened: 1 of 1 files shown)


paulkerschen:
`sudo` is old hat!

PR #1543, opened 2019-05-17
BOAC-2218 Incorporate JSON caching into bulk advisor lookup

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2218


johncrossman:
Fetch, cache and repeat.

PR #1550, opened 2019-05-20
[BOAC-2235] Changing AWS auth scheme to use STS tokens and app_roles.

sandeepmjay:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2235

1) Boto authentication is switched to create session using STS credentials. This limits credential exposure. The commit changes the auth scheme for BOA while interacting with AWS services like S3. A Short term credential(includes temp key, secret, token) is created for every interaction which is alive for 15 minutes duration using STS service. Permissions are given to the developer group to assume roles in the context of the app and environment they will be running which limits cross stack access, limiting the blast radius.  

2) Elastic Beanstalk changes to represent the use of IAM instance roles. IAM instance profile on EC2 servers will have the app_role attached to it which provides the same permissions for instances to assume roles and interact with AWS services. 

3) S3 Encryption mode will be part of configs going forward so that S3 encryption schemes can be changed as necessary by the Dev Sec Ops team while rotating encryption keys and policies without functionality disruption.

### DIFF
--- a/.ebextensions/00_ami.config
+++ b/.ebextensions/00_ami.config
@@ -78,4 +78,4 @@ Resources:
             "Fn::GetOptionSetting":
               Namespace: "aws:autoscaling:launchconfiguration"
               OptionName: "IamInstanceProfile"
-              DefaultValue: "aws-elasticbeanstalk-ec2-role"
+              DefaultValue: "RTL-app-boa-dev-role"
--- a/.ebextensions/12_set_default_region.config
+++ b/.ebextensions/12_set_default_region.config
@@ -1,0 +1,6 @@
+#
+# Set AWS Default region for sts token to be created in required region.
+#
+container_commands:
+  01_set_default_region:
+    command: 'export AWS_DEFAULT_REGION=us-west-2'
--- a/boac/externals/s3.py
+++ b/boac/externals/s3.py
@@ -31,18 +31,15 @@ import smart_open
 """Client code to run file operations against S3."""
 
 
-def build_s3_url(bucket, key, credentials=True):
-    if credentials:
-        credentials = ':'.join([app.config['AWS_ACCESS_KEY_ID'], app.config['AWS_SECRET_ACCESS_KEY']])
-        return f's3://{credentials}@{bucket}/{key}'
-    else:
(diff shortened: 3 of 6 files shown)


.ebextensions/11_install_pandas.config
```diff
@@ -1,6 +1,9 @@
 #
 # Install pandas once its dependencies, specified by requirements.txt, are in place.
+# Set AWS Default region for sts token to be created in required region.
 #
 container_commands:
   01_install_pandas:
     command: '/opt/python/run/venv/bin/pip install pandas==0.23.3'
+  02_set_default_region:
+    command: 'export AWS_DEFAULT_REGION=us-west-2'
```
paulkerschen: @sandeepmjay Does this have any relation to the pandas installation in particular, or does it perhaps belong in a different container command?

.ebextensions/00_ami.config
```diff
@@ -78,4 +78,4 @@ Resources:
             "Fn::GetOptionSetting":
               Namespace: "aws:autoscaling:launchconfiguration"
               OptionName: "IamInstanceProfile"
-              DefaultValue: "aws-elasticbeanstalk-ec2-role"
+              DefaultValue: "RTL-app-boa-dev-role"
```
sandeepmjay: The Value will be parameterized in the next iteration.

.ebextensions/11_install_pandas.config
sandeepmjay: Not totally related to pandas. But, it is a container command that needs to explicitly set the region to prevent unnecessary redirects while interacting with S3 and STS.

.ebextensions/00_ami.config
paulkerschen: Right; so the new command probably shouldn't be living in a file called `.ebextensions/11_install_pandas.config`. Could we rename that file or add a new .ebextensions file for the new command?

paulkerschen:
Aside from the one file placement comment, looks good to me!

.ebextensions/00_ami.config
sandeepmjay: Done. added a separate file for the region.

PR #1552, opened 2019-05-20
BOAC-2180, basic HTML support in BOA service announcements

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2180

The corresponding Teena change is queued up.


paulkerschen:
Now, when there's nothing in particular to announce, the default announcement could be "advising note of the day."

PR #1554, opened 2019-05-20
BOAC-2224 Don't let that old notes checkbox disappear

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2224

As far as I can tell, having the text button wink in and out of existence was confusing Bootstrap's flex implementation. Simply toggling CSS visibility keeps things stable.


johncrossman:
Don't let your babies grow up to be checkboxes.

PR #1558, opened 2019-05-21
BOAC-2233, canned topic is unavailable if you've already added it

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2234
https://jira.ets.berkeley.edu/jira/browse/BOAC-2233

Includes fix to disable 'Add' button when input is blank.


paulkerschen:
Can it!

PR #1562, opened 2019-05-21
BOAC-2231 Constrain topics to 50 characters

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2231

With arbitrary truncation of what's already in the db, since we haven't let real users near it.


lyttam:
Seems sensible.

PR #1566, opened 2019-05-22
BOAC-2228 Boring old dropdown select for advising note topics

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2228

### DIFF
--- a/src/components/note/AdvisingNoteTopics.vue
+++ b/src/components/note/AdvisingNoteTopics.vue
@@ -9,19 +9,13 @@
       <b-form-row class="pb-1">
         <b-col cols="9">
           <b-input-group>
-            <b-form-input
-              id="add-note-topic"
+            <span id="add-note-topic-instructions" class="sr-only">Use up and down arrows to review categories and enter to select.</span>
+            <b-form-select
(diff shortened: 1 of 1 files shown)


src/components/note/AdvisingNoteTopics.vue
```diff
@@ -111,23 +107,13 @@ export default {
...
-    }
-  },
   created: function() {
-    this.topicOptions = [];
     this.each(this.suggestedTopics, suggestedTopic => {
       this.topicOptions.push({
         text: suggestedTopic,
-        value: suggestedTopic,
-        disabled: this.includes(this.topics, suggestedTopic)
+        value: suggestedTopic
```
johncrossman: @paulkerschen If we're editing an existing note, with one or more topics, what is the mechanism for disabling the already-selected topics?

johncrossman:
My only other comment...

image

src/components/note/AdvisingNoteTopics.vue
paulkerschen: Thanks, @johncrossman - that line was removed in error. Amending.

PR #1567, opened 2019-05-22
BOAC-1855 Filtered note search by topic

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1855


johncrossman:
How topical.

PR #1568, opened 2019-05-23
BOAC-2243 Note search: translate to legacy topic formats as required

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2243


johncrossman:
God Scéaw!

PR #1569, opened 2019-05-24
BOAC-2245, update studentCount properly when student is removed from curated group

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2245


paulkerschen:
That student never belonged in there anyway.

PR #1570, opened 2019-05-24
BOAC-2252, true lazy-load of myCohorts and myCuratedGroups

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2252


paulkerschen:
It’s not lazy till it’s truly lazy.

PR #1571, opened 2019-05-24
BOAC-2247 Keep "author" attached to individual notes

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2247


paulkerschen:
What an odd little byway in that code path.

PR #1577, opened 2019-05-30
BOAC-1995 Correct local time display for note timestamps

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1995


johncrossman:
Looks right – thanks @paulkerschen Where's our Travis build?! I'll investigate.

PR #1584, opened 2019-05-31
[QA] BOAC-2294: Removes feature flag restriction from topics API

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2294

### DIFF
--- a/src/components/sidebar/SearchForm.vue
+++ b/src/components/sidebar/SearchForm.vue
@@ -180,16 +180,21 @@ export default {
     }
   },
   created() {
-    getTopics().then(data => {
-      this.each(data, topic => {
-        this.topicOptions.push({
-          text: topic,
-          value: topic
-        })
-      });
-    });
+    if (this.featureFlagEditNotes) {
+      this.getNoteTopics();
+    }
   },
   methods: {
+    getNoteTopics() {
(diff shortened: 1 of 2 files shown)


src/components/sidebar/SearchForm.vue
```diff
@@ -180,16 +180,21 @@ export default {
...
-    getTopics().then(data => {
-      this.each(data, topic => {
-        this.topicOptions.push({
-          text: topic,
-          value: topic
-        })
-      });
-    });
+    if (this.featureFlagEditNotes) {
+      this.getNoteTopics();
```
johncrossman: @lyttam - In `data()`, we init `topicOptions` with a non-empty array. Will the `b-form-select` render with one option?

lyttam:
@johncrossman Actually, what we probably want is for the topics list to display in the search form even when note editing is turned off, right? So users can still search by topic.

johncrossman:
@lyttam That's a good point. I would test locally by removing the `@feature_flag_edit_notes` on /api/notes/topics, set feature flag to false and verify search works fine. Thanks!

johncrossman:
This looks good @lyttam  Ready for merge?

lyttam:
@johncrossman yes, thanks

PR #1585, opened 2019-06-03
BOAC-2300 Fix loading-aware conditional in cohort/group sort

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2300

By an accident of history, `this.loading` is a data value, while `this.loaded()` is a setter function returning nothing.


johncrossman:
I think this might also fix https://jira.ets.berkeley.edu/jira/browse/BOAC-2302   I'll verify a bit more.

PR #1587, opened 2019-06-03
BOAC-2302, 'continueEdit-if-back-button' logic in /cohort is fooled by clicking around sidebar

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2302

The 'continueExistingSession' logic allows users to click to (1) view a student and then (2) use back-button to return to the working state of cohort view. For example, page number and filter edits are preserved.

In the 'continueExistingSession' logic, we're getting false-positives from `$routerHistory.hasForward()` when user is clicking around links in sidebar. This PR avoids such false-positives by only supporting back button when user goes /cohort to /student and then back-button to /cohort. IMO, this is an intuitive limit on back-button support.

### DIFF
--- a/src/mixins/Util.vue
+++ b/src/mixins/Util.vue
@@ -81,6 +81,7 @@ export default {
       }
     },
     split: _.split,
+    startsWith: _.startsWith,
     stripHtmlAndTrim: html => {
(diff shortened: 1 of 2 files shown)


paulkerschen:
I'm persuaded.

PR #1596, opened 2019-06-04
BOAC-2308, secondary, tertiary, etc sort is always ASC; rogue icons begone

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2246
https://jira.ets.berkeley.edu/jira/browse/BOAC-2308


raydavis:
With any luck, this will hit a couple of other JIRA tickets, too.

PR #1603, opened 2019-06-04
[QA] BOAC-2165: Preserves cohort filters when returning via back button

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2165

QA version of #1602


johncrossman:
Nice. Thx @lyttam

PR #1606, opened 2019-06-05
BOAC-1879: Expands scope of note search

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1879


paulkerschen:
Thanks for all the extra tests in the search controller, @lyttam - it needed those.

PR #1607, opened 2019-06-05
[QA] BOAC-2319, border-bottom on each student row of class page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2319


lyttam:
Stay clazzy.

PR #1609, opened 2019-06-05
BOAC-2308, nil in column-type number is treated as zero during sort

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2308
https://jira.ets.berkeley.edu/jira/browse/BOAC-2317

### DIFF
--- a/src/assets/styles/bootstrap-overrides.css
+++ b/src/assets/styles/bootstrap-overrides.css
@@ -32,14 +32,15 @@ legend {
   min-width: 160px !important;
   width: 160px !important;
 }
-.b-table th[aria-sort="none"]::after {
-  content: none !important;
-}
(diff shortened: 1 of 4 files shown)


src/assets/styles/bootstrap-overrides.css
```diff
@@ -32,15 +32,16 @@ legend {
...
 .b-table th[aria-sort]::after {
   display: inline !important;
   font-family: "Font Awesome 5 Free";
   padding-left: 10px !important;
   position: inherit !important;
 }
+.b-table th[aria-sort="none"]::after {
+  color: transparent;
+  content: '\f0d8' !important;
+}
```
johncrossman: Above is the fix to BOAC-2317 – column widths are static, no matter state of sort.

PR #1610, opened 2019-06-06
BOAC-2165: Fixes sidebar links with programmatic routing

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2165

### DIFF
--- a/src/components/sidebar/Cohorts.vue
+++ b/src/components/sidebar/Cohorts.vue
@@ -5,12 +5,14 @@
         Cohorts
       </div>
       <div class="ml-2 mr-2">
-        <router-link
+        <a
           id="cohort-create"
+          class="sidebar-create-link"
           aria-label="Create cohort"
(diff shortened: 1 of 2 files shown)


src/components/sidebar/Cohorts.vue
```diff
@@ -36,11 +39,17 @@
 </template>
 
 <script>
-import UserMetadata from '@/mixins/UserMetadata';
-import Util from '@/mixins/Util';
+import UserMetadata from "@/mixins/UserMetadata";
+import Util from "@/mixins/Util";
+import router from '@/router';
```
raydavis: Obvs not a PR blocker but I'm curious about this single-quoting holdout.

raydavis:
I took the path less traveled, because I was forced to.

src/components/sidebar/Cohorts.vue
lyttam: Odd; not sure how those double quotes got there.

PR #1612, opened 2019-06-07
BOAC-2271, add column 'sids' to cohort_filters for fast batch note creation

johncrossman:

For batch note creation, we must quickly convert a list of cohort ids to proper list o’ SIDs. This new column provides that fast access and is maintained same way of ‘student_count’.  We use SQLAlchemy's `deferred(db.Column)` so `sids` array is only loaded if `undefer(...)` is used.

Other noteworthy changes:
* `to_api_json()` is invoked under-the-hood (i.e., the model's CRUD methods) because it enforces vital biz logic (e.g., update student_count and sids when `update()` is invoked.) IMO, this change makes code more readable.
* The front-end no longer posts 'studentCount'. The value only changes when `/update` and the back-end has it covered.

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -225,7 +225,7 @@ def load_filtered_cohort_counts():
     from boac.models.cohort_filter import CohortFilter
     for cohort in CohortFilter.query.all():
         # Remove!
-        cohort.update_student_count(None)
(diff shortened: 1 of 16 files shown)


paulkerschen:
Looks solid. I didn't even know we had that `np.in1d` fanciness in there, but I guess it's battle-tested.

PR #1614, opened 2019-06-07
BOAC-2271, notes_controller support for batch note creation

johncrossman:

More tests are on the way.


boac/api/notes_controller.py
```diff
@@ -50,30 +52,47 @@ def mark_read(note_id):
...
+            author_dept_codes=dept_codes,
+            subject=subject,
+            body=process_input_from_rich_text_editor(body),
+            topics=topics,
+            sid=_sid,
+            attachments=attachments,
+        )
+    if is_batch_create:
+        for sid in sids:
+            _create_note(sid)
```
johncrossman: I'm going to investigate bulk inserts with SQLAlchemy: https://docs.sqlalchemy.org/en/13/faq/performance.html#i-m-inserting-400-000-rows-with-the-orm-and-it-s-really-slow

PR #1615, opened 2019-06-10
BOAC-2271, notes_controller support for batch note creation

johncrossman:

Using postgres native `json_populate_recordset` for bulk inserts seems to be faster than what's listed in https://docs.sqlalchemy.org/en/13/_modules/examples/performance/bulk_inserts.html   The use of `INSERT ... returning id` allows `db.session.execute(...)` to treat the query as SELECT, getting a list of note `ids` in result-set.

Notice the log message "...records inserted in ... seconds" – I will further profile batch creation once in `boa-dev`.


lyttam:
Thanks for the research.

PR #1619, opened 2019-06-11
BOAC-2341, batch-note feature flag; no more flag on simple create/edit note

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2341


paulkerschen:
If this config could settle down, it would settle down.

PR #1624, opened 2019-06-12
BOAC-1174 Switch student photo source to S3 signed URLs

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1174


raydavis:
Very handy, that boto3.

PR #1625, opened 2019-06-12
BOAC-2238: Makes ASC notes searchable

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2238

First attempt at squeezing ASC notes into the note search results paradigm. Since they don't have a note body, they won't have a note snippet to display - just the advisor name and date.


paulkerschen:
I'm impressed you were able to munge this in so as to avoid another database round trip. Let's see how it plays!

PR #1626, opened 2019-06-12
BOAC-2343, organize flask_caching config per docs; throw out unused '/api/team_groups/all'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2343

Config & code cleanup. Proper flask_caching init is needed if we intend to wire it up properly.


paulkerschen:
No team groups! Only curated groups! Although you could still curate a group for a team.

PR #1628, opened 2019-06-13
BOAC-2345, static assets compressed and cached at Apache layer

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2345

Informed by https://medium.com/@ali.dev/how-to-setup-caching-gzip-compression-in-apache-web-server-with-htaccess-292a0f689553


paulkerschen:
Thanks @johncrossman - this config is actually comprehensible, which is always encouraging.

PR #1629, opened 2019-06-13
BOAC-2324 Cutoff for homepage student-with-alert lists

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2324

### DIFF
--- a/boac/models/alert.py
+++ b/boac/models/alert.py
@@ -34,7 +34,6 @@ from boac.api.errors import BadRequestError
 from boac.externals import data_loch
 from boac.lib.berkeley import current_term_id, section_is_eligible_for_alerts, term_name_for_sis_id
 from boac.lib.util import camelize, unix_timestamp_to_localtime, utc_timestamp_to_localtime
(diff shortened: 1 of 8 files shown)


boac/models/alert.py
```diff
@@ -162,15 +172,12 @@ def result_to_dict(result):
...
+                'firstName': result.get('first_name'),
+                'lastName': result.get('last_name'),
                 'alertCount': alert_counts_by_sid.get(result.get('sid')),
             }
-            scope = get_student_query_scope()
-            if 'UWASC' in scope or 'ADMIN' in scope:
-                result_dict['isActiveAsc'] = result.get('athleticsProfile', {}).get('isActiveAsc')
             return result_dict
-        return [result_to_dict(result) for result in get_full_student_profiles(sids)]
+        return [result_to_dict(result) for result in data_loch.get_basic_student_data(sids)]
```
paulkerschen: This is a side fix to avoid the expensive profile fetch when all we need are ids and names.

PR #1631, opened 2019-06-14
BOAC-2361, use font-awesome Vue component for performance purposes

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2361

See Jira description for details.


paulkerschen:
So many more results when you grep the codebase for `awesome`. As it should be.

PR #1632, opened 2019-06-14
[BOAC-2323] Apache mpm_worker configuration overrides to increase number of requests served.

sandeepmjay:
Load tests + mpm details in JIRA 
https://jira.ets.berkeley.edu/jira/browse/BOAC-2323


paulkerschen:
Thanks for all the research that went into this, @sandeepmjay. One question: .ebextensions/00_ami.config also has configuration values for NumProcesses and NumThreads. Should those be updated to match the changed Apache conf? I'm not sure which would control.

sandeepmjay:
Boa, for the most part, seems to be using the daemon process & threads. In my Load tests, the NumProcess and NumThreads tweaks didn't seem to make a difference. Although some research is still needed on ami configs as a follow-up.

paulkerschen:
Thanks for clarification!

PR #1635, opened 2019-06-14
BOAC-2365, remove expensive CohortFilter.all_owned_by call when verifying user perm to edit/view cohort

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2365

To verify that `current_user` has access to cohort we do not need the expensive:
 `cls.query.filter(cls.owners.any(uid=uid)).order_by(cls.name).all()`

### DIFF
--- a/boac/api/cohort_controller.py
+++ b/boac/api/cohort_controller.py
@@ -24,7 +24,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """
 
 from boac.api.errors import BadRequestError, ForbiddenRequestError, ResourceNotFoundError
-from boac.api.util import get_my_cohorts, is_unauthorized_search
+from boac.api.util import get_cohort_owned_by, is_unauthorized_search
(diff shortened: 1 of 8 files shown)


boac/api/cohort_controller.py
```diff
@@ -39,7 +39,11 @@
 @app.route('/api/cohorts/my')
 @login_required
 def my_cohorts():
-    return tolerant_jsonify(get_my_cohorts())
+    cohorts = []
+    for cohort in CohortFilter.summarize_alert_counts_in_all_owned_by(current_user.id):
+        cohort['isOwnedByCurrentUser'] = True
+        cohorts.append(cohort)
+    return tolerant_jsonify(cohorts)
```
johncrossman: Code snippet above is what was previously in `CohortFilter. get_my_cohorts`. This rest-endpoint is the _only_ place that the expensive student query is needed.

PR #1636, opened 2019-06-17
BOAC-2377, user_session object for flask-login integration; remove hack on authorized_user model

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2377

Until now, BOA had no definitive user-session object. Attributes of the session (eg, the `is_authorized_to_use_boac` method) have been spread across the code-base. This PR gets us more in line with flask_login recommendations: https://flask-login.readthedocs.io/en/latest/

With better encapsulation in user profile object, we can toss out `/api/user/status`.

I added `is_safe_url()` prior to flask.redirect (as recommended in link above) to "validate the value of the next parameter ... If you do not, your application will be vulnerable to open redirects."


boac/api/cohort_controller.py
```diff
@@ -200,11 +200,11 @@ def create_cohort():
         for key in ['inIntensiveCohort', 'isInactiveAsc', 'isInactiveCoe']:
             filter_criteria[key] = to_bool(params.get(key))
     cohort = CohortFilter.create(
-        uid=current_user.get_id(),
+        uid=current_user.get_uid(),
```
johncrossman: Main reason that this PR touches so many files: `current_user.get_id()` is no longer UID getter. Use `get_uid()` instead. (Use `get_id()` to retrieve `authorized_user.id`)

paulkerschen:
Thanks @johncrossman - a welcome change for sure.

PR #1639, opened 2019-06-17
BOAC-2365, performance bump for /all/cohorts: bulk LDAP query, db.all_cohorts() is expensive

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2365


paulkerschen:
Why, it _does_ look slimmer.

PR #1644, opened 2019-06-19
BOAC-2377, cache user_session json by user_id; flush on login, demoMode toggle

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2377

In flask-login we `login_user(..., remember=True)` so this cached user-session-json will persist across browser restarts. This cuts `/api/my/profile`time in half which is one of the main offenders in blocking the "painting" if initial page load. (Cache is refreshed when user logs in again or toggles demo-mode.)


paulkerschen:
Like Andy Warhol, we must paint fast. Makes sense to me.

PR #1657, opened 2019-06-21
BOAC-2409, obsolete cohort filter values (eg, 'Sociology BA') should not bust front-end

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2409

Cohort created by COE advisor, in March, had 'Sociology BA' major selected. 'Sociology BA' is no longer in COE scope. This does not break queries but it does disrupt front-end when rendering menu options. I'm working around the problem by tolerating nulls in menu render.

This bug is unlikely to recur after we desilo-ize but if it does, we can consider a more sophisticated fix.


paulkerschen:
Looks like a convenient Band-Aid to me.

PR #1661, opened 2019-06-24
BOAC-2279, refresh student in /student if note created per sid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2279

On front-end, our vue mixin `EditNoteSession` is now `StudentEditSession`, lending itself to a nice code pattern whereby student profile can be refreshed when, for example, a note is created via batch-create. Using `$eventHub` to emit events was not a good solution here.

In order to make this work we needed two small BOA API changes: (1) `/api/student` takes SID instead of UID and (2) if create note is  "batch-mode" then server always return an array of SIDs.

IMO, it's a fairly intuitive arrangement.

### DIFF
--- a/src/views/Student.vue
+++ b/src/views/Student.vue
@@ -17,7 +17,7 @@
         </div>
       </div>
       <div class="m-3">
-        <AcademicTimeline :student="student" />
+        <AcademicTimeline :key="loadedAt" :student="student" />
         <AreYouSureModal
           v-if="showAreYouSureModal"
           :function-cancel="cancelTheCancel"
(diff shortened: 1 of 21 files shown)


src/views/Student.vue
```diff
@@ -17,7 +17,7 @@
         </div>
       </div>
       <div class="m-3">
-        <AcademicTimeline :student="student" />
+        <AcademicTimeline :key="loadedAt" :student="student" />
```
johncrossman: Force component reload when `:key` changes.

paulkerschen:
The StudentEditSession is definitely a nicer pattern, @johncrossman. My only question would be: is there an issue in having user-visible URLs (which might be copied and passed around, etc) based on SID, which is technically protected data, as opposed to UID, which is public?

johncrossman:
@paulkerschen - I will amend and get SID out of URL path. Thanks

johncrossman:
Amendment is in and SID is out of URL path. I had to add `/api/student/by_sid/:sid` and consolidate a bit of backend code. However, the diff is still clean – `test_student_controller` got some overdue cleanup.

paulkerschen:
This is life at a university where everyone gets multiple IDs. Nicely handled!

PR #1662, opened 2019-06-24
BOAC-2415, NewNoteModal > non-batch mode > targetStudentCount is one

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2415

A few forgotten snippets.


paulkerschen:
One Sid is just as important as a thousand Sids, in BOA's student-first advising math.

PR #1663, opened 2019-06-24
BOAC-2417, when batch note create, mark as read by author

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2417

You wrote it, you read it.


paulkerschen:
While plenty of writing out there leaves me skeptical that the authors are in fact reading their own work, let's give advisors the benefit of the doubt.

PR #1666, opened 2019-06-25
BOAC-2399, valid termId in element.id

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2399


raydavis:
Looks good, @johncrossman -- waiting on T.

PR #1674, opened 2019-06-26
BOAC-2418, GoogleAnalytics for note creation; split /note/create and /note/batch_create

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2418


paulkerschen:
Tidier for sure.

PR #1675, opened 2019-06-26
[QA] BOAC-2428, odd case of ::after in bootstrap-override requires generic unicode

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2428

font-good-enough instead of font-awesome. See https://en.wikipedia.org/wiki/Geometric_Shapes


paulkerschen:
I was hoping for something like CIRCLE WITH ALL BUT UPPER LEFT QUADRANT BLACK, but I guess you use the awesome that's given to you.

PR #1677, opened 2019-06-26
BOAC-2418, more and consistent use of gaTrackEvent in src/api

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2418

* consistency in use of `@/api/api-utils`
* smooth out edges of Google Analytics integration

### DIFF
--- a/src/mixins/UserMetadata.vue
+++ b/src/mixins/UserMetadata.vue
@@ -13,7 +13,6 @@ export default {
   },
   methods: {
     ...mapActions('user', [
-      'gaEvent',
       'gaCohortEvent',
       'gaCuratedEvent',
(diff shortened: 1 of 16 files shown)


src/mixins/UserMetadata.vue
```diff
@@ -13,7 +13,6 @@ export default {
   },
   methods: {
     ...mapActions('user', [
-      'gaEvent',
```
johncrossman: Generic `gaEvent` is not available via mixin cuz we want to enforce conventions with `gaCurated`, `gaCohort`, etc.

paulkerschen:
What are people searching for? We gotta know!

PR #1679, opened 2019-06-27
BOAC-2296, minor package/py third-party upgrades; lock down runtime-deps in package.json

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2296

And, `FEATURE_FLAG_BATCH_NOTES` is default True in boa-dev.


paulkerschen:
It's the ten-thousand-mile tune-up.

PR #1682, opened 2019-06-27
BOAC-1853 BOAC-1854 Filter search by individual advisor and/or student

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1853
https://jira.ets.berkeley.edu/jira/browse/BOAC-1854

Including a first-pass autocomplete component. The most obvious thing it's missing is browser-side caching of suggestions, which would make the whole experience less jittery; but in my local testing it seemed usable even so.

### DIFF
--- a/boac/api/notes_controller.py
+++ b/boac/api/notes_controller.py
@@ -26,6 +26,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 import urllib.parse
 
 from boac.api.errors import BadRequestError, ForbiddenRequestError, ResourceNotFoundError
+from boac.externals import data_loch
 from boac.lib.http import tolerant_jsonify
 from boac.lib.util import is_int, process_input_from_rich_text_editor
 from boac.merged.advising_note import get_boa_attachment_stream, get_legacy_attachment_stream, note_to_compatible_json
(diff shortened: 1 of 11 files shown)


src/views/Search.vue
```diff
@@ -178,8 +178,6 @@ export default {
         true,
         false,
         this.noteOptions,
-        this.user.isAsc ? false : null,
-        this.user.isCoe ? false : null
```
paulkerschen: I think these elements are dead code from a time when the front end got to decide whether ASC or COE inactives were filtered out of searches.

boac/api/notes_controller.py
```diff
@@ -168,6 +169,25 @@ def delete_note(note_id):
...
+    def _author_feed(a):
+        return {
+            'label': ' '.join([a.get('first_name', ''), a.get('last_name', '')]),
+            'sid': a.get('sid'),
+            'uid': a.get('uid'),
+        }
+    return tolerant_jsonify([_author_feed(a) for a in authors])
+
+
 @app.route('/api/notes/topics', methods=['GET'])
```
paulkerschen: Tests for new back-end methods here and below to come in a follow-up PR.

johncrossman:
Autocomplete? I'd say, auto-awesome.

PR #1684, opened 2019-06-27
BOAC-2277, add Student, by name or SID (with autosuggest), in 'New Note' form

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2277

Minor override options in `Autocomplete` component.

### DIFF
--- a/src/components/note/CreateNoteAddStudent.vue
+++ b/src/components/note/CreateNoteAddStudent.vue
@@ -1,0 +1,99 @@
+<template>
+  <div>
+    <div>
+      <label
+        for="create-note-add-student"
+        class="font-size-14 input-label text mt-2">
+        <span class="sr-only">Add a </span><span class="font-weight-bolder">Student</span> (name or SID)
+      </label>
+    </div>
+    <div class="mb-2">
+      <Autocomplete
+        id="create-note-add-student"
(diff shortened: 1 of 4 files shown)


paulkerschen:
@johncrossman Overrides make sense! One thing I learned in the course of writing this component is that `v-model` is just shorthand for a `value` property plus an `@input` listener. So I’m curious whether just adding `@input=“addStudent”` to your Autocomplete element would allow you to do without `studentModel` and the watcher.

johncrossman:
@paulkerschen - Just experimented with your suggestion and it works. Amendment coming soon.

src/components/note/CreateNoteAddStudent.vue
```diff
@@ -0,0 +1,99 @@
...
+    </div>
+    <div class="mb-2">
+      <Autocomplete
+        id="create-note-add-student"
+        :key="resetAutoCompleteKey"
+        class="w-75"
+        :on-esc-form-input="onEscFormInput"
+        :show-add-button="true"
+        :source="findStudentsByNameOrSid"
+        @input="addStudent">
```
johncrossman: Amended.

paulkerschen: 👌

PR #1688, opened 2019-06-28
BOAC-2435 Fix timezone adjustment in note search date filters

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2435

### DIFF
--- a/boac/lib/util.py
+++ b/boac/lib/util.py
@@ -80,7 +80,7 @@ def localize_datetime(dt):
 
 def localized_timestamp_to_utc(_str):
     localized_datetime = datetime.strptime(_str, '%Y-%m-%dT%H:%M:%S')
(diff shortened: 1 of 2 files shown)


boac/lib/util.py
```diff
@@ -80,7 +80,7 @@ def localize_datetime(dt):
 
 def localized_timestamp_to_utc(_str):
     localized_datetime = datetime.strptime(_str, '%Y-%m-%dT%H:%M:%S')
-    localized_datetime.replace(tzinfo=pytz.timezone(app.config['TIMEZONE']))
+    localized_datetime = localized_datetime.replace(tzinfo=pytz.timezone(app.config['TIMEZONE']))
```
paulkerschen: This back-end assignment goof is the principal bug. The front-end change below is for better tracking of daylight savings adjustments, but those usually don't surface in a date-only display.

PR #1690, opened 2019-06-28
BOAC-2440, loadConfig with or without authenticated user

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2440

What a difference an if-block makes.

Review with `?w=1`


johncrossman:
Blame goes to PR https://github.com/ets-berkeley-edu/boac/pull/1683/files#diff-f13b284b5b1dd5ddc12805ea4a616b1bL64

paulkerschen:
I figured it was Google's fault somehow.

PR #1695, opened 2019-07-01
[QA] BOAC-2430, flask-login persists user_id as string; we force int when de-serialized

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2430

Type matters.


raydavis:
"God made the int; all else is the work of man."

PR #1697, opened 2019-07-01
NOJIRA Fix erroneous ORM mapping of Integer ForeignKey to String

raydavis:
This didn't cause BOAC-2430, but it's certainly distracting.


paulkerschen:
Dear me!

PR #1698, opened 2019-07-01
BOAC-2379 All advisors can see all students, but not all per-student data

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2379

Our concept of "scope" doesn't go away, but it now has less to do. A search for students via name or cohort filter now pulls in all students that BOA knows about, with their profiles and advising notes accessible, no matter who you are.

Restrictions that remain in place:
 - Advisors can't view cohorts belonging to other departments.
 - Advisors can't perform cohort searches with data from other departments as a criterion.
 - Advisors don't see data from other departments in student feeds.


johncrossman:
Don't fence me in (with your scoped SQL queries)!

PR #1701, opened 2019-07-03
BOAC-2443: Fixes boxplot tooltip bleed-through

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2443


johncrossman:
BOA in 3D!

PR #1706, opened 2019-07-08
BOAC-2449, if student was added to batch-note then do not auto-suggest again

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2449

### DIFF
--- a/src/components/note/CreateNoteAddStudent.vue
+++ b/src/components/note/CreateNoteAddStudent.vue
@@ -15,7 +15,7 @@
         class="w-75"
         :on-esc-form-input="onEscFormInput"
         :show-add-button="true"
-        :source="findStudentsByNameOrSid"
+        :source="studentsByNameOrSid"
         @input="addStudent">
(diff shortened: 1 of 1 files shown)


src/components/note/CreateNoteAddStudent.vue
```diff
@@ -90,6 +89,14 @@ export default {
         this.removeSid(student.sid);
         this.alertScreenReader(`Student '${student.name}' removed`);
       }
+    },
+    studentsByNameOrSid(query, limit) {
+      const sids = this.map(this.addedStudents, 'sid');
+      return new Promise(resolve => {
+        findStudentsByNameOrSid(query, limit).then(students => {
```
paulkerschen: @johncrossman Is `this` not needed on this line? Something to do with the promise changing scope?

paulkerschen: Never mind, I see it's a function import here, not a mixin.

PR #1712, opened 2019-07-09
BOAC-2449, recalculateStudentCount() in batch-note, even for simple add/remove SID

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2449


paulkerschen:
It's simple. It's smart. It's BOA.

PR #1717, opened 2019-07-11
BOAC-2420, export list of BOA users as CSV from /admin

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2420

Download link appears as:
image

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -27,7 +27,7 @@ from boac.api import errors
 from boac.api.util import admin_required, authorized_users_api_feed
 from boac.lib import util
 from boac.lib.berkeley import BERKELEY_DEPT_CODE_TO_NAME
-from boac.lib.http import tolerant_jsonify
+from boac.lib.http import response_with_csv_download, tolerant_jsonify
 from boac.merged import calnet
 from boac.models.authorized_user import AuthorizedUser
 from flask import current_app as app, request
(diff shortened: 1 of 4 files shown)


boac/api/user_controller.py
```diff
@@ -63,6 +63,50 @@ def user_by_uid(uid):
...
+        in_demo_mode = request.get_json().get('demoMode', None)
+        if in_demo_mode is None:
+            raise errors.BadRequestError('Parameter \'demoMode\' not found')
+        user = AuthorizedUser.find_by_id(current_user.get_id())
+        user.in_demo_mode = bool(in_demo_mode)
+        current_user.flush_cached()
+        app.login_manager.reload_user()
+        return tolerant_jsonify(current_user.to_api_json())
+    else:
+        raise errors.ResourceNotFoundError('Unknown path')
```
johncrossman: This is an odd diff. No change to `demo_mode`.

raydavis: Thanks for the timesaving note, @johncrossman

boac/api/user_controller.py
```diff
@@ -63,6 +63,50 @@ def user_by_uid(uid):
...
+
+
+@app.route('/api/users/csv')
+@admin_required
+def download_boa_users_csv():
+    rows = []
+    for dept in _get_boa_user_groups():
+        for user in dept['users']:
+            email = user.get('campusEmail') or user.get('email')
+            rows.append(
```
raydavis: Mild curiosity rather than objection: What made you decide to format explicitly rather than using something like `csv.DictWriter`?

johncrossman: @raydavis - I liked the conciseness of list-of-lists but `DictWriter` would be safest w.r.t escaping special chars. I'll amend.

johncrossman:
Amended.  cc: @raydavis

raydavis:
Thank you, @johncrossman

PR #1719, opened 2019-07-12
BOAC-2477, remove FEATURE_FLAG_BATCH_NOTES from code + config

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2477

v2.21 release instructions have been updated.


paulkerschen:
It was a grand old flag.

PR #1723, opened 2019-07-15
BOAC-2310, disable 'Export List' with precarious bool expression

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2310

See latest comment in Jira.


paulkerschen:
Precarious but precious.

PR #1724, opened 2019-07-16
BOAC-2479, do not show updated_at if note has not been modified since creation

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2479

Lenient comparison between `created_at` and `updated_at`.


paulkerschen:
Let creation and updates never be confused again.

PR #1729, opened 2019-07-16
[QA] BOAC-2482, if cancel note edit then properly preserve sid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2482


raydavis:
Looks sensible, but Travis Rules Everything Around Me.

PR #1730, opened 2019-07-16
BOAC-2482, if cancel note edit then properly preserve sid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2482

The qa PR is #1729


paulkerschen:
Sid? Sid.

PR #1732, opened 2019-07-17
BOAC-2484: highcharts, but not too high

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2484


paulkerschen:
Settle down, you guys.

PR #1736, opened 2019-07-18
BOAC-2496 Ignore planless advisor-student associations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2496

Most advisor-student associations from our SIS source are associated with an academic plan, but a handful of them represent the plan as either null or a blank string. Cohort filters aren't playing well with this state of affairs; let's suppress blank plans in the app and look into fixing the data upstream.


johncrossman:
Painless "planless".

PR #1737, opened 2019-07-18
BOAC-2487, new advising note topic: Minors

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2487


paulkerschen:
Minors now!!

PR #1746, opened 2019-07-22
BOAC-2509 Add soft delete logic to authorized_users

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2509


johncrossman:
@paulkerschen This looks good. Travis is a bit unhappy. Will we show deleted users on `/admin` page?

paulkerschen:
Thanks for review, @johncrossman. We presumably don't want deleted users on the admin page and I added in a filter to that effect. The Travis failure was something that I couldn't reproduce locally and went away when I flipped the order of two tests; unfortunately, I have no explanation.

PR #1747, opened 2019-07-23
BOAC-2353 Refresh authorized users from SIS advising tables

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2353

Tests to come in a follow-up; I wanted to give the job itself an early opportunity to get kicked around.

### DIFF
--- a/boac/models/authorized_user.py
+++ b/boac/models/authorized_user.py
@@ -69,10 +69,10 @@ class AuthorizedUser(Base):
 
     @classmethod
     def create_or_restore(cls, uid, is_admin=False, in_demo_mode=False):
-        existing_note = cls.query.filter_by(uid=uid).first()
-        if existing_note:
-            existing_note.deleted_at = None
-            return existing_note
+        existing_user = cls.query.filter_by(uid=uid).first()
+        if existing_user:
+            existing_user.deleted_at = None
+            return existing_user
         else:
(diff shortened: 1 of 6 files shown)


boac/models/authorized_user.py
```diff
@@ -67,6 +67,15 @@ def __repr__(self):
                     created={self.created_at}>
                 """
 
+    @classmethod
+    def create_or_restore(cls, uid, is_admin=False, in_demo_mode=False):
+        existing_note = cls.query.filter_by(uid=uid).first()
```
johncrossman: @paulkerschen Is `existing_note` a proper var name here?

paulkerschen: @johncrossman It is not! Just pushed a follow-up commit.

boac/models/university_dept.py
```diff
@@ -57,3 +60,22 @@ def create(cls, dept_code, dept_name, automate_memberships):
         db.session.add(dept)
         std_commit()
         return dept
+
+    def delete_all_members(self):
+        sql = """
+            DELETE FROM university_dept_members WHERE university_dept_id = :id;
+            UPDATE authorized_users SET deleted_at = now()
+                WHERE is_admin = false
```
johncrossman: @paulkerschen Not sure it matters: `IS false`  instead of  `= false`

johncrossman:
Looks good @paulkerschen  Waiting on a happy Travis

PR #1752, opened 2019-07-24
BOAC-2526: Uses bulk LDAP query to load user profiles

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2526


paulkerschen:
Simple and effective.

PR #1756, opened 2019-07-25
BOAC-2452: Sets inactivity lifespan on session cookie

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2452

The goal is to log a user out after they close the browser - this will also log them out if the browser hasn't made any requests for 60 minutes.

This will likely be annoying for developers; I recommend overriding `INACTIVE_SESSION_LIFETIME` in your local config.


paulkerschen:
Thanks for making the annoyance configurable, @lyttam.

PR #1761, opened 2019-07-26
BOAC-2452: Keeps session alive unless browser is closed

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2452

### DIFF
--- a/config/default.py
+++ b/config/default.py
@@ -61,7 +61,9 @@ CSRF_SESSION_KEY = 'secret'
 # Used to encrypt session cookie.
 SECRET_KEY = 'secret'
 # Minutes of inactivity before session cookie is destroyed
-INACTIVE_SESSION_LIFETIME = 60
+INACTIVE_SESSION_LIFETIME = 20
+# Millisecond interval for request to keep session alive
+PING_FREQUENCY = 900000
 
 TIMEZONE = 'America/Los_Angeles'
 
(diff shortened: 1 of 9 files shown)


config/default.py
```diff
@@ -61,7 +61,9 @@
 # Used to encrypt session cookie.
 SECRET_KEY = 'secret'
 # Minutes of inactivity before session cookie is destroyed
-INACTIVE_SESSION_LIFETIME = 60
+INACTIVE_SESSION_LIFETIME = 20
+# Millisecond interval for request to keep session alive
+PING_FREQUENCY = 900000
```
lyttam: 900000 ms = 15 minutes

paulkerschen:
@lyttam Nothing against the ping-pong workflow, but on both BOA and Nessie, our current `/api/ping` is called that because it's regularly pinged by Nagios to check app status. Any changes to it will require Darlene to change Nagios configs.

If the existing ping doesn't work for browser session verification, do you think we could set up a differently named endpoint to live side by side?

lyttam:
@paulkerschen thanks for pointing that out.  I didn't want to use the existing ping since it performs a couple of database queries.  Do you think it's a good idea to use the existing ping to keep the session alive?

paulkerschen:
@alyttle At once every 15 minutes, I think the existing ping is ok to use and won’t incur much overhead.

lyttam:
@paulkerschen amended to use the existing ping endpoint.

paulkerschen:
Thanks @lyttam!

PR #1770, opened 2019-07-31
BOAC-2538: Disables export link until download has started

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2538


paulkerschen:
Makes sense.

PR #1775, opened 2019-08-02
BOAC-2524 Remove vestiges of PHYSI department from test code

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2524


raydavis:
I think it's unfair that Chemistry gets a college & Physics doesn't, but I guess we'll have to take that up outside GitHub.

PR #1778, opened 2019-08-07
[QA] BOAC-2574 Suppress null advising note topics from Redshift

paulkerschen:
QA-first PR for https://jira.ets.berkeley.edu/jira/browse/BOAC-2574.

https://github.com/ets-berkeley-edu/nessie/pull/505 cleared out null topics from RDS, but turns out BOA is still querying Redshift for notes. The right fix is to have BOA start querying RDS (and maybe keep null topics out of our internal Redshift as well). But rather than make an architectural change the day before release, this should plug the gap.


raydavis:
"I'm at the Pizza Hut." "I'm at the Taco Bell."

PR #1786, opened 2019-08-09
BOAC-2356: Records user logins

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2356


raydavis:
Nice & straightforward.

PR #1788, opened 2019-08-09
BOAC-2421: Fills gaps in expected grad terms

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2421

### DIFF
--- a/boac/merged/cohort_filter_options.py
+++ b/boac/merged/cohort_filter_options.py
@@ -27,7 +27,7 @@ from copy import deepcopy
 
 from boac.api.util import authorized_users_api_feed
 from boac.externals import data_loch
-from boac.lib.berkeley import BERKELEY_DEPT_NAME_TO_CODE, COE_ETHNICITIES_PER_CODE, term_name_for_sis_id
+from boac.lib.berkeley import BERKELEY_DEPT_NAME_TO_CODE, COE_ETHNICITIES_PER_CODE, term_ids_range, term_name_for_sis_id
 from boac.merged import athletics
 from boac.merged.calnet import get_csid_for_uid
 from boac.merged.student import get_student_query_scope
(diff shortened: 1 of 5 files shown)


boac/merged/cohort_filter_options.py
```diff
@@ -394,8 +394,10 @@ def _genders():
 
 
 def _grad_terms():
-    term_ids = [r['expected_grad_term'] for r in data_loch.get_expected_graduation_terms()]
-    return [{'name': term_name_for_sis_id(term_id), 'value': term_id} for term_id in term_ids]
+    earliest_term_id = data_loch.get_min_expected_graduation_term()['term']
+    latest_term_id = add_terms(future_term_id(), 12)
+    term_ids = term_ids_range(earliest_term_id, latest_term_id)
+    return [{'name': ' '.join(term_name_for_sis_id(term_id).split()[::-1]), 'value': term_id} for term_id in term_ids]
```
lyttam: Reversing term name for readability, so "Summer 2019" becomes "2019 Summer".

boac/merged/cohort_filter_options.py
```diff
@@ -394,8 +394,10 @@ def _genders():
 
 
 def _grad_terms():
-    term_ids = [r['expected_grad_term'] for r in data_loch.get_expected_graduation_terms()]
-    return [{'name': term_name_for_sis_id(term_id), 'value': term_id} for term_id in term_ids]
+    earliest_term_id = data_loch.get_min_expected_graduation_term()['term']
+    latest_term_id = add_terms(future_term_id(), 12)
```
paulkerschen: @lyttam Are we confident that now-plus-12-terms will always cover the required range? I get the logic, but sometimes what's in the database defies logic.

lyttam: @paulkerschen I'm not confident about it, and it's definitely up for debate. I could pick a higher number or make it configurable.

paulkerschen: @lyttam What if you just selected the max from the database, same as you're selecting the min for the start of the interval?

lyttam: @paulkerschen everyone agrees that's a good idea! I'll amend.

lyttam:
@paulkerschen All amended now.

paulkerschen:
Thanks, @lyttam. Term math!

PR #1791, opened 2019-08-13
BOAC-2582, GUEST user gets /cohort filter options

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2582

...with tests.


paulkerschen:
Thanks for expanding the guest suite, @johncrossman.

PR #1793, opened 2019-08-13
BOAC-2537, do not alter 'isModified' (cohort filters) when paginate

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2537

### DIFF
--- a/src/store/modules/cohort-edit-session.ts
+++ b/src/store/modules/cohort-edit-session.ts
@@ -126,6 +126,31 @@ const mutations = {
   }
 };
 
+export function $_cohortEditSession_applyFilters({ commit, state }, orderBy: string) {
+  return new Promise(resolve => {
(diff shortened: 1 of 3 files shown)


src/store/modules/cohort-edit-session.ts
```diff
@@ -126,6 +126,31 @@ const mutations = {
   }
 };
 
+export function $_cohortEditSession_applyFilters({ commit, state }, orderBy: string) {
```
paulkerschen: Just out of curiosity, is this function name formatting style a TypeScript thing?

johncrossman: @paulkerschen  It seems to be Vue best practice: https://vuejs.org/v2/style-guide/#Private-property-names-essential    Having shared/private methods in a 'store' module is a strange and mysterious thing.

PR #1794, opened 2019-08-14
BOAC-2541, note_template/topic/attachment tables and SQLAlchemy models

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2541

Note:
* Soft-delete for `note_template_attachments` (see `deleted_at` column) because file uploads might be audited.
* `note_template_topics` table has neither `author_uid` nor `deleted_at`.  Those attributes will be tracked in the actual notes.
* If an authorized_user is deleted then we `DELETE CASCADE` per `note_templates.creator_id`
* The initial tests in `test_note_templates_controller` are for basic verification of the models.

### DIFF
--- a/boac/api/note_templates_controller.py
+++ b/boac/api/note_templates_controller.py
@@ -1,0 +1,41 @@
+"""
+Copyright ©2019. The Regents of the University of California (Regents). All Rights Reserved.
+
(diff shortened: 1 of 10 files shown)


paulkerschen:
Looks like models!

PR #1802, opened 2019-08-19
BOAC-2575, front-end 'api/note-templates.ts' to match /note_templates API

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2575

No front-end components will use this API until after `qa` branch cut. (Includes typo fix in `notes.ts`.)


paulkerschen:
Looks wired.

PR #1803, opened 2019-08-19
BOAC-2594 Include CalNet attributes in daily refresh

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2594


paulkerschen:
Thanks, @raydavis - this seems to do the job for BOA users. In addition to `calnet_user_for_uid_%` keys, we're also saddled with a large number of `calnet_user_for_csid_%` cache keys, which come into play when showing SIS advising notes for which we know author CSID but not UID. Refresh logic and requirements may be a little different for these, so they're probably best addressed as a separate task.

PR #1804, opened 2019-08-19
BOAC-2593 Move search filters into subqueries for performance

paulkerschen:
Per https://jira.ets.berkeley.edu/jira/browse/BOAC-2593, we need to consolidate our three-and-counting advising note indexes into a single source, but this ugly-SQL solution gets us around some serious slowdown in the interim.


johncrossman:
Ugly SQL does the job.

PR #1807, opened 2019-08-20
BOAC-2606, complex computed value not updating when note deleted

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2606

There are limits to the power of `comupted` values in Vue. In this case (and IMO), a complex `v-for` is best. (PR includes fix to benign bug in how we remove id from `openMessages` list when note deletion is confirmed.)

QA PR will happen after this is reviewed.


paulkerschen:
It was too good to be true, that complex `computed`.

PR #1809, opened 2019-08-20
BOAC-2612, allNotesExpanded = false when you click to a new tab

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2612

### DIFF
--- a/src/components/student/profile/AcademicTimeline.vue
+++ b/src/components/student/profile/AcademicTimeline.vue
@@ -548,7 +548,10 @@ export default {
     setFilter(filter) {
       this.searchResults = null;
       this.notesQuery = null;
-      this.filter = filter;
+      if (filter !== this.filter) {
+        this.filter = filter;
+        this.allNotesExpanded = false;
+      }
     },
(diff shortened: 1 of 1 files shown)


src/components/student/profile/AcademicTimeline.vue
```diff
@@ -546,9 +546,12 @@ export default {
       }
     },
     setFilter(filter) {
-      this.searchResults = null;
-      this.notesQuery = null;
-      this.filter = filter;
+      if (filter !== this.filter) {
+        this.searchResults = null;
+        this.notesQuery = null;
```
paulkerschen: @johncrossman Makes sense, but one of the design requirements in https://jira.ets.berkeley.edu/jira/browse/BOAC-2261 was that re-selecting the "Notes" filter when you're viewing search results clears the search, so these first two assignments might need to move outside the conditional.

johncrossman: Thanks @paulkerschen – I'll amend.

PR #1814, opened 2019-08-20
BOAC-2608 Fixes to expected graduation term filter options

paulkerschen:
Like the customer ordered in https://jira.ets.berkeley.edu/jira/browse/BOAC-2608:
- Revert term interpolation (#1788);
- Order latest to earliest;
- Insert a divider between present term and past terms.

I don't love the 'divider' convention but I didn't want to further redo the filter framework for the sake of putting one divider in one dropdown.


johncrossman:
Flip it, divide, reverse it, and done.

PR #1817, opened 2019-08-21
BOAC-2548, delete template modal and API call

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2548


paulkerschen:
I can't wait to declutter some bad templates!

PR #1818, opened 2019-08-21
BOAC-2575, rename api/store components associated with notes/templates

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2575

Refactoring for readability.


paulkerschen:
It's not your grandma's api/store.

PR #1821, opened 2019-08-22
BOAC-2575, discard template edit; consistent naming for modals

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2575

Small steps toward template CRUD support. Plus, consistent var naming for modal-related.

PR #1823, opened 2019-08-22
BOAC-2632, take care when invoking 'transientId' on a transient object

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2632

Bug introduced by PR #1821


raydavis:
Thanks for the quick fix, @johncrossman !

PR #1840, opened 2019-08-28
NS-658 Revise advising notes search query against consolidated index

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/NS-658

It's still not beautiful SQL, but at least there's less of it.


johncrossman:
Less is more

PR #1841, opened 2019-08-28
BOAC-2655, no more click-to-add on topics dropdown; options disable properly when load template

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2655
https://jira.ets.berkeley.edu/jira/browse/BOAC-2450

And, `putFocus` appropriately.


paulkerschen:
So long, click-to-add!

PR #1846, opened 2019-08-29
BOAC-2589: Adds pending_advisees table

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2589

### DIFF
--- a/boac/models/manually_added_advisee.py
+++ b/boac/models/manually_added_advisee.py
@@ -1,0 +1,47 @@
+"""
+Copyright ©2019. The Regents of the University of California (Regents). All Rights Reserved.
+
+Permission to use, copy, modify, and distribute this software and its documentation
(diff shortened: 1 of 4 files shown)


raydavis:
Question about the naming: does `pending_advisees` mean the table will be truncated once the SIDs show up in Nessie's regular nightly runs?

lyttam:
@raydavis I guess it could go either way, but I was imagining we might want to do something like that.  "Pending" sounds temporary; it could be named `manually_added_advisees` if they are going to stay.

paulkerschen:
@raydavis @lyttam My two cents are that the number of advisees might be small enough that we could just keep them around indefinitely, and skip the complication of syncing to Nessie, verifying the sync, and such.

lyttam:
Sounds good to me.  I changed the name.

PR #1852, opened 2019-09-03
BOAC-2664, note body is nullable

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2664


paulkerschen:
Sometimes saying it with the subject is enough.

PR #1854, opened 2019-09-03
BOAC-2669, note_template body can be null

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2669

As seen in BOAC-2664


paulkerschen:
@johncrossman Sorry, should have asked before merging, but as a follow-up, should we drop `NOT NULL` from `note_templates.body` in `schema.sql`? The column definitions in `note.py` and `note_template.py` should also perhaps change to `nullable=True`.

johncrossman:
@paulkerschen Whoops. I'll include that in separate PR.

PR #1855, opened 2019-09-03
BOAC-2597 Distinguish ASC, CoE inactivity

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2597


johncrossman:
One person's inactive is another person's...

PR #1859, opened 2019-09-04
BOAC-2665: Adds L&S minor advisors to the major advising pool

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2665


paulkerschen:
It's a minor miracle!

PR #1863, opened 2019-09-05
BOAC-2654 Better display for inactive student plans

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2654


raydavis:
Inactive-and-teresting.

PR #1865, opened 2019-09-05
BOAC-2651, front-end for rename template; moved create-note components to 'note/create' dir

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2651

And, moved rename/delete methods to Header component to reduce `props` complexity.


paulkerschen:
Big modals have little modals upon their backs to bite 'em,
And little modals have lesser modals, and so, ad infinitum.

PR #1866, opened 2019-09-06
BOAC-2673, improved markup in note templates feature

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2673
https://jira.ets.berkeley.edu/jira/browse/BOAC-2681


paulkerschen:
Better than a Sharpie correction.

PR #1867, opened 2019-09-06
BOAC-2675 Use low-tech dropdown separators that don't render as fake list items

paulkerschen:
For accessibility bug https://jira.ets.berkeley.edu/jira/browse/BOAC-2675.


johncrossman:
You went old school on 'em.

PR #1871, opened 2019-09-07
BOAC-2668, progress on create-note is shown (spinner) at top of student timeline

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2668

While note is being created, user sees small spinner + note subject + today's date at top of academic timeline. One fix required: the _Promise_ returned by `createAdvisingNote()` in vuex store needed to come from the `action`.

### DIFF
--- a/src/components/note/create/CreateNoteModal.vue
+++ b/src/components/note/create/CreateNoteModal.vue
@@ -25,7 +25,6 @@
           'd-none': isNil(mode),
           'modal-open': mode === 'docked',
           'modal-open modal-minimized': mode === 'minimized',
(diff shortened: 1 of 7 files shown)


src/components/note/create/CreateNoteModal.vue
```diff
@@ -286,7 +285,6 @@ export default {
         this.showAlert('Creating template...', 60);
       }
       createNoteTemplate(title, this.model.subject, this.model.body, this.model.topics, this.model.attachments).then(() => {
-        this.onSuccessfulCreate();
```
johncrossman: This callback only necessary after create-note.

PR #1878, opened 2019-09-09
BOAC-2647 Access to inactive students via SID search and curated groups

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2647

Let the bugs begin. I apologize for tossing this much new and rearranged code into the works without corresponding test coverage; given our timeline, it seemed important to get this into shared code as soon as possible to avoid merge conflicts. I'll plan to follow up with some fleshed-out fixtures and new tests.

### DIFF
--- a/boac/api/student_controller.py
+++ b/boac/api/student_controller.py
@@ -25,7 +25,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 
 from boac.api.errors import BadRequestError, ResourceNotFoundError
 from boac.api.util import put_notifications
-from boac.externals.data_loch import match_students_by_name_or_sid
+from boac.externals.data_loch import match_students_by_name_or_sid, query_historical_sids
 from boac.lib.http import tolerant_jsonify
 from boac.merged.student import get_student_and_terms_by_sid, get_student_and_terms_by_uid, query_students
 from flask import current_app as app, request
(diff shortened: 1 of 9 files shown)


boac/api/student_controller.py
```diff
@@ -82,6 +82,7 @@ def validate_sids():
         else:
             summary = []
             available_sids = query_students(sids=sids, sids_only=True)['sids']
+            available_sids += [row['sid'] for row in query_historical_sids(sids)]
```
raydavis: It might be better to only query the not-yet-returned SIDs, since rows aren't automatically removed from the hist_enr tables when a non-advisee gets promoted. I think the `sids` returned from `query_students` is an array rather than a set, which would result in duplicate summaries.

paulkerschen: Thanks, @raydavis - I think that logical point is likely to surface at a couple other points in code.

boac/merged/student.py
```diff
@@ -209,46 +210,70 @@ def get_summary_student_profiles(sids, term_id=None):
     term_gpas = get_term_gpas_by_sid(sids)
     benchmark('end term GPA query')
 
+    if include_historical:
+        benchmark('begin historical profile supplement')
+        historical_profile_rows = data_loch.get_historical_student_profiles_for_sids(sids)
+        historical_profiles = [json.loads(row['profile']) for row in historical_profile_rows]
+        _merge_photo_urls(historical_profiles)
```
raydavis: The hist_enr SIDs won't have imported photo data.

paulkerschen: Indeed it won't; that's just to get a placeholder photo URL in place so as to show a generic student head rather than a broken image. I'll add a comment to clarify.

raydavis:
Thanks for jumping on this, @paulkerschen . As you may have noticed (and been politely silent about), I've been skipping a whole lot of test-coverage myself.

Sounds like you have some amendments brewing and so I'll wait to merge.

paulkerschen:
Thanks for the review, @raydavis - amendments pushed.

PR #1879, opened 2019-09-10
BOAC-2692: Adds SIDs to manually_added_advisees when searched or curated

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2692

### DIFF
--- a/boac/merged/student.py
+++ b/boac/merged/student.py
@@ -32,6 +32,7 @@ from boac.externals import data_loch, s3
 from boac.lib import analytics
 from boac.lib.berkeley import current_term_id, future_term_id, term_name_for_sis_id
 from boac.lib.util import get_benchmarker
+from boac.models.manually_added_advisee import ManuallyAddedAdvisee
 from flask import current_app as app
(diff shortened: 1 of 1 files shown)


boac/merged/student.py
```diff
@@ -464,6 +467,7 @@ def search_for_students(
 
 
 def search_for_student_historical(sid):
+    ManuallyAddedAdvisee.find_or_create(sid)
```
paulkerschen: @lyttam This should probably go inside the `if profile` condition below? At this point in the code we don't yet know if it's a real SID or any old string of numbers.

paulkerschen:
Code so simple indicates sensible structures around it! I think one adjustment is advisable before merge.

boac/merged/student.py
lyttam: @paulkerschen I wasn't sure if the profile could be missing for a valid reason - but we definitely don't want any old strings of numbers being written to a table. I'll move it.

lyttam:
So simple it threw me for a loop initially!

PR #1882, opened 2019-09-10
BOAC-2693, StudentWithdrawalCancel has optional prop: termId

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2696

### DIFF
--- a/src/components/student/profile/StudentClasses.vue
+++ b/src/components/student/profile/StudentClasses.vue
@@ -10,10 +10,16 @@
         class="term-no-enrollments">
         <h3 class="student-term-header">{{ currentEnrollmentTerm }}</h3>
         <div class="term-no-enrollments-description">No enrollments</div>
-        <StudentWithdrawalCancel :withdr="student.sisProfile.withdrawalCancel" :term-id="String(currentEnrollmentTermId)" />
+        <StudentWithdrawalCancel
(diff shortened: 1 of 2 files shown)


src/components/student/profile/StudentClasses.vue
```diff
@@ -10,10 +10,16 @@
         class="term-no-enrollments">
         <h3 class="student-term-header">{{ currentEnrollmentTerm }}</h3>
         <div class="term-no-enrollments-description">No enrollments</div>
-        <StudentWithdrawalCancel :withdr="student.sisProfile.withdrawalCancel" :term-id="String(currentEnrollmentTermId)" />
+        <StudentWithdrawalCancel
+          v-if="student.sisProfile.withdrawalCancel"
+          :withdrawal="student.sisProfile.withdrawalCancel"
+          :term-id="currentEnrollmentTermId" />
```
paulkerschen: @johncrossman If we drop the string casting here, will that cause problems with the `withdrawal.termId === termId` check inside the component?

johncrossman: Good catch @paulkerschen  Although I don't think casting the prop is right. I'll put the leniency inside the component.

paulkerschen:
We are no longer out of options.

PR #1883, opened 2019-09-10
BOAC-2695 Don't expand or link courses when full profile is pending

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2695


raydavis:
Don't try to expand 'em,
Just rope 'em, throw, and brand 'em.
Soon we'll be livin' high and wide.

PR #1884, opened 2019-09-11
BOAC-2663, RichTextEditor wraps ckeditor so we can 'correct the DOM', if necessary

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2663

The cause of this bug is explained in [my post on Stack Overflow](https://stackoverflow.com/questions/57877722/how-can-i-get-ckeditor-5-link-dialog-box-to-pin-to-custom-dom-element-instead). Unfortunately, CKEditor config options do offer a path to a solution.

The fix in this PR involves a bit of a hack – see `correctTheDOM()`  – but this is an opportunity to reduce redundant code. For example, fewer lines in the advising-notes-related components.


paulkerschen:
More like wackEditor. Thanks for coming up with this workaround, @johncrossman.

PR #1885, opened 2019-09-11
BOAC-2685, if create template and then immediately create note: avoid dupe files in S3

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2685

This fixes an edge case. The standard workflow, (1) load template and (2) create note(s), has been working: only one file in S3.

### DIFF
--- a/src/components/note/create/CreateNoteModal.vue
+++ b/src/components/note/create/CreateNoteModal.vue
@@ -277,15 +277,15 @@ export default {
         // File upload might take time; alert will be overwritten when API call is done.
         this.showAlert('Creating template...', 60);
       }
-      createNoteTemplate(title, this.model.subject, this.model.body, this.model.topics, this.model.attachments).then(() => {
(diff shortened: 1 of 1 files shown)


src/components/note/create/CreateNoteModal.vue
```diff
@@ -277,15 +277,15 @@ export default {
...
         this.setModel({
           id: undefined,
-          subject: this.model.subject,
-          body: this.model.body,
-          topics: this.model.topics,
-          attachments: this.model.attachments,
+          subject: template.subject,
+          body: template.body,
+          topics: template.topics,
+          attachments: template.attachments,
```
johncrossman: Because `template.attachments` have ids, `/notes/create` API will receive 'templateAttachmentIds' param and reference existing files in S3 rather than uploading new files.

paulkerschen:
Makes sense.

PR #1895, opened 2019-09-17
[QA] BOAC-2745, bulk SIDs error msg: 'SIDs must be separated by...'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2745


paulkerschen:
Treat your SIDs right and you won't believe what they'll do!

PR #1902, opened 2019-09-17
BOAC-2748, db column tool_settings::value type is now 'text'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2748

We are not Twitter. Verbosity is a choice.


paulkerschen:
That guy's always in verbose mode!

PR #1904, opened 2019-09-17
BOAC-2650 Everyone's groups

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2650

Like it says on the tin. Note that the curated group view is modified to accommodate the previously impossible state of viewing a curated group you don't own, with editing functions disabled.

### DIFF
--- a/src/views/AllGroups.vue
+++ b/src/views/AllGroups.vue
@@ -1,0 +1,47 @@
+<template>
+  <div class="m-3">
+    <Spinner />
+    <div v-if="!loading">
+      <h1 ref="pageHeader" class="mb-4">Everyone's Groups</h1>
+
+      <div v-if="!rows.length">
+        <div>There are no saved groups</div>
+      </div>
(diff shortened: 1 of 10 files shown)


src/views/AllGroups.vue
```diff
@@ -0,0 +1,47 @@
+<template>
+  <div class="m-3">
+    <Spinner />
+    <div v-if="!loading">
+      <h1 ref="pageHeader" class="mb-4">Everyone's Groups</h1>
+
+      <div v-if="!rows.length">
+        <div>There are no saved cohorts</div>
```
johncrossman: @paulkerschen Copy-and-paste of "cohorts" here?

paulkerschen: I'm found out!

johncrossman:
Good coders copy; great coders paste.  I'll merge when green.

PR #1908, opened 2019-09-19
[QA] BOAC-2758 Plug one more leak of duplicate students from multiple tables

paulkerschen:
QA-first PR for https://jira.ets.berkeley.edu/jira/browse/BOAC-2758.

BOA code runs this check in a couple of other places, but our code just has so many fun ways to retrieve a set of students.


raydavis:
Makes sense, @paulkerschen -- I'll merge post-Travis.

PR #1918, opened 2019-09-24
BOAC-2755 BOAC-2757 Respect privacy of users without dept affiliations

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2755
https://jira.ets.berkeley.edu/jira/browse/BOAC-2757


raydavis:
The empty set stands alone.

PR #1920, opened 2019-09-26
BOAC-2775, add_user scripts need created_by value

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2775


paulkerschen:
Every chicken has its egg.

PR #1922, opened 2019-09-26
BOAC-2711: Allows user to be blocked from nightly refresh

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2711

### DIFF
--- a/boac/models/authorized_user.py
+++ b/boac/models/authorized_user.py
@@ -24,6 +24,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """
 
 from boac import db, std_commit
+from boac.lib.util import utc_now
 from boac.models.base import Base
 from boac.models.db_relationships import cohort_filter_owners
 from sqlalchemy import text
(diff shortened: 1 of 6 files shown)


boac/models/authorized_user.py
```diff
@@ -71,13 +73,25 @@ def __repr__(self):
...
+        user.is_blocked = True
+        std_commit()
+        return user
+
     @classmethod
     def create_or_restore(cls, uid, created_by, is_admin=False, can_access_canvas_data=True):
         existing_user = cls.query.filter_by(uid=uid).first()
         if existing_user:
+            if existing_user.is_blocked:
+                return existing_user
```
paulkerschen: @lyttam In this path, do we need to null out the user's `deleted_at` and `is_blocked` before returning?

lyttam: @paulkerschen I think not, since we don't want them to be restored or un-blocked.

paulkerschen: @lyttam I see, `is_blocked` is supposed to prevent them from being restored. But I wonder if it would be better to put this check in a higher-up method (possibly in the refresh logic?) It seemed counterintuitive to me that it would be impossible ever to restore a blocked user at the model level.

lyttam: @paulkerschen ah, yes that sounds like a better idea.  If you are thinking of a more self-explanatory name than `is_blocked` I'm more than happy to change that too.

paulkerschen: (Either that, or have the method return `False` in this case, so calling code isn't misled into thinking that it's restored a user when it hasn't.)

lyttam: Maybe I'll do that so we avoid having to do an extra query in `CacheUtils.refresh_department_memberships` just to find out if the user is blocked.

paulkerschen: +1

PR #1924, opened 2019-09-30
BOAC-2620, cohort filter 'GPA' provided as free-form range

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2620

The "Last Name" filter was the only of type 'range'. Now we have two: "Last Name" and GPA. For the sake of reusable UI components and server-side serialization, the cohort-filter code needed a bit of an overhaul. Also, the db script converts `lastNameRange` filter (int array) to `lastNameRanges` (array of `{min, max}` objects), the new standard for range filters.


paulkerschen:
Free-range filters. Thanks for all the lifting and shifting, @johncrossman.

PR #1929, opened 2019-10-01
BOAC-2784, /cohort front-end: value=true when add boolean filter

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2784


paulkerschen:
True.

PR #1931, opened 2019-10-01
BOAC-2785, /cohort page: mo' better validation for add/update range filter

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2785

Tricky logic. Readable code, I hope. QA PR will follow.


lyttam:
Readable enough!

PR #1933, opened 2019-10-01
[QA] BOAC-2788, last_name_ranges SQL requires AND (...OR...)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2788

See Jira desc. Master PR will follow.

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -680,11 +680,10 @@ def get_students_query(     # noqa
     # Generic SIS criteria
     if gpa_ranges:
         sql_ready_gpa_ranges = [f"numrange({gpa_range['min']}, {gpa_range['max']}, '[]')" for gpa_range in gpa_ranges]
-        query_filter += _numranges_to_sql('sas.gpa', sql_ready_gpa_ranges)
-    query_filter += _numranges_to_sql('sas.units', unit_ranges) if unit_ranges else ''
+        query_filter += _number_ranges_to_sql('sas.gpa', sql_ready_gpa_ranges)
+    query_filter += _number_ranges_to_sql('sas.units', unit_ranges) if unit_ranges else ''
     if last_name_ranges:
-        for last_name_range in last_name_ranges:
-            query_filter += _query_filter_last_name_range(last_name_range['min'], last_name_range['max'])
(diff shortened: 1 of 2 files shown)


boac/externals/data_loch.py
```diff
@@ -857,15 +856,37 @@ def numrange_to_sql(column, numrange):
...
+def _last_name_ranges_to_sql(last_name_ranges):
+    query_filter = ''
+    count = len(last_name_ranges)
+    if count:
+        query_filter += ' AND ('
+        for idx, last_name_range in enumerate(last_name_ranges):
+            range_min = last_name_range['min']
+            range_max = last_name_range['max']
+            if range_max == range_min:
+                query_filter += f'(sas.last_name ILIKE \'{range_min}%\')'
```
paulkerschen: @johncrossman Do you think `UPPER(sas.last_name) LIKE` would be more performant here?

paulkerschen: @johncrossman Never mind, I think it wouldn't. In general it might be more performant to keep things like ILIKE and UPPER out of the queries because they might force RDS to do full table scans instead of using its indexes, but that only works if we're normalizing the index name columns beforehand. Something to keep in mind if cohort performance becomes a problenm, but probably not worth worrying about now.

tests/test_api/test_cohort_controller.py
```diff
@@ -696,31 +696,34 @@ def test_students_per_filters_unauthorized(self, client, asc_advisor_login):
...
                 ],
+                'orderBy': 'last_name',
             },
         )
-        assert 'totalStudentCount' in api_json
-        assert 'students' in api_json
+        students = api_json['students']
+        assert len(students) == 5
+        assert api_json.get('totalStudentCount') == 5
+        assert ['Barney', 'Doolittle', 'Farestveit', 'Kerschen', 'Schlemiel'] == [s['lastName'] for s in students]
```
paulkerschen: The fab five.

boac/externals/data_loch.py
johncrossman: Thanks for the scrutiny. I'll keep this point in mind.

PR #1934, opened 2019-10-01
BOAC-2786, margin and font weight/color fix on student profile

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2786

A very quick and satisfying fix.


paulkerschen:
I'm satisfied.

PR #1945, opened 2019-10-04
BOAC-2711: Revamps admin page

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2844

The new table allows filtering by department, but that doesn't include not-real departments like Admin and Notes Only.  I'll submit another PR to expand filtering capabilities to include permissions and status.

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -108,11 +108,17 @@ def delete_university_dept_membership(university_dept_id, authorized_user_id):
     )
 
 
-@app.route('/api/users/authorized_groups')
+@app.route('/api/users/all')
 @admin_required
-def authorized_user_groups():
+def all_users():
     sort_users_by = util.get(request.args, 'sortUsersBy', None)
-    return tolerant_jsonify(_get_boa_user_groups(sort_users_by))
+    return tolerant_jsonify(_get_boa_users(sort_users_by))
+
+
+@app.route('/api/users/departments')
+@admin_required
+def departments():
+    return tolerant_jsonify(_get_boa_departments())
 
 
 @app.route('/api/user/demo_mode', methods=['POST'])
@@ -152,6 +158,16 @@ def download_boa_users_csv():
         filename_prefix='boa_users',
         fieldnames=['last_name', 'first_name', 'uid', 'email', 'dept_code', 'dept_name'],
     )
+
+
+def _get_boa_users(sort_user_by=None):
+    users = AuthorizedUser.get_all_users()
+    return authorized_users_api_feed(users, sort_user_by)
+
+
+def _get_boa_departments():
+    departments = [{'name': BERKELEY_DEPT_CODE_TO_NAME[dept_code], 'code': dept_code} for dept_code in BERKELEY_DEPT_CODE_TO_NAME.keys()]
+    return sorted(departments, key=lambda dept: dept['name'])
 
 
 def _get_boa_user_groups(sort_users_by=None):
--- a/boac/api/util.py
+++ b/boac/api/util.py
@@ -35,6 +35,7 @@ from boac.merged import calnet
 from boac.merged.advising_note import get_advising_notes
 from boac.models.alert import Alert
 from boac.models.curated_group import CuratedGroup
+from dateutil.tz import tzutc
 from flask import current_app as app, request
 from flask_login import current_user
 
@@ -79,13 +80,18 @@ def authorized_users_api_feed(users, sort_by='lastName'):
         profile.update({
             'id': user.id,
             'isAdmin': user.is_admin,
+            'isBlocked': user.is_blocked,
+            'canAccessCanvasData': user.can_access_canvas_data,
+            'deletedAt': _isoformat(user.deleted_at),
             'departments': {},
         })
         for m in user.department_memberships:
             profile['departments'].update({
                 m.university_dept.dept_code: {
+                    'deptName': m.university_dept.dept_name,
                     'isAdvisor': m.is_advisor,
                     'isDirector': m.is_director,
+                    'automateMembership': m.automate_membership,
                 },
             })
         profiles.append(profile)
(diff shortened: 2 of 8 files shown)


lyttam:
image

boac/api/user_controller.py
```diff
@@ -108,13 +108,26 @@ def delete_university_dept_membership(university_dept_id, authorized_user_id):
...
+def authorized_users():
+    sort_users_by = util.get(request.args, 'sortUsersBy', None)
+    return tolerant_jsonify(_get_boa_users(sort_users_by))
+
+
 @app.route('/api/users/authorized_groups')
 @admin_required
 def authorized_user_groups():
     sort_users_by = util.get(request.args, 'sortUsersBy', None)
     return tolerant_jsonify(_get_boa_user_groups(sort_users_by))
```
johncrossman: @lyttam We can remove `/api/users/authorized_groups` endpoint, yes?

boac/api/user_controller.py
```diff
@@ -108,13 +108,26 @@ def delete_university_dept_membership(university_dept_id, authorized_user_id):
     )
 
 
+@app.route('/api/users/authorized')
+@admin_required
+def authorized_users():
+    sort_users_by = util.get(request.args, 'sortUsersBy', None)
+    return tolerant_jsonify(_get_boa_users(sort_users_by))
```
johncrossman: @lyttam - No need to show deleted users on new Admin page? (I don't remember the requirement.)

src/store/modules/user.ts
```diff
@@ -1,18 +1,20 @@
...
 
 const state = {
   calnetUsersByCsid: {},
+  departments: undefined,
   preferences: {
     sortBy: 'last_name'
   },
   user: undefined,
-  userGroups: undefined
+  userGroups: undefined,
```
johncrossman: @lyttam - Can we get rid of `userGroups` (above) and `setUserGroups` (below)?

johncrossman:
Looks really good @lyttam  Thank you!  I posted a few questions. I have one request: Would you mind adding a line of documentation in `authorized_user.py` to explain the `is_blocked` property? Its purpose would not be obvious if I were seeing it for the first time.

boac/api/user_controller.py
lyttam: @johncrossman We will need to show them actually - I'll make that change.

src/store/modules/user.ts
lyttam: Yes, I'll remove it.

boac/api/user_controller.py
lyttam: @johncrossman Yes you're right, I'll remove it.

lyttam:
@johncrossman I've addressed your points; this is ready for a second look.

boac/lib/http.py
```diff
@@ -140,3 +140,8 @@ def response_with_csv_download(rows, filename_prefix, fieldnames=None):
     csv_writer.writeheader()
     csv_writer.writerows(rows)
     return response
+
+
+def _stringify_date(o):
+    if isinstance(o, datetime):
+        return o.__str__()
```
johncrossman: @lyttam What ends up in resulting JSON if `not isinstance(o, datetime)`?

lyttam: @johncrossman If `deleted_at` is populated with a `datetime`, I get a `TypeError: datetime.datetime is not JSON serializable`

johncrossman: @lyttam - Understood. The convention elsewhere (e.g, `advising_note.note_to_compatible_json()`)  has been `isoformat()` to stringify. Can we add that in `authorized_users_api_feed()`?  I'm nervous about overriding the default serializer.

src/components/admin/Users.vue
```diff
@@ -0,0 +1,244 @@
...
+      const key = sortBy === 'name' ? 'lastName' : sortBy
+      let aValue = this.get(a, key);
+      let bValue = this.get(b, key);
+      return this.sortComparator(aValue, bValue);
+    }
+  }
+}
+</script>
+
+<style>
```
johncrossman: @lyttam Should these styles be "scoped"?

src/components/admin/Users.vue
```diff
@@ -0,0 +1,244 @@
...
+    filterFields: ['depts'],
+    items: [],
+    perPage: 10,
+    rowCount: 0,
+    sortBy: 'lastName',
+    sortDesc: false
+  }),
+  created() {
+    this.items = this.cloneDeep(this.users);
+    this.rowCount = this.users ? this.users.length : 0;
```
johncrossman: @lyttam - A minor concern: `this.size(this.users)" is alternative to line above.

src/components/admin/Users.vue
lyttam: @johncrossman Yes, except then the `class` attributes in the `fields` definition will be ignored.  Seems like a quirk to me but it is in the [documentation](https://bootstrap-vue.js.org/docs/components/table#field-definition-reference):  "class, thClass, tdClass etc. will not work with classes that are defined in scoped CSS, unless you are using VueLoader's Deep selector."

I could scope everything but those classes, if it seems worth it.

johncrossman: Ah, thanks for clarifying. It's fine as is.

boac/api/util.py
```diff
@@ -79,13 +80,18 @@ def authorized_users_api_feed(users, sort_by='lastName'):
         profile.update({
             'id': user.id,
             'isAdmin': user.is_admin,
+            'isBlocked': user.is_blocked,
+            'canAccessCanvasData': user.can_access_canvas_data,
+            'deletedAt': _isoformat(user.deleted_at),
```
lyttam: @johncrossman I moved the stringification of the `datetime` to `authorized_users_api_feed()` as you suggested.

johncrossman:
Thanks @lyttam  Waiting on Travis

PR #1947, opened 2019-10-08
BOAC-2714: Displays user's dept membership data on admin page

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2714


lyttam:
image

PR #1951, opened 2019-10-09
BOAC-2805, correctly identify advisors and schedulers for proper routing

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2805


paulkerschen:
A Field Guide to Northern California Advisors and Schedulers.

PR #1954, opened 2019-10-09
BOAC-2849, '/appt/desk' view gets service alerts and other shared components

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2849

Includes `vue-router` version bump: 3.0.7 -> 3.1.3


paulkerschen:
BOA don't tolerate old Vue!

PR #1957, opened 2019-10-09
BOAC-2832, stub 'appointment' API endpoints as prep for /student profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2832

A little prep before real work begins.


paulkerschen:
Raise high the scaffolding, carpenter.

PR #1959, opened 2019-10-10
BOAC-2860, Vue discourages router.push if target path matches current path

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2860

The recent upgrade of `vue-router` brought in its new promise API. See https://github.com/vuejs/vue-router/issues/2872  The fix is a noop callback.


paulkerschen:
Whatever you say, `vue-router`.

PR #1961, opened 2019-10-10
BOAC-2737, persistent and unobtrusive warning, ''This BOA ... for testing only'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2737

I opted for no {{isProduction}} trickery in _config_controller_. (The new-batch-note button gets extra padding if fixed-warning-in-footer is present.)

I also alphabetized configs in `default.py` because it was a mess. Feel free to diff against the existing file.


config/qa.py
```diff
@@ -0,0 +1,26 @@
...
+OF THE POSSIBILITY OF SUCH DAMAGE.
+
+REGENTS SPECIFICALLY DISCLAIMS ANY WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE. THE
+SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
+"AS IS". REGENTS HAS NO OBLIGATION TO PROVIDE MAINTENANCE, SUPPORT, UPDATES,
+ENHANCEMENTS, OR MODIFICATIONS.
+"""
+
+FIXED_WARNING_ON_ALL_PAGES = 'This BOA instance is for testing only, and will be periodically refreshed.'
```
johncrossman: IMO, the redundancy of this config is acceptable.

paulkerschen:
@johncrossman I think our environment-naming conventions might be causing confusion. "development," "demo" and "production" are the names of different modes that BOA can/could be run in (though I don't think we use "demo" any more) - but unless I'm remembering wrong, all of our BOA EB environments run in production and configure off production.py and production-local.py. 

So I think the place to set these configurations is not in development.py, demo.py, etc., but as release instructions to modify the S3 local configs for non-prod environments.

johncrossman:
I agree @paulkerschen  I'll amend.

lyttam:
Sorry @johncrossman, I merged too soon.

PR #1965, opened 2019-10-10
BOAC-2747 Everyone sees the same dept-specific inactive labels

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2747

Removing the tiniest bit of complexity.


johncrossman:
Looks good to me but Travis is in the red.

paulkerschen:
Travis is solvent now.

PR #1967, opened 2019-10-11
BOAC-2832, AdvisingAppointment component covers basics of display and mark-as-read

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2832


paulkerschen:
They're notes, and then again, they're not notes at all...

PR #1970, opened 2019-10-11
BOAC-2811: Adds support for appointments

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2811

I removed `scheduler_uid` from the `appointment_topics` table since topics cannot be added to an appointment after it's been created (`appointments.created_by` will tell us who added the topics).


lyttam:
I neglected schema.sql and drop_schema.sql, so I'll be amending this.

boac/models/appointment.py
```diff
@@ -0,0 +1,173 @@
...
+    details = db.Column(db.Text, nullable=False)
+    created_by = db.Column(db.String(255), nullable=False)
+    checked_in_at = db.Column(db.DateTime, nullable=True)
+    checked_in_by = db.Column(db.String(255), nullable=True)
+    cancelled_at = db.Column(db.DateTime, nullable=True)
+    cancelled_by = db.Column(db.String(255), nullable=True)
+    updated_by = db.Column(db.String(255), nullable=False)
+    deleted_at = db.Column(db.DateTime, nullable=True)
+    deleted_by = db.Column(db.String(255), nullable=True)
+    topics = db.relationship(
```
johncrossman: @lyttam A few questions/comments on the above:
* Do reqts require that we capture `student_name`?
* In `notes` table, the `body` is nullable. Are appointments different in that `details` are always required?
* What about `updated_at`?
* With so many columns, I think we should alphabetize the items above.

boac/models/appointment.py
```diff
@@ -0,0 +1,173 @@
...
+        for topic in topics:
+            appointment.topics.append(
+                AppointmentTopic.create(appointment, topic),
+            )
+        db.session.add(appointment)
+        std_commit()
+        return appointment
+
+    @classmethod
+    def check_in(cls, appointment_id, checked_in_by):
```
johncrossman: @lyttam - I believe `check_in()` needs advisor_uid, advisor_name, advisor_role, and advisor_dept_codes.

boac/models/appointment.py
```diff
@@ -0,0 +1,173 @@
...
+            db.session.refresh(appointment)
+            return appointment
+        else:
+            return None
+
+    @classmethod
+    def cancel(cls, appointment_id, cancelled_by):
+        appointment = cls.find_by_id(appointment_id=appointment_id)
+        if appointment:
+            appointment.status = 'Cancelled'
```
johncrossman: @lyttam - I prefer the double-L spelling but I understand the BOA convention to be `Canceled`.

boac/models/appointment.py
lyttam: @johncrossman 
- We need to display student name, but not necessarily store it on the appointment. Would you prefer querying Redshift for the student name?
- Good question, I think it should be nullable.
- `updated_at` is included via the parent class `Base`.
- I will alphabetize.

boac/models/appointment.py
```diff
@@ -0,0 +1,173 @@
...
+    def to_api_json(self):
+        topics = [t.to_api_json() for t in self.topics if not t.deleted_at]
+        return {
+            'id': self.id,
+            'advisorUid': self.advisor_uid,
+            'advisorName': self.advisor_name,
+            'advisorRole': self.advisor_role,
+            'advisorDeptCodes': self.advisor_dept_codes,
+            'student_sid': self.student_sid,
+            'student_name': self.student_name,
```
johncrossman: @lyttam - camelCase needed above

boac/models/appointment.py
lyttam: I will add those.

boac/models/appointment.py
lyttam: Thanks, will fix.

scripts/db/migrate/20191009-BOAC-2809/pre_deploy_03_create_appointments_table.sql
```diff
@@ -0,0 +1,32 @@
...
+  created_at TIMESTAMP WITH TIME ZONE NOT NULL,
+  created_by VARCHAR(255) NOT NULL,
+  checked_in_at TIMESTAMP WITH TIME ZONE NOT NULL,
+  checked_in_by VARCHAR(255) NOT NULL,
+  cancelled_at TIMESTAMP WITH TIME ZONE NOT NULL,
+  cancelled_by VARCHAR(255) NOT NULL,
+  updated_at TIMESTAMP WITH TIME ZONE NOT NULL,
+  updated_by VARCHAR(255) NOT NULL,
+  deleted_at TIMESTAMP WITH TIME ZONE,
+  deleted_by VARCHAR(255) NOT NULL
```
johncrossman: @lyttam - Same column changes needed here. I suppose `cancelled_at ` and `cancelled_by` should be changed to `canceled_at` and `canceled_by`.

boac/models/appointment.py
johncrossman: @lyttam When creating a note, we only save the student's `sid`.  We can do the same here; `student_name` is not needed.

johncrossman:
@lyttam - Based on comments in https://jira.ets.berkeley.edu/jira/browse/BOAC-2751, and follow-up discussion with Steven, I think we need two more columns:
* `cancel_reason`
* `cancel_reason_explained`

In the UI, `cancel_reason` will be a (required) dropdown with 2-3 options and `cancel_reason_explained` will be an optional textarea.

PR #1972, opened 2019-10-11
BOAC-2836, only user.isDropInAdvisor can check-in/cancel appointments on student profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2836

Also, both `isScheduler` and `isDropInAdvisor` can access the Intake Desk view.

### DIFF
--- a/src/components/appointment/AppointmentDetailsModal.vue
+++ b/src/components/appointment/AppointmentDetailsModal.vue
@@ -44,18 +44,19 @@
       <div class="modal-footer">
         <form @submit.prevent="functionConfirm">
           <b-btn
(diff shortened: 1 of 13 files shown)


src/components/appointment/AppointmentDetailsModal.vue
```diff
@@ -44,18 +44,19 @@
       <div class="modal-footer">
         <form @submit.prevent="functionConfirm">
           <b-btn
+            v-if="!appointment.status"
```
lyttam: @johncrossman I wound up removing the `status` column from the appointments table.

PR #1977, opened 2019-10-14
BOAC-2737, id on dismiss-warning button; on dismiss, alert screen-reader

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2737

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -16,10 +16,11 @@
         </div>
         <div class="btn-wrapper ml-0 align-top">
           <b-btn
+            id="speedbird"
             class="btn-dismiss pl-2 pt-0 text-white"
             variant="link"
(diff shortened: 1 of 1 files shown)


src/App.vue
```diff
@@ -16,10 +16,11 @@
         </div>
         <div class="btn-wrapper ml-0 align-top">
           <b-btn
+            id="dismiss-warning-in-footer"
```
pfarestveit: @johncrossman, the QA Dept rejects this ID as dull.  In keeping with the BOAC theme, I'd expect something like id="speedbird".

johncrossman: image

Done.

pfarestveit:
Living the High Life @johncrossman !

PR #1978, opened 2019-10-14
BOAC-2872, URL of Intake Dashboard (appt waitlist) must include dept_code

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2872

See description in Jira above.


paulkerschen:
Our stable of test users grows and grows.

PR #1985, opened 2019-10-15
BOAC-2866, cancel appointment modal used on student profile and /appt/desk

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2866

### DIFF
--- a/boac/api/appointments_controller.py
+++ b/boac/api/appointments_controller.py
@@ -77,6 +77,25 @@ def appointment_check_in(appointment_id):
         raise ForbiddenRequestError(f'You are unauthorized to manage {appointment.dept_code} appointments.')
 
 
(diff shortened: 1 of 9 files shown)


boac/models/appointment.py
```diff
@@ -99,6 +99,8 @@ def get_waitlist(cls, dept_code):
                 cls.canceled_at == None,
                 cls.checked_in_at == None,
                 cls.deleted_at == None,
+                cls.canceled_at == None,
+                cls.checked_in_at == None,
```
paulkerschen: @johncrossman These two added fields look like duplicates - were they meant to be something else?

paulkerschen:
Looks good from here - left one question.

boac/models/appointment.py
johncrossman: Ah yes, I'll amend.

PR #1986, opened 2019-10-15
BOAC-2831: Adds appointment search index

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2831

### DIFF
--- a/boac/models/appointment.py
+++ b/boac/models/appointment.py
@@ -32,6 +32,7 @@ from boac.models.base import Base
 from dateutil.tz import tzutc
 from sqlalchemy import and_
 from sqlalchemy.dialects.postgresql import ARRAY
+from sqlalchemy.sql import text
 
 
 class Appointment(Base):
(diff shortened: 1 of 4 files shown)


boac/models/appointment.py
```diff
@@ -133,6 +134,7 @@ def create(
             )
         db.session.add(appointment)
         std_commit()
+        cls.refresh_search_index()
```
johncrossman: @lyttam Are canceled appointments searchable? If not, then we'll need refresh when `cancel()`

lyttam: @johncrossman I think we'd only need to refresh if `cancel()` changes the details of the appointment.  Cancelled appointments will be searchable unless we add `WHERE canceled_at IS NULL` to the search query.

lyttam: I just realized we should probably index `cancel_reason` and `cancel_reason_explained`, in which case we will need to refresh the index on cancel.

johncrossman: @lyttam Would you mind creating a Jira with the details above and question, "what is searchable?" In the meantime, this PR is ready to merge.

PR #1989, opened 2019-10-15
BOAC-2858: Hides become button and fixes filtering for role-less users

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2858

### DIFF
--- a/src/components/admin/Users.vue
+++ b/src/components/admin/Users.vue
@@ -177,7 +177,7 @@
             <span class="sr-only">Edit user</span>
           </b-btn>
           <b-btn
-            v-if="devAuthEnabled && (row.item.uid !== user.uid) && !row.item.isExpiredPerLdap"
+            v-if="canBecome(row.item)"
             :id="'become-' + row.item.uid"
             :title="`Log in as ${row.item.name}`"
             variant="link"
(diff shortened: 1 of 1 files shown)


src/components/admin/Users.vue
```diff
@@ -274,16 +274,23 @@ export default {
...
 
       return nameUidMatch && deptMatch && permissionsMatch && statusMatch;
     },
     become(uid) {
       becomeUser(uid).then(() => (window.location.href = '/home'));
     },
+    canBecome(user) {
+      const isNotMe = user.uid !== this.user.uid;
+      const expiredOrInactive = user.isExpiredPerLdap || user.deletedAt || user.isBlocked;
+      const hasAnyRole = find(user.departments, (dept) => dept.isAdvisor || dept.isDirector || dept.isDropInAdvisor || dept.isScheduler);
```
johncrossman: @lyttam - Would it make sense to replace the line above with the equivalent `user.isActive` (which is computed server-side, in `user_session`).

lyttam: @johncrossman It would if there was an `isActive` property on each user in the list, but it looks like it only exists on the currently logged-in user.  I could add a similar property to each `AuthorizedUser` returned by `/api/users/all`.

johncrossman: @lyttam  Makes sense. Although the filters include an "Active" checkbox (which gives zero results). Does that need to be wired up to similar logic?

lyttam: @johncrossman the "Active" filter shows users who are not deleted or blocked... though the filters in general may need some rework to clarify their behavior/purpose (see BOAC-2874).

johncrossman:
Thanks for clarification, @lyttam

PR #1991, opened 2019-10-16
BOAC-2822, drop-in advisor homepage shares DropInWaitlist component with /appt/desk

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2822

Not much new code – mostly reshuffling.


paulkerschen:
Drop that beat.

PR #1993, opened 2019-10-16
BOAC-2801 Do not show obsolete waitlists

raydavis:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2801


paulkerschen:
- Let's go.
- We can't.
- Why not?

PR #1994, opened 2019-10-16
BOAC-2816 Appointment checkin from waitlist

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2816

This code establishes the happy path from the scheduler's point of view.

### DIFF
--- a/boac/api/user_controller.py
+++ b/boac/api/user_controller.py
@@ -24,7 +24,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 """
 
 from boac.api import errors
-from boac.api.util import admin_required, advisor_required, authorized_users_api_feed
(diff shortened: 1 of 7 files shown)


johncrossman:
Who doesn't like a happy path?!  Thanks @paulkerschen

src/components/appointment/AppointmentDetailsModal.vue
```diff
@@ -52,6 +52,13 @@
         </div>
       </div>
       <div class="modal-footer">
+        <b-btn
+          id="btn-appointment-cancel"
```
lyttam: @paulkerschen I know I'm late to the party, but you have a duplicate ID here (see line 63).

PR #1998, opened 2019-10-17
BOAC-2891, split BOA /admin into (1) Flight Desk and (2) Passenger Manifest

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2891

image


paulkerschen:
All aboard!

PR #1999, opened 2019-10-17
BOAC-2836 Appointment advisor and type in academic timeline display

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2836

Includes back-end support for appointment type field.


johncrossman:
Collect 'em all!

PR #2004, opened 2019-10-18
BOAC-2843, select multiple topics for new appt; render 'Select...' instead of blank

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2843

`v-model` is used solely to reset the dropdown.


paulkerschen:
These topics are hot!

PR #2008, opened 2019-10-18
BOAC-2830 BOAC-2881 Styling, ids and such

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2830
https://jira.ets.berkeley.edu/jira/browse/BOAC-2881


johncrossman:
I think I like the "such" most of all.

PR #2009, opened 2019-10-18
BOAC-2898, consistent styling of sortable-Alerts; responsive /home/dept_code

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2898


paulkerschen:
They're sortable, they're portable, they're BOA alerts.

PR #2011, opened 2019-10-19
BOAC-2808, every beforeEnter() in router must invoke a next()

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2808

I also removed unnecessary `store.dispatch('user/loadUser')`


paulkerschen:
Every `beforeEnter()` is sacred.

PR #2016, opened 2019-10-22
BOAC-2899, Create appt: you get a do-over if you choose the wrong student

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2899

Review with `?w=1`. Includes ids and styles.


paulkerschen:
Now I can choose with confidence!

PR #2018, opened 2019-10-22
BOAC-2911, avoid permalink collisions on student profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2911

Also, resolves https://jira.ets.berkeley.edu/jira/browse/BOAC-2910

@paulkerschen - You introduced that regex. Do you know if this change will impact beyond note and appointment links in search results?

### DIFF
--- a/src/components/student/profile/AcademicTimeline.vue
+++ b/src/components/student/profile/AcademicTimeline.vue
@@ -116,7 +116,7 @@
         </tr>
         <tr
           v-for="(message, index) in (searchResults ? filterSearchResults() : (isShowingAll ? messagesPerType(filter) : slice(messagesPerType(filter), 0, defaultShowPerTab)))"
-          :id="`message-row-${message.id}`"
+          :id="`permalink-${message.type}-${message.id}`"
           :key="index"
           class="message-row border-top border-bottom"
           :class="{ 'message-row-read': message.read }"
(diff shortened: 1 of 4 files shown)


src/components/student/profile/AcademicTimeline.vue
```diff
@@ -422,20 +422,21 @@ export default {
   },
   mounted() {
     if (this.anchor) {
-      const match = this.anchor.match(/#(\d[0-9A-Z-]+\d)/);
-      if (match) {
-        const messageId = match[1];
-        const note = this.find(this.messages, function(m) {
+      const match = this.anchor.match(/^#(\w+)-(\d+)/);
```
paulkerschen: @johncrossman I think the legacy note ids already have hyphens in them, so this new permalink format might yield a fragment of the form `note-1234567-00001`, which I'm not sure this regex would catch.

paulkerschen: Plus the ASC notes sometimes put letters in their note ids. You might need `this.anchor.match(/^#(\w+)-(\d[0-9A-Z-]+\d)/);` to catch all cases.

johncrossman: @paulkerschen  Thank you. Forgot about those. But this latest regex would not catch one or two digit ids, right?

johncrossman: I think I can tweak it right.

paulkerschen: You're right, it wouldn't. Trust you to fix it.

johncrossman:
Regex amended. cc: @paulkerschen

PR #2021, opened 2019-10-22
BOAC-2879, spinner component optional props for srAlert: prefix, is-plural

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2879

Meaningful `srAlert` when page is loading / has loaded.


paulkerschen:
I load, thou loadest, it loadeth.

PR #2028, opened 2019-10-23
BOAC-2913 Appointment status indicator on academic timeline

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2913


johncrossman:
Got a status? There's a pill for it.

PR #2034, opened 2019-10-24
BOAC-2841: Adapts sidebar search form for appointments

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2841
https://jira.ets.berkeley.edu/jira/browse/BOAC-2842

### DIFF
--- a/boac/api/appointments_controller.py
+++ b/boac/api/appointments_controller.py
@@ -137,10 +137,33 @@ def mark_appointment_read(appointment_id):
     return tolerant_jsonify(AppointmentRead.find_or_create(current_user.get_id(), int(appointment_id)).to_api_json())
 
 
-@app.route('/api/appointments/topics')
+@app.route('/api/appointments/advisors/find_by_name', methods=['GET'])
+@advisor_required
+def find_appointment_advisors_by_name():
+    if not app.config['FEATURE_FLAG_ADVISOR_APPOINTMENTS']:
+        raise ResourceNotFoundError('Unknown path')
+    query = request.args.get('q')
+    if not query:
+        raise BadRequestError('Search query must be supplied')
+    limit = request.args.get('limit')
+    query_fragments = filter(None, query.upper().split(' '))
+    advisors = Appointment.find_advisors_by_name(query_fragments, limit=limit)
+
+    def _advisor_feed(a):
+        return {
+            'label': a.advisor_name,
+            'uid': a.advisor_uid,
+        }
+    return tolerant_jsonify([_advisor_feed(a) for a in advisors])
+
+
+@app.route('/api/appointments/topics', methods=['GET'])
 @scheduler_required
 def get_appointment_topics():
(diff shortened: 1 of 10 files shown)


boac/models/appointment.py
```diff
@@ -102,6 +100,25 @@ def __init__(
     def find_by_id(cls, appointment_id):
         return cls.query.filter(and_(cls.id == appointment_id, cls.deleted_at == None)).first()  # noqa: E711
 
+    @classmethod
+    def find_advisors_by_name(cls, prefixes, limit=None):
```
lyttam: Because the `appointments` table has a combined `name` column and we don't have a separate `appointment_advisor_names` table, name search is currently limited to first name only or first + last name.

boac/api/appointments_controller.py
```diff
@@ -137,10 +137,31 @@ def mark_appointment_read(appointment_id):
     return tolerant_jsonify(AppointmentRead.find_or_create(current_user.get_id(), int(appointment_id)).to_api_json())
 
 
-@app.route('/api/appointments/topics')
+@app.route('/api/appointments/advisors/find_by_name', methods=['GET'])
+@advisor_required
+def find_appointment_advisors_by_name():
+    query = request.args.get('q')
```
johncrossman: @lyttam - If it makes sense to you, and it keeps client-side code cleaner, then we should return an empty array if feature-flag if false.

boac/merged/advising_note.py
```diff
@@ -177,10 +178,11 @@ def search_advising_notes(
     else:
         search_terms = []
 
+    author_uid = get_uid_for_csid(app, author_csid) if (not author_uid and author_csid) else author_uid
```
johncrossman: @lyttam - Does it make sense to move this line outside the method and then get rid `author_csid` from list of method args?

src/views/Search.vue
```diff
@@ -129,6 +130,7 @@ export default {
     loadingAdditionalNotes: undefined,
     noteAndAppointmentOptions: {
       advisorCsid: undefined,
+      advisorUid: undefined,
```
johncrossman: @lyttam - Is it possible to pass along a single advisor identifier or are we forced to track both?

boac/merged/advising_note.py
lyttam: @johncrossman `author_csid` is still needed in order to find SIS notes (`author_uid`) is not populated).  ASC notes have the opposite problem - `author_uid` is populated but `author_csid` is not.

src/views/Search.vue
lyttam: @johncrossman we still need both, unfortunately.

boac/api/appointments_controller.py
lyttam: @johncrossman  I updated the method to return a 404 if feature flag is false, and the client logic prevents the API from being called in that case.

johncrossman:
Thank you @lyttam

PR #2035, opened 2019-10-24
BOAC-2940 Arrow functions and 'this', the old gotcha

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2940


lyttam:
Don't get got.

PR #2039, opened 2019-10-24
BOAC-2942, click check-in and the modal must close

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2942


paulkerschen:
Avast.

PR #2040, opened 2019-10-24
BOAC-2874: Simplifies passenger manifest checkbox logic

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2874
https://jira.ets.berkeley.edu/jira/browse/BOAC-2882

I think it makes sense for the checkboxes to work like an inclusive OR. Thoughts?

### DIFF
--- a/src/components/admin/Users.vue
+++ b/src/components/admin/Users.vue
@@ -287,7 +287,7 @@ export default {
       const permissionsMatch = adminMatch || advisorMatch || canvasAccessMatch || directorMatch || dropInAdvisorMatch || schedulerMatch || noPermissionsMatch;
       const statusMatch = activeMatch || deletedMatch || blockedMatch || expiredMatch;
 
-      return nameUidMatch && deptMatch && permissionsMatch && statusMatch;
+      return nameUidMatch && deptMatch && (permissionsMatch || statusMatch);
     },
     become(uid) {
       becomeUser(uid).then(() => (window.location.href = '/home'));


lyttam:
From Steven's comments it sounds like his main use case will be searching for a particular user in order to edit them - in which case we don't want the checkboxes to prevent that user from showing up in a UID or name search.

src/components/admin/Users.vue
```diff
@@ -287,7 +287,7 @@ export default {
       const permissionsMatch = adminMatch || advisorMatch || canvasAccessMatch || directorMatch || dropInAdvisorMatch || schedulerMatch || noPermissionsMatch;
       const statusMatch = activeMatch || deletedMatch || blockedMatch || expiredMatch;
 
-      return nameUidMatch && deptMatch && permissionsMatch && statusMatch;
+      return nameUidMatch && deptMatch && (permissionsMatch || statusMatch);
```
johncrossman: @lyttam - IMO, this is less intuitive. As a dev who uses the feature to find test-users, the feature is good enough in its current state. I think another opinion would be useful.  cc: @paulkerschen

paulkerschen: If I'm reading this right, replacing AND with OR means that we couldn't use the checkboxes in combination to say "Show me all active advisors" or "Show me all active admins," which is the kind of thing I often do in development. There's some ugliness no matter how we tackle it, but I'd also vote for keeping the current arrangement as good enough.

lyttam: That's how I would use this feature too (e.g. to find test users).  I'll close this until someone has a better idea.

PR #2047, opened 2019-10-25
NOJIRA, localhost only: screenReaderAlert is the dismissible message in footer

johncrossman:
This will only surface on developer workstations. The goal is to raise awareness of screenreader alerts and, hopefully, improve 'em.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -12,15 +12,15 @@
           <b>BOA {{ getBoaEnvLabel() }} Environment</b>
         </div>
         <div>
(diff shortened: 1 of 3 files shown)


paulkerschen:
This is a good idea.

johncrossman:
Thanks @paulkerschen  I now see that it surfaces in boa-dev, despite my reassurance of "it will not".  I'm undecided on whether that's a bad thing. But, I'm fairly certain it will surface in neither `boa-qa` nor `boa-demo`. We can discuss...

PR #2055, opened 2019-10-28
BOAC-2895 Script and cronjob for nightly appointments cleanup

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2895

The idea is that this gets run nightly at 10:00 UTC. Will it work?


johncrossman:
It might work!

PR #2056, opened 2019-10-28
BOAC-2947: Fixes advisor autocomplete browser compatibility issue

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2947


johncrossman:
Promises, promises.

PR #2058, opened 2019-10-28
BOAC-2894: Fixes sql bug in appointments search index

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2894

### DIFF
--- a/scripts/db/migrate/20191009-BOAC-2809/pre_deploy_05_create_appointments_search_index.sql
+++ b/scripts/db/migrate/20191009-BOAC-2809/pre_deploy_05_create_appointments_search_index.sql
@@ -3,8 +3,9 @@ BEGIN;
 DROP MATERIALIZED VIEW IF EXISTS appointments_fts_index;
 
 CREATE MATERIALIZED VIEW appointments_fts_index AS (
-  SELECT 
-    id, to_tsvector('english', trim(concat(details, ' ', cancel_reason, ' ', cancel_reason_explained))) AS fts_index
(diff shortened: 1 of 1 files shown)


scripts/db/migrate/20191009-BOAC-2809/pre_deploy_05_create_appointments_search_index.sql
```diff
@@ -3,8 +3,9 @@ BEGIN;
 DROP MATERIALIZED VIEW IF EXISTS appointments_fts_index;
 
 CREATE MATERIALIZED VIEW appointments_fts_index AS (
-  SELECT 
-    id, to_tsvector('english', trim(concat(details, ' ', cancel_reason, ' ', cancel_reason_explained))) AS fts_index
+  SELECT
+    id,
+    to_tsvector('english', details || ' ' || coalesce(cancel_reason, '') || ' ' || coalesce(cancel_reason_explained, '')) AS fts_index
```
johncrossman: @lyttam If `cancel_reason` is null, what happens with `coalesce()` above?  I would expect `coalesce(cancel_reason || '', '')`

lyttam: @johncrossman `coalesce` returns the first non-null argument, so in that case `coalesce(cancel_reason, '')` will return `''`.

johncrossman:
Thanks @lyttam

PR #2062, opened 2019-10-29
BOAC-2973, canceled appointments to bottom of waitlist

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2973

Renamed `src/views/AppointmentWaitlist` to `DropInDesk` to get it side-by-side, in alpha order, with our other drop-in view.


paulkerschen:
Modular desks are always in style.

PR #2063, opened 2019-10-30
BOAC-2946: Adds SR alerts for note/appt search filter toggle

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2946

### DIFF
--- a/src/store/modules/context.ts
+++ b/src/store/modules/context.ts
@@ -68,6 +68,7 @@ const mutations = {
 
 const actions = {
   alertScreenReader: ({ commit }, alert) => {
+    commit('setScreenReaderAlert', '');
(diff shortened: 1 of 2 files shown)


src/store/modules/context.ts
```diff
@@ -68,6 +68,7 @@ const mutations = {
 
 const actions = {
   alertScreenReader: ({ commit }, alert) => {
+    commit('setScreenReaderAlert', '');
```
lyttam: If the screen reader alert text is not cleared each time before updating it, NVDA will voice only the text that changed - right down to the individual letters. This results in nonsensical alerts.

paulkerschen: That's a good tip!

PR #2065, opened 2019-10-30
BOAC-2971 Shared appointments polling for scheduler and drop-in advisor

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2971

A shared DropInWaitlistContainer mixin is silly, but the alternative I was attempting (making DropInWaitlist responsible for its own data updates) really didn't play well with the nested components model.


johncrossman:
The mixin mixup. Works for me! Thx @paulkerschen

PR #2068, opened 2019-11-01
BOAC-2993, refactor appointments-related schema and necessary model/API work

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2993

The appointments test data did not represent expected use cases (e.g., advisor metadata was inserted in db without checked_in status being set) so this refactor included test data overhaul.  Therefore, many expectations in tests had to change.  I'm hoping for very few client-side breakages – we'll take this one step at a time.

### DIFF
--- a/boac/models/appointment.py
+++ b/boac/models/appointment.py
@@ -30,12 +30,13 @@ from boac import db, std_commit
 from boac.externals import data_loch
 from boac.lib.berkeley import BERKELEY_DEPT_CODE_TO_NAME
 from boac.lib.util import camelize, search_result_text_snippet
+from boac.models.appointment_event import appointment_event_type, AppointmentEvent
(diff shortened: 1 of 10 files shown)


boac/models/appointment.py
```diff
@@ -134,8 +130,10 @@ def get_waitlist(cls, dept_code, include_resolved=False):
             )  # noqa: E711
         else:
             criterion = and_(
-                cls.canceled_at == None,
-                cls.checked_in_at == None,
+                or_(
+                    cls.status == None,
+                    cls.status.in_(['reserved', 'unreserved']),
```
johncrossman: This seems to be the preferred SQLAlchemy syntax querying against enum db type. Having `not_(cls.status.in_([...])` did not work.

paulkerschen:
Onward!

PR #2069, opened 2019-11-01
BOAC-2985 Refactor drop-in advisor status into its own table and feed element

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2985


johncrossman:
The boa db had a good day. Our little baby is growing up!

PR #2071, opened 2019-11-01
BOAC-2994, add 'reservedAt' and 'reservedBy' to API

johncrossman:

More tests, too.


paulkerschen:
Right this way, sir.

PR #2072, opened 2019-11-01
BOAC-2986 API wiring for drop-in availability toggle

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2986


lyttam:
Commence the toggling.

PR #2073, opened 2019-11-01
BOAC-2893: Allows custom columns in cohort and group CSV exports

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2893


lyttam:
image
"I love lambdas"

johncrossman:
Thank you @lyttam  I agree with Michael Scott: lambdas are dope.

PR #2074, opened 2019-11-02
BOAC-2995, reserve and un-reserve API endpoints; front-end basics

johncrossman:


paulkerschen:
This endpoint I reserved is undercooked; I want to send it back.

PR #2075, opened 2019-11-03
BOAC-2995, most recent appointment_event for appointments_fts_index; simplify API

johncrossman:

Notable:
* fix `appointments_fts_index`: only grab latest event from `appointment_events` table, no dupe rows per appointment_id
* remove `canceledAt`, `canceledBy`, `reservedAt`, etc. in favor of `statusBy` and `statusDate`
* remove unnecessary calls to `refresh_search_index()`


paulkerschen:
I'll take simplification where we can find it. Thanks @johncrossman.

PR #2076, opened 2019-11-04
BOAC-2990 Drop-in availability toggle, front end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2990


johncrossman:
You just toggle the toggle!

PR #2078, opened 2019-11-04
BOAC-2987 Scheduler may check in only available advisors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2987


johncrossman:
Tune in, turn on, and check in.

PR #2080, opened 2019-11-04
BOAC-2959: Incrementally improves appointment advisor name autocomplete

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2959

With notes, the author's first and last name are stored in separate columns and are separately queryable.  User can search 'sa sm' and get Sam Smith.

Appointments have the entire name in one column, so in order to achieve the same search functionality we need to either split the name or do a full wildcard search.  I picked the latter but put in some benchmarking so I can monitor the query's performance as the appointments table grows in dev.


paulkerschen:
Thanks for adding in the benchmarker, @lyttam - we'll see how it does out there in the big BOA world.

PR #2081, opened 2019-11-05
BOAC-2996, reserve/unreserve from student profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2996


paulkerschen:
It's awesome even without fonts.

PR #2082, opened 2019-11-05
BOAC-2976, if drop-in advisor then show it on user profile page

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2976

For now, this page is only available in dev and qa.


raydavis:
I am proud to work for an organization that implemented an `oxfordJoin`.

PR #2083, opened 2019-11-05
BOAC-2988 BOAC-2991 Redesigned scheduler homepage, drop-in status polling

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2988
https://jira.ets.berkeley.edu/jira/browse/BOAC-2991


johncrossman:
But wait... there's more!!

PR #2087, opened 2019-11-07
BOAC-3004, add/update/delete topics per L&S feedback

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3004

We decided to exclude:
* Degree Check, APR Update, EGT Update
* Probation, Dismissal


paulkerschen:
An ounce of Pre-Health is worth a pound of cure.

PR #2095, opened 2019-11-07
BOAC-2969, clean up Vue router w.r.t drop-in auth

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2969

Shared router logic for drop-in advisor and scheduler.


paulkerschen:
Find your path through the MacArthur Maze.

PR #2096, opened 2019-11-07
BOAC-3008, most recent appointment event: order by created_at, desc

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3008


paulkerschen:
Appointment pirates!

PR #2098, opened 2019-11-08
BOAC-3010, if advisor uid when create appt then all advisor info please

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3010

Passing all advisor info from the client, as done in this PR, is surprisingly less complex than using `advisor_uid` on server-side to pull from various data-sources.


paulkerschen:
It's the combo meal!

PR #2099, opened 2019-11-08
NOJIRA Remove some obsolete admin endpoints

raydavis:
Because it's trash day.


paulkerschen:
The things you find in the back of the freezer.

PR #2104, opened 2019-11-08
BOAC-2961 BOAC-2971 Error handlin' modals

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2961


johncrossman:
Handled.

PR #2108, opened 2019-11-12
BOAC-3015, Admin user gets Drop-in advisor UX on /home/{deptCode}

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3015


paulkerschen:
Thanks @johncrossman!

PR #2110, opened 2019-11-14
BOAC-3026, Pasenger Manifest user search: ignore case

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3026


paulkerschen:
How insensitive.

PR #2130, opened 2019-11-19
BOAC-3029, 'canceled' has been cancelled, in favor of 'cancelled'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3029

Double-L is cooler.

image


paulkerschen:
Call the chancellor!

PR #2132, opened 2019-11-19
BOAC-3046 Revert well-intentioned vue-a11y @blur mandate

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3046


johncrossman:
When good intentions turn bad!

PR #2136, opened 2019-11-20
BOAC-3037, reduce padding in Drop-in Waitlist

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3037


paulkerschen:
Lose the pads.

751cb703ed9302c650bb82257f67f96b

PR #2145, opened 2019-11-21
BOAC-3028 Advisors get confirmation modal and undo for checkin/cancel

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3028


johncrossman:
BOA is rolling forward!

PR #2147, opened 2019-11-21
BOAC-2882, include deleted users in Passenger Manifest search/filter

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2882

* As seen elsewhere, boolean search param can be `true`, `false`, or `null`; each with its own intention.
* Certain `/user` APIs must have include-deleted option


boac/api/user_controller.py
```diff
@@ -180,15 +182,22 @@ def user_search():
...
             pattern = re.compile(any_ + any_.join(snippet.split()) + any_, re.IGNORECASE)
             users = list(filter(lambda u: u.get('name') and pattern.match(u['name']), users))
     else:
         users = []
-    return tolerant_jsonify([{'label': u['name'], 'uid': u['uid']} for u in users])
+
+    def _label(user):
+        name = user['name']
+        return f"{name} ({user['uid']})" if name else user['uid']
+    return tolerant_jsonify([{'label': _label(u), 'uid': u['uid']} for u in users])
```
johncrossman: Front-end `AutoComplete` widget requires non-null label.

lyttam:
Thanks for the fix!

PR #2149, opened 2019-11-22
BOAC-3063, Passenger Manifest > Delete and un-delete users

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3063


paulkerschen:
They don't stay dead.

PR #2150, opened 2019-11-22
BOAC-3061, if author.uid or author.csid then lazy-load calnet profile

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3061

Cause of regression: PR #2128 

Review with `?w=1`


paulkerschen:
Thanks for the fast fix, @johncrossman. Looks right.

PR #2156, opened 2019-11-22
BOAC-3021 Make appointment details modal editable

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3021

### DIFF
--- a/boac/models/appointment.py
+++ b/boac/models/appointment.py
@@ -29,7 +29,7 @@ import re
 from boac import db, std_commit
 from boac.externals import data_loch
 from boac.lib.berkeley import BERKELEY_DEPT_CODE_TO_NAME
-from boac.lib.util import camelize, get_benchmarker, search_result_text_snippet, utc_now
+from boac.lib.util import camelize, get_benchmarker, search_result_text_snippet, titleize, utc_now, vacuum_whitespace
 from boac.merged import calnet
 from boac.models.appointment_event import appointment_event_type, AppointmentEvent
 from boac.models.appointment_read import AppointmentRead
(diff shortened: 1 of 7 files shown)


paulkerschen:
Please don't merge yet - just noticed a glitch needs fixing.

paulkerschen:
Ready for review.

boac/models/appointment.py
```diff
@@ -327,6 +327,12 @@ def search(
         response = [_to_json(search_terms, dict(zip(keys, row))) for row in result.fetchall()]
         return response
 
+    def update(self, details=None, topics=()):
+        self.details = details
+        _update_appointment_topics(self, topics)
+        std_commit()
+        db.session.refresh(self)
```
johncrossman: @paulkerschen - Set `updated_by` on `appointments` table?

src/components/appointment/AppointmentDetailsModal.vue
```diff
@@ -95,20 +93,49 @@ export default {
...
+      }
+    },
+    update() {
+      this.saving = true;
+      this.updateAppointment(
+        this.details,
+        this.topics,
+      );
+      this.showDetailsModal = false;
+      this.saving = false;
```
johncrossman: @paulkerschen - Because `updateAppointment()` returns a Promise, should the two lines above live in `then` callback? (Prolly not a big deal.)

paulkerschen: @johncrossman Having these lines outside the callback means that the modal disappears right away, and the appointment then spins in the waitlist until the server response comes back, which seems consistent with other waitlist actions.

johncrossman: Ah, right. Makes sense.

boac/models/appointment.py
paulkerschen: @johncrossman Agreed, amending.

johncrossman:
Thanks @paulkerschen

PR #2157, opened 2019-11-22
BOAC-3065, Autocomplete dropdown should not bully other form elements

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3065


paulkerschen:
Kidpower!

PR #2158, opened 2019-11-22
BOAC-3067, if /search has no params then go /home

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3067


paulkerschen:
Sleep it off, search view.

PR #2159, opened 2019-11-24
BOAC-3045, topic_unique_constraint includes deleted_at – for notes and appts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3045

For the indecisive advisor: No error when re-adding a topic.


paulkerschen:
A constraint neither too loose nor too tight.

PR #2160, opened 2019-11-25
BOAC-2972, header nav menu > reset Passenger view; no console errors

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2972

No more `NavigationDuplicated error: Navigating to current location ("/admin/passengers") is not allowed`


lyttam:
Looks like one Travis succeeded but another is still queued.  I think Travis is just confused.

PR #2161, opened 2019-11-25
[QA] BOAC-3068, no 404 when advisor not found; nil tolerant front-end

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3068

IMO, the safest fix is to _not_ return 404 from API and make sure front-end is null tolerant.


lyttam:
Makes sense to me.

PR #2166, opened 2019-11-25
BOAC-3068, get-advisor-info is strictly CalNet-based, no authorized_user dependency

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3068

QA PR is #2165


sandeepmjay:
Thanks, @johncrossman. Decoupling advisor authorization dependency from calnet profiles is an elegant solution.

PR #2168, opened 2019-11-26
BOAC-3069, remove unused /api/user/by_csid

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3069

The work of BOAC-3068 made `/api/user/by_csid` obsolete. *Note:* client-side caching of CalNet data is overkill cuz it's cached on server-side; we don't need the overhead in vue-store.


sandeepmjay:
Thanks for the cleanup on unused api's @johncrossman.

PR #2170, opened 2019-11-27
BOAC-3002: Looks to the loch for term configs

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3002

boa-dev will continue to rely on the term configs in boac-dev.py until those are switched to 'auto'.   I'll wait until things are settled on the Nessie side before making that switch.

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -29,7 +29,7 @@ from threading import Thread
 
 from boac import std_commit
 from boac.externals import data_loch
-from boac.lib import berkeley
+from boac.merged.sis_terms import all_term_ids, current_term_id
 from boac.models.alert import Alert
 from boac.models.curated_group import CuratedGroupStudent
 from boac.models.job_progress import JobProgress
@@ -93,6 +93,7 @@ def continue_request_handler():
 def background_thread_refresh(app_arg, term_id):
     with app_arg.app_context():
         try:
+            refresh_current_term_index()
             load_term(term_id)
             JobProgress().end()
         except Exception as e:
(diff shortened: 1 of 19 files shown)


boac/api/cache_utils.py
```diff
@@ -116,15 +117,15 @@ def load_all_terms():
         JobProgress().update(f'Term {term_id} loaded', properties={'terms_done': terms_done})
 
 
-def load_term(term_id=berkeley.current_term_id()):
+def load_term(term_id=current_term_id(use_cache=False)):
```
lyttam: Cache may not yet be initialized when this is called, so we'll bypass it and go straight to the loch.

boac/merged/sis_terms.py
```diff
@@ -0,0 +1,69 @@
...
+        return index and index['current_term_id']
+    return sis_term_id_for_name(term_name)
+
+
+def current_term_name():
+    term_name = app.config['CANVAS_CURRENT_ENROLLMENT_TERM']
+    if term_name == 'auto':
+        index = get_current_term_index()
+        return index and index['current_term_name']
+    return term_name
```
paulkerschen: @lyttam Would it be worth having `current_term_name` take a `use_cache=True` default argument as well, in order to collapse some code? It seems that if we did that, `current_term_id` could be rewritten as a one-liner `return sis_term_id_for_name(current_term_name(use_cache))`.

paulkerschen: @lyttam The deeper question is perhaps whether we want to pull both term ids and term names from the new table, or if we'd prefer to keep things DRY by pulling one representation only and relying on our conversion functions for the other. (For instance, it looks like the `future_term_name` value isn't actually used by application code.) Feel free to push back if this seems too draconian or not worth the trouble to implement.

lyttam: @paulkerschen I think that's a good idea; we'll have to keep logic in both Nessie and BOA to convert term names to term IDS, but we'd be keeping it in both places either way since it's needed for more than just the current and future terms.
I'll amend this PR and then submit another Nessie PR with that change.

lyttam: @paulkerschen Amended.

paulkerschen:
Looks good, @lyttam!

PR #2173, opened 2019-12-02
BOAC-3073 Update nightly appt cancellation script to new schema

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3073


johncrossman:
Makes sense to me. Thanks @paulkerschen

PR #2176, opened 2019-12-03
BOAC-3078, exclude deleted_at when joining with drop_in_advisors table

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3078

In addition to fixing the SQL behind Passenger Manifest I've also snuck in an image for README. Get a preview at https://github.com/johncrossman/boac/tree/BOAC-3078_users-sql


paulkerschen:
What a big beautiful snub-nosed plane.

PR #2179, opened 2019-12-04
BOAC-3081, if self-check-in then CheckInModal need not wait for advisor list

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3081

### DIFF
--- a/src/components/admin/Users.vue
+++ b/src/components/admin/Users.vue
@@ -6,6 +6,7 @@
           <b-form-select
             id="user-filter-options"
             v-model="filterType"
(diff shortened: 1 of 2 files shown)


src/components/admin/Users.vue
```diff
@@ -57,6 +61,7 @@
               <b-form-select
                 id="user-status-options"
                 v-model="filterBy.status"
+                :disabled="isBusy"
```
johncrossman: Snuck in changes above. Passenger Manifest dropdowns are disabled during fetch-users.

PR #2180, opened 2019-12-04
BOAC-3088, scheduler waitlist count is waitlist.unresolved.length

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3088


paulkerschen:
Sensible.

PR #2189, opened 2019-12-05
[QA] BOAC-3089, remove 'config' dependency from auth.ts; no more scheduler freeze

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3089

I can reproduce on boa-dev but not locally. Nonetheless, I have a hunch on infinite loop. This is a no-risk change.


paulkerschen:
Let's give it a go.

PR #2198, opened 2019-12-06
BOAC-3070: Deletes orphan dropin advisors on dept refresh

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3070


paulkerschen:
orphan

PR #2209, opened 2019-12-09
BOAC-3102, do not set apptDeskRefreshInterval on user-initiated appt update

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3102

We do not want another refresh-waitlist thread every time user edits an appt.


paulkerschen:
Looks good, @johncrossman - but I think `DropInDesk.vue` needs the same treatment?

johncrossman:
Amending...

paulkerschen:
Waiting on T... QA PR too?

PR #2215, opened 2019-12-11
BOAC-3128, Passenger Manifest: Quick links to drop-in advisors, L&S advisors, and schedulers

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3128

image

### DIFF
--- a/src/components/admin/Users.vue
+++ b/src/components/admin/Users.vue
@@ -74,7 +74,48 @@
         </b-col>
       </b-row>
     </b-container>
-    <div class="font-size-14 mb-2 ml-4 total-user-count">
+    <div class="align-items-center d-flex pl-4">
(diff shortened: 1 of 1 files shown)


src/components/admin/Users.vue
```diff
@@ -74,7 +74,48 @@
...
+        |
+      </div>
+      <div>
+        <b-btn
+          id="quick-link-advisors"
+          :disabled="isBusy"
+          class="pl-2 pr-2"
+          variant="link"
+          @click="quickLink('advisor', 'QCADV')">
+          L&S Advisors
```
paulkerschen: @johncrossman Handy! But Github has put an angry red box above. I guess we're still escaping ampersands in 2019?

PR #2217, opened 2019-12-11
BOAC-3111 BOAC-3127 Add minors to profile, move other elements around

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3111
https://jira.ets.berkeley.edu/jira/browse/BOAC-3127


johncrossman:
Thanks @paulkerschen

PR #2219, opened 2019-12-11
BOAC-3129, remove appointments feature flag and includeAppointments param in search

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3129

A lot of clean up in `test_search_controller`, too.


paulkerschen:
Thanks for the extra cleanup, @johncrossman. That feature flag really got around, didn't it.

PR #2225, opened 2019-12-12
BOAC-3102, fix for un-assign self from appt, then appt not updated immediately

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3102

The appointment API – in case of reserve, unreserve, cancel, and check-in – does not return an appointment object (JSON). This streamlines the transactions and has the front-end feeling zippier.

### DIFF
--- a/boac/api/util.py
+++ b/boac/api/util.py
@@ -65,6 +65,7 @@ def advisor_required(func):
             and (
                 current_user.is_admin
                 or _has_role_in_any_department(current_user, 'isAdvisor')
+                or _has_role_in_any_department(current_user, 'isDirector')
             )
         if is_authorized or _api_key_ok():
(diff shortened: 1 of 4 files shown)


boac/api/util.py
```diff
@@ -65,6 +65,7 @@ def _advisor_required(*args, **kw):
             and (
                 current_user.is_admin
                 or _has_role_in_any_department(current_user, 'isAdvisor')
+                or _has_role_in_any_department(current_user, 'isDirector')
```
johncrossman: This PR also fixes a bug whereby a user with "Director" role was unable to view an appointment. This would cause error on student profile page.

paulkerschen:
Seems reasonable to me.

PR #2230, opened 2019-12-13
BOAC-3108: Allows non-current students to turn up in auto-suggest by SID only

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3108

I chatted with Steven about this and we agreed to try it out in dev. If it makes for a confusing UX we'll look at alternatives.


paulkerschen:
The scales fall from my eyes! Let's see how dev behaves.

PR #2231, opened 2019-12-13
BOAC-3134, if we load $currentUser and $config before Vue.init then we get rid of messy vuex code

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3134

See `main.ts` and what is done prior to `new Vue(...).$mount()`.

Example of code simplification:
* No more `store.dispatch('user/loadUser')...` every time we want to verify auth.
* We do not need a mixin to reference configs (eg, `devAuthEnabled`). Instead, use `$config.devAuthEnabled` in your template/component.


paulkerschen:
@johncrossman This makes a lot of sense and I'm glad to lose the need for those mixins. A lot of our components are probably now unnecessarily including `Context` and `UserMetadata` mixins, in a way that's not easy to clean up automatically. Should the tedious work of taking out those mixins be left for a follow-up?

PR #2234, opened 2019-12-16
BOAC-3110: Denotes non-current students in waitlist with an icon

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3110


johncrossman:
Thanks @lyttam  Let's give it a go.

PR #2239, opened 2019-12-17
BOAC-3146, remove mixins/Session; init pingFrequency at startup

johncrossman:

* removed `suggestedNoteTopics` from UserMetadata – it's no longer cached in `store`
* by moving `ping` out of `mixin/Session` we avoid risk of kicking off multiple ping loops


paulkerschen:
Makes sense. One ping per customer.

PR #2243, opened 2019-12-17
BOAC-3148 Discontinued minors below the fold

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3148


johncrossman:
Below the fold, but newsworthy.

PR #2244, opened 2019-12-17
BOAC-3146, remove noteTemplates from mixin/UserMetadata; retire store/modules/note.ts

johncrossman:

And, we no longer load noteTemplates at startup – they are lazy-loaded by `CreateNoteModal`.


paulkerschen:
I had already forgotten user-specific note templates were a thing. Like sands through the hourglass...

PR #2246, opened 2019-12-18
BOAC-1459: Adds new cohort filter for visa type

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1459


paulkerschen:
Looks good!

PR #2253, opened 2019-12-19
BOAC-3146, move Google Analytics out of vuex store and UserMetadata mixin

johncrossman:

In `main.ts`, two async calls to init user and ga:
```
$core.initializeCurrentUser();
$core.mountGoogleAnalytics();
```

`$core` is where we keep low-level utilities that initialize prototype objects. These are invoked again when user logs in via dev-auth.


paulkerschen:
Cool, I always wanted a core.

PR #2254, opened 2019-12-19
BOAC-3146, process.env variables put to $config and removed from 'store'

johncrossman:

A "store" (vuex) should not maintain static configs. This PR reduces `context.ts` to state variables that might actually mutate.


paulkerschen:
The store is not the warehouse.

PR #2256, opened 2019-12-19
BOAC-3146, CurrentUserExtras has all user related and drops 'user' from store

johncrossman:

Last of the `UserMetadata` refactor. This is a big diff but the types of changes are few. Moving `user` out of store means many files updated. If `currentUserExtras` is an odd name then please suggest some other.  Let's be judicious w.r.t using `actions` in vuex store – they're only necessary when you want to map to a method in a mixin.  In most cases, a `store.commit` (direct call to mutation) is preferred.

The import of `CurrentUserExtras` can prolly be removed from certain components; that can wait.


paulkerschen:
I thought all those `actions` were a Best Practice! Thanks @johncrossman - I won't miss them.

PR #2261, opened 2019-12-23
BOAC-3170: Allows admin to assign supervisor on-call role to users

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3170


paulkerschen:
Super!

PR #2264, opened 2020-01-03
BOAC-3168: Allows user to change drop-in roles on Settings page

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3168

### DIFF
--- a/boac/api/cache_utils.py
+++ b/boac/api/cache_utils.py
@@ -167,7 +167,6 @@ def refresh_department_memberships():
     depts = UniversityDept.query.all()
     for dept in depts:
         dept.delete_automated_members()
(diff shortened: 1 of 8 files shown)


boac/api/cache_utils.py
```diff
@@ -187,6 +186,7 @@ def refresh_department_memberships():
                     is_director=False,
                     is_scheduler=False,
                 )
+    DropInAdvisor.delete_orphans()
```
lyttam: Unit tests helped uncover this bug I introduced in a previous PR.  We don't want to delete any `drop_in_advisor` records until department memberships have been restored.

paulkerschen:
Assume, relinquish, assume, relinquish.

PR #2278, opened 2020-01-07
BOAC-3190, search history server-side, with USER_SEARCH_HISTORY_MAX_SIZE config

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3190


paulkerschen:
It's a pretty nice search history, but it changes from day to day.

PR #2279, opened 2020-01-07
BOAC-3191, suggestions from search-history using Autocomplete component

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3191

Suggestions from search history are shown when you first put focus on input and the list narrows as you type. The new `restrict` prop in `Autocomplete` allows for a query that is _not_ in list o' suggestions.

### DIFF
--- a/src/components/util/Autocomplete.vue
+++ b/src/components/util/Autocomplete.vue
@@ -4,18 +4,22 @@
       :id="`${id}-input`"
       ref="autocompleteInput"
       v-model="query"
+      :aria-readonly="disabled"
+      :class="inputClass"
       :disabled="disabled"
       :placeholder="placeholder"
       maxlength="56"
       name="autocomplete-name"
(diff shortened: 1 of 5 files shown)


src/components/util/Autocomplete.vue
```diff
@@ -159,20 +205,21 @@ export default {
...
     },
-    _onTextInput() {
-      this.$emit('input', null);
-      if (this.query.length > 1) {
-        this.isOpen = true;
-        this.isLoading = true;
-        this.suggestions = [];
-        this.source(this.query.replace(/[^\w ]+/g, ''), this.limit).then(results => {
-          this.populateSuggestions(results);
-        });
```
johncrossman: The function above was renamed `makeSuggestions`.

paulkerschen:
It's Google!

PR #2280, opened 2020-01-07
BOAC-3153 Show discontinued plans above the fold if no active majors

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3153


johncrossman:
BOA: the stuff you need, above the fold.

PR #2283, opened 2020-01-08
BOAC-2023: Adds descending sort for cohorts and groups

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2023


johncrossman:
Let's give it a go. Thanks @lyttam

PR #2287, opened 2020-01-09
BOAC-3205, escape search string when RegExp

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3205


paulkerschen:
escape

PR #2296, opened 2020-01-10
BOAC-3158 Prefer full Calnet names for advisors where feasible

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3158


johncrossman:
We prefer what is feasible.

PR #2302, opened 2020-01-13
BOAC-3196, server-side /api/reports will serve front-end /analytics/:deptCode

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3196

Designs: https://jira.ets.berkeley.edu/jira/browse/BOAC-2792


paulkerschen:
The stats don't lie.

PR #2303, opened 2020-01-13
BOAC-3181: Adds setting to enable drop-in advising by department

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3181

### DIFF
--- a/boac/api/util.py
+++ b/boac/api/util.py
@@ -97,8 +97,8 @@ def scheduler_required(func):
         is_authorized = current_user.is_authenticated \
             and (
                 current_user.is_admin
-                or current_user.is_drop_in_advisor
-                or _has_role_in_any_department(current_user, 'isScheduler')
+                or _is_drop_in_advisor(current_user)
+                or _is_drop_in_scheduler(current_user)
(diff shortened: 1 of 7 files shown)


config/test.py
```diff
@@ -30,6 +30,8 @@
 
 DATA_LOCH_RDS_URI = 'postgres://boac:boac@localhost:5432/boac_loch_test'
 
+DEPARTMENTS_SUPPORTING_DROP_INS = ['QCADV']
+
```
johncrossman: @lyttam - Do we need the `test` config above?  You've nicely used `override_config` instead. Plus, it's nice to have minimal reliance on this test config file.

johncrossman:
PR looks good, although Travis test failures need attention. Thanks @lyttam

config/test.py
lyttam: @johncrossman I'll remove it as part of the fix for the failing tests.

lyttam:
Ready for another look, @johncrossman.

johncrossman:
Thanks @lyttam

PR #2304, opened 2020-01-13
BOAC-3221 Expand drop-in advisor status types

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3221

No functional change as yet, but we quietly move our four-valued enum into place behind the scenes.


johncrossman:
image

PR #2307, opened 2020-01-14
BOAC-3196, /analytics page: percentage of notes with attachments, etc.

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3196


boac/api/admin_controller.py
```diff
@@ -74,9 +73,3 @@ def start_continuation_of_interrupted_job():
 def start_refresh():
     response = cache_utils.refresh_request_handler(term())
     return tolerant_jsonify(response, status=500 if 'error' in response else 200)
-
-
-@app.route('/api/admin/report/low_assignment_scores')
```
johncrossman: I moved the route above to `reports_controller` and consolidated report generators in `merged/reports`.

PR #2310, opened 2020-01-15
BOAC-3223 Three-way toggle for drop-in advisor status

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3223

Gets the front-end element in place and working, but doesn't yet disentangle the new supervisor status from the old supervisor person type.


johncrossman:
Up, down, and side-to-side. Thanks @paulkerschen

PR #2312, opened 2020-01-15
BOAC-3223 Revise appearance of drop-in status toggle

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3223

Design revision in https://jira.ets.berkeley.edu/jira/browse/BOAC-3227.


johncrossman:
image

PR #2315, opened 2020-01-16
BOAC-3161: Allows advisor to toggle own drop-in participation

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3161


paulkerschen:
Thanks for sorting through this tangle, @lyttam. I’m glad we get to keep our fontawesome toggle somewhere.

PR #2317, opened 2020-01-16
BOAC-2023: Sorts null term GPAs last regardless of sort direction

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2023


paulkerschen:
Xor!

PR #2319, opened 2020-01-17
BOAC-3238 Standardize representation of appointment status and events

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3238

### DIFF
--- a/boac/models/appointment.py
+++ b/boac/models/appointment.py
@@ -115,7 +115,8 @@ class Appointment(Base):
             token_conditions.append(
                 f"""JOIN appointments a{idx}
                 ON UPPER(a{idx}.advisor_name) LIKE :token_{idx}
-                AND a{idx}.advisor_uid = a.advisor_uid""",
+                AND a{idx}.advisor_uid = a.advisor_uid
+                AND a{idx}.advisor_name = a.advisor_name""",
             )
             params[f'token_{idx}'] = f'%{token}%'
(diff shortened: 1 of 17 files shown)


boac/models/appointment.py
```diff
@@ -115,7 +115,8 @@ def find_advisors_by_name(cls, tokens, limit=None):
             token_conditions.append(
                 f"""JOIN appointments a{idx}
                 ON UPPER(a{idx}.advisor_name) LIKE :token_{idx}
-                AND a{idx}.advisor_uid = a.advisor_uid""",
+                AND a{idx}.advisor_uid = a.advisor_uid
+                AND a{idx}.advisor_name = a.advisor_name""",
```
paulkerschen: Fixes an edge case where, if the advisor name associated with a UID changes, the non-matching name would come back in results.

tests/util.py
```diff
@@ -54,13 +54,17 @@ def override_config(app, key, value):
     """Temporarily override an app config value."""
     old_value = app.config[key]
     app.config[key] = value
-    yield
-    app.config[key] = old_value
+    try:
+        yield
+    finally:
+        app.config[key] = old_value
```
paulkerschen: Ensure that the config resets even if a test fails; otherwise confusing later failures ensue.

johncrossman:
Another hard night of refactoring.

image

Thanks @paulkerschen

PR #2322, opened 2020-01-18
BOAC-3236 Adjust dates in CSV export to app timezone

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3236


johncrossman:
Timezones be crazy AF.

PR #2327, opened 2020-01-23
BOAC-3145 Handle server 401s by redirecting to login page

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3145


johncrossman:
Beware the interceptors!

PR #2329, opened 2020-01-24
BOAC-3098, get_sis_note_attachments: priority given to is_historical=FALSE (per SID, note_id)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3098

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -506,9 +506,11 @@ def get_sis_advising_note_attachment(sid, filename):
 
 
 def get_sis_advising_note_attachments(sid):
-    sql = f"""SELECT advising_note_id, created_by, sis_file_name, user_file_name
+    # Priority is given to is_historical=FALSE.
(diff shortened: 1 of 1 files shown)


paulkerschen:
@johncrossman Tricky! Out of curiosity, is there any reason one couldn’t do the same thing directly in SQL using an aggregate function? Something like

{code}
SELECT advising_note_id, created_by, sis_file_name, user_file_name, MIN(is_historical)
...
GROUP BY advising_note_id, created_by, sis_file_name, user_file_name
{code}

johncrossman:
@paulkerschen – I was looking for such magical SQL. In my tests, I found `MIN(...)` to be incompatible with type boolean. https://stackoverflow.com/questions/20281125/why-aggregate-functions-in-postgresql-do-not-work-with-boolean-data-type

I can play around a bit more.

johncrossman:
Ah, I think `MIN(is_historical::int)` is the secret sauce.

johncrossman:
I'll test then, if all good, amend the PR.

paulkerschen:
There it is!

johncrossman:
Amended

PR #2331, opened 2020-01-25
BOAC-3266 BOAC-3267 Revert a bunch of drop-in work

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3266
https://jira.ets.berkeley.edu/jira/browse/BOAC-3267


johncrossman:
We zigged when we were supposed to zag.

PR #2334, opened 2020-01-27
BOAC-3268: Switches drop-in advisor soft delete to a hard delete

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3268


paulkerschen:
Hardcore!

PR #2337, opened 2020-01-28
BOAC-3269 Back end for drop-in status text

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3269

Adding a new status field required cleaning up some previous loose usage of the word "status."

The "isAvailable" boolean feed value represents not a "status" but an "availability."

The entire feed object corresponding to a row in the drop_in_advisors table is not a "status" but a "membership."

### DIFF
--- a/scripts/db/migrate/2020/20200128-BOAC-3269/pre_deploy_01_add_drop_in_advisors_status.sql
+++ b/scripts/db/migrate/2020/20200128-BOAC-3269/pre_deploy_01_add_drop_in_advisors_status.sql
@@ -1,0 +1,1 @@
+ALTER TABLE drop_in_advisors ADD COLUMN status character varying(255);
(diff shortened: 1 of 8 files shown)


scripts/db/migrate/2020/20200128-BOAC-3269/pre_deploy_01_add_drop_in_advisors_status.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE drop_in_advisors ADD COLUMN status character varying(255);
```
johncrossman: @paulkerschen - Is this column nullable?

paulkerschen: @johncrossman Yes, and starts out null when new drop-in memberships are created.

PR #2340, opened 2020-01-29
BOAC-3265: Consolidates role booleans into an enum

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3265


paulkerschen:
Losing 100 net lines of code certainly feels refreshing.

PR #2350, opened 2020-01-30
BOAC-3254: Unreserve appointments when advisor stops doing drop-ins

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3254


paulkerschen:
When the drop-in drops out. Thanks @lyttam.

PR #2355, opened 2020-02-03
BOAC-3310, cohort with curated-group filter viewable by others

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3310


lyttam:
Curated cohorts for all.

PR #2359, opened 2020-02-05
BOAC-3271: Adds college cohort filter

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3272


johncrossman:
The cohort is in the college and the college is in the cohort.

PR #2361, opened 2020-02-05
BOAC-3321, add 'domain' property to cohort object; filter-options per domain

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3321

Also added config `FEATURE_FLAG_ADMITTED_STUDENTS`.


paulkerschen:
Cohort filters dominate the default domain!

PR #2366, opened 2020-02-06
BOAC-3278 BOAC-3279 Cohort history front end

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3278
https://jira.ets.berkeley.edu/jira/browse/BOAC-3279


johncrossman:
@paulkerschen - do both cohort types get a "History"?

paulkerschen:
@johncrossman I suppose the answer is "yes," assuming the new cohort type includes the old CohortPageHeader component. My guess is that both the "Export List" and "History" options in that header are making false assumptions about students in the cohort, and will need to be either reworked or suppressed.

johncrossman:
Thanks @paulkerschen

PR #2368, opened 2020-02-07
BOAC-3341 Cohort history: include historical students, handle gaps

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3341


johncrossman:
Gaps must be minded.

PR #2371, opened 2020-02-07
BOAC-3343, 'admitted_students' cohorts live in the present, no history

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3343


paulkerschen:
And in the future!

PR #2373, opened 2020-02-10
BOAC-3336 Ensure unique keys for cohort filter rows

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3336


lyttam:
Tricky!

PR #2378, opened 2020-02-11
BOAC-3339 Don't let curated group edit modes loiter around the store

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3339


lyttam:
@paulkerschen I saw the same test failing intermittently. I think we have a race condition somewhere.

paulkerschen:
Rebased to clear up test failure.

PR #2380, opened 2020-02-12
BOAC-3332: Displays OUA record on admit page

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3332


paulkerschen:
It's a whole new FontAwesome world.

PR #2394, opened 2020-02-19
BOAC-3388 BOAC-3390 Back end for drop-in scheduler management

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3388
https://jira.ets.berkeley.edu/jira/browse/BOAC-3390


johncrossman:
Ain't nothin' like self-service.

PR #2395, opened 2020-02-19
BOAC-3349: API for searching admits

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3349

### DIFF
--- a/boac/api/search_controller.py
+++ b/boac/api/search_controller.py
@@ -27,10 +27,11 @@ from datetime import timedelta
 from itertools import islice
 
 from boac.api.errors import BadRequestError, ForbiddenRequestError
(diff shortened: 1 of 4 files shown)


paulkerschen:
API looks good to me, @lyttam. As far as front-end functionality, will “Admits” be another checkbox selectable in any combination with existing checkboxes? If “Admits” is selected in combination with another checkbox, will the front end then fire off two simultaneous requests to get the full result set?

lyttam:
@paulkerschen Yes, that's the idea. I went back and forth but ultimately decided to give admit search a separate endpoint in order to leverage the existing `ce3_required` function. Any arguments against firing 2 requests?

paulkerschen:
No, I think sealing off the endpoint behind a standard permission check is a good idea. Thanks!

PR #2401, opened 2020-02-20
BOAC-3349: Adds option to search admits

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3346

### DIFF
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -1,6 +1,5 @@
 import axios from 'axios';
 import utils from '@/api/api-utils';
-import Vue from "vue";
 
 export function getMySearchHistory() {
   return axios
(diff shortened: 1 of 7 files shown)


src/api/search.ts
```diff
@@ -39,9 +38,15 @@ export function search(
       offset: offset || 0,
       limit: limit || 50
     })
-    .then(response => {
-      Vue.prototype.$ga.searchEvent(`Search phrase: ${phrase}`);
```
johncrossman: @lyttam - Intentional removal of Google Analytics call?

lyttam: @johncrossman Yes, it looked like there were 3 GA calls being made for every search, so I reduced it to one (in Search.vue).

johncrossman:
Thanks for the `ga` cleanup.

PR #2403, opened 2020-02-21
BOAC-3347: Displays first 50 admit search results

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3347


paulkerschen:
islice, uslice, we all see no admit twice.

PR #2405, opened 2020-02-21
BOAC-3373: Adds range valiation for dependents cohort filters

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3373

I don't want to guess at a maximum number of dependents, especially if it's student-reported. This validates that min > 0 and max is no more than 5 characters (99999, if someone really wants to filter like that).


paulkerschen:
Thinking about large numbers of dependents is stressing me out.

PR #2406, opened 2020-02-24
BOAC-3409: Adapts admit cohort filters to real data

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3409


paulkerschen:
So many strings! Thanks, @lyttam, let's get filtering.

PR #2409, opened 2020-02-25
BOAC-3407: Reuses admit search result list for cohort list

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3407

I will add pagination in a subsequent PR.


johncrossman:
Reuse, recycle, and admit.

PR #2410, opened 2020-02-25
BOAC-3404: Adds int'l status, etc to admit page

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3404

Also, table stripes for readability.


paulkerschen:
172d1f977cc0a280214411e7fff20e63

PR #2412, opened 2020-02-27
BOAC-3407: Adds sort dropdown for admit cohorts

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3407

Sorting via dropdown (rather than column headers) is required for pagination to work.

- Switches source for '1st Gen' 
- Admit cohort list view gets its own component separate from admit search results view
- Adds `admitSortBy` preference because sort selection must be stored in the store (not in the components)
- Makes regular `sortBy` preference sticky between cohorts and curated groups


paulkerschen:
It had to happen!

PR #2413, opened 2020-02-27
BOAC-3374: Puts first name first when sorting by first_name

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3374


paulkerschen:
A very good place to start.

PR #2414, opened 2020-02-27
BOAC-3334 So long, cohort_filter_owners

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3334


johncrossman:
One cohort, one owner. It's the law.

PR #2416, opened 2020-02-28
BOAC-3324: Links to admit's student page only if there's a student record

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3324


src/views/AdmitStudent.vue
```diff
@@ -10,10 +10,10 @@
         tabindex="0">
         {{ admit.firstName }}&nbsp;{{ admit.middleName }}&nbsp;{{ admit.lastName }}
       </h1>
-      <div v-if="admit.uid" class="pt-2 pb-3">
+      <div v-if="admit.studentUid" class="pt-2 pb-3">
         <router-link
-          :id="`link-to-student-${admit.uid}`"
-          :to="studentRoutePath(admit.uid, $currentUser.inDemoMode)"
+          :id="`link-to-student-${admit.studentUid}`"
+          :to="studentRoutePath(admit.studentUid, $currentUser.inDemoMode)"
```
johncrossman: @lyttam - If OUA sends us `student_uid` do we assume that BOA can serve that student profile?

lyttam: @johncrossman OUA is sending `uid`; `student_uid` is coming from the student profile (which may or may not exist).  BOA will rely solely on `student_uid` to link to the student profile, and use `uid` for display only.

johncrossman:
Thanks @lyttam

PR #2418, opened 2020-02-28
BOAC-3420 Admit cohorts: ensure student count returned even when no results

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3420


johncrossman:
'No results' is a result.

PR #2438, opened 2020-03-06
BOAC-3426: Allows user to view unfiltered list of admits

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3426


paulkerschen:
When is a cohort not a cohort?

PR #2448, opened 2020-03-17
BOAC-2981: Adds cohort filter: intended major

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2981


johncrossman:
Thanks @lyttam  Looks good.

PR #2449, opened 2020-03-19
BOAC-3477 Expand appointments model/table to cover scheduled appts

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3477

### DIFF
--- a/boac/models/appointment.py
+++ b/boac/models/appointment.py
@@ -43,11 +43,27 @@ from boac.models.base import Base
 from dateutil.tz import tzutc
 from flask import current_app as app
 from sqlalchemy import and_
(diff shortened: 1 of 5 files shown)


boac/models/appointment.py
```diff
@@ -43,13 +43,29 @@
...
+    'email',
+    'phone',
+    name='appointment_student_contact_types',
+    create_type=False,
+)
+
+
+appointment_type_enum = ENUM(
+    'Drop-in',
+    'Scheduled',
```
paulkerschen: These capitalized values don't quite follow enum convention, but this keeps consistent with existing data.

johncrossman:
Conventions are overrated. Thx @paulkerschen

PR #2452, opened 2020-03-23
BOAC-3479 New models extend AuthorizedUser for per-dept roles

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3479

Same-day advisors and appointment schedulers are acquiring further attributes that will need to be stored in dedicated tables, as with drop-in advisors. Since all these model types have similar interdependencies with our vanilla user/membership tables, we collect them in a class hierarchy under AuthorizedUserExtension.


boac/models/authorized_user_extension.py
```diff
@@ -25,17 +25,58 @@
...
 from boac.models.base import Base
+from sqlalchemy.ext.declarative import declared_attr
 
 
-class DropInAdvisor(Base):
-    __tablename__ = 'drop_in_advisors'
+class AuthorizedUserExtension(Base):
+    __abstract__ = True
+
+    @declared_attr
```
paulkerschen: This decorator is SQLAlchemy's recommended trick for handling relationships under model inheritance: https://docs.sqlalchemy.org/en/13/orm/extensions/declarative/mixins.html

PR #2478, opened 2020-04-17
BOAC-3527: Locks down @apollographql/apollo-tools version

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3527


package-lock.json
```diff
@@ -10339,7 +10339,7 @@
       "integrity": "sha512-NJdV8aGVGCrcm1Pbq8IbmV5FfLCAEPxgplh9EJU7qAP+Z4PenkuH6V6RAPRqIwwZ287m8/aDXKW+X0qFe8gyUQ==",
       "dev": true,
       "requires": {
-        "@apollographql/apollo-tools": "^0.4.0",
+        "@apollographql/apollo-tools": "0.4.0",
```
johncrossman, line 10342: @lyttam I think we need to lock this down in `package.json`.  As I understand it, the lock file is kinda hands-off.

paulkerschen: Yes, the problem with making manual changes to the lock file is that they tend to get overwritten the next time `npm install` is run.

lyttam: @johncrossman @paulkerschen  Adding apollo-tools to package.json would make it seem a direct BOA dependency though. Is that what we want?

johncrossman: Yes, it's our only option. Unless we want to downgrade what is causing the apollo version dependency.

PR #2479, opened 2020-04-17
BOAC-3527: Locks down @apollographql/apollo-tools version

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3527


johncrossman:
@lyttam @paulkerschen - The underlying issue was fixed: https://github.com/vuejs/vue-cli/issues/5403   The Diablo builds were also failing and now they're not.

PR #2482, opened 2020-04-21
BOAC-3521, force AWS CodeBuild failure to test Slack integration

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3521

In Jira above, Darlene asks, "Is there a way to trigger a build failure to see if the CodePipeline notification rule catches it?"

Yes there is...


sandeepmjay:
Interesting. Thanks, @johncrossman.

PR #2485, opened 2020-04-21
BOAC-3498: Includes BOA note authors in advisor search filter autocomplete

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3498


paulkerschen:
At last, these authors get the credit they deserve.

PR #2488, opened 2020-04-24
BOAC-3537 Friends don't let friends surface blank options in cohort filter menus

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3537


johncrossman:
Blank is the new null.

PR #2494, opened 2020-04-28
[QA] BOAC-3542: Adds advisor index view, refactors advisor autocomplete

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3542

master PR is #2493


paulkerschen:
The PR so nice we'll say it twice.

PR #2507, opened 2020-06-12
BOAC-3577: Displays Data Science notes in Academic Timeline

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3577


paulkerschen:
More data! More data about the data!

PR #2508, opened 2020-06-16
BOAC-3578: Includes Data Science notes in search results

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3578


johncrossman:
@lyttam  Looks good although Travis has failure, "'Attribute' object has no attribute 'id'".

PR #2517, opened 2020-06-29
BOAC-3580 Remove fallback to deprecated CalNet attributes

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3580


johncrossman:
CalNet attributes are not forever.

PR #2520, opened 2020-07-14
BOAC-3594 BOAC-3595 Updates to alert refresh logic

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3594
https://jira.ets.berkeley.edu/jira/browse/BOAC-3595


johncrossman:
Thank you for the inline documentation, @paulkerschen

PR #2535, opened 2020-08-06
BOAC-3614, improve create-note perf by using bulk-create for all

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3614

A single endpoint for both single and batch note creation. Less code and faster create.


johncrossman:
A couple client-side enhancements will follow.

paulkerschen:
Cheaper by the dozen.

PR #2540, opened 2020-08-10
BOAC-3315, cohort filter: Academic Status

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3315


paulkerschen:
There they are!

PR #2551, opened 2020-08-18
BOAC-3631, no location, no em-dash

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3631


paulkerschen:
Location—location—location.

PR #2552, opened 2020-08-18
BOAC-3614, speed up modal actions to, in eyes of Teena/user, end transaction more quickly

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3614

Running Teena locally tells me that things are "faster".


paulkerschen:
Vroom!

PR #2557, opened 2020-08-19
BOAC-3315, 'Academic Standing' cohort filter grouped by term

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3315

Also:
- Academic Standing term info in list view and on student profile
- "grouped options" is new to the cohort options paradigm


paulkerschen:
It’s about time we grouped our groups.

PR #2561, opened 2020-08-20
BOAC-3468: Updated 'create cohort' header text

cesarvh:


lyttam:
Looks great @cesarvh. Don't forget to put a link to the Jira in this PR (and vice versa).

PR #2587, opened 2020-08-27
BOAC-3630, academic_standing alerts: use action_date for created_at and preserve

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3630

In this scheme for `academic_standing` alerts, we set `created_at = action_date` and then leverage `preserve_creation_date`. But, with a fixed `created_at` date we run up against the uniqueness constraint of the `alerts` table: if we `create_or_activate()` twice then we hit uniqueness violation. We work around this by adding (1) `action_date` to the key and (2) optional `create_or_activate(... force_use_existing)` arg – for now, used solely by `academic_standing` alerts.

I'll submit QA PR if/when this is merged.

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -372,7 +372,7 @@ def extract_valid_sids(sids):
 
 
 def get_academic_standing(sids):
-    sql = f"""SELECT sid, term_id, acad_standing_status as status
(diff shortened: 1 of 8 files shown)


johncrossman:
Build is green. (Why Travis kicks off two is a mystery.)

paulkerschen:
Seems like a good threading of a tricky problem to me, @johncrossman. Thanks!

PR #2592, opened 2020-08-28
BOAC-3660, benchmark the performance of note creation

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3660


paulkerschen:
Hi again, benchmarker!

PR #2597, opened 2020-09-02
BOAC-3666: Skips admit search when search phrase not provided

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3666

### DIFF
--- a/boac/api/search_controller.py
+++ b/boac/api/search_controller.py
@@ -89,6 +89,8 @@ def search():
 def search_admits():
     params = request.get_json()
     search_phrase = util.get(params, 'searchPhrase', '').strip()
+    if not len(search_phrase):
+        raise BadRequestError('Invalid or empty search input')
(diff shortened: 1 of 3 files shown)


boac/api/search_controller.py
```diff
@@ -89,6 +89,8 @@ def search():
 def search_admits():
     params = request.get_json()
     search_phrase = util.get(params, 'searchPhrase', '').strip()
+    if not len(search_phrase):
+        raise BadRequestError('Invalid or empty search input')
```
johncrossman, line 93: @lyttam - Could it be that an empty search phrase is acceptable? For example, search by topic alone, without a phrase.

lyttam: @johncrossman An empty search phrase is valid for searching notes and appointments, but not for searching admitted students (nor students nor courses).

PR #2598, opened 2020-09-09
BOAC-3723, when cohort viewed by non-owner, always query with owner_id

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3723


johncrossman:
Follow-up PR will resolve the `/api/cohort/download_csv_per_filters` scenario mentioned in the Jira.

PR #2599, opened 2020-09-10
BOAC-3723, CSV is same for owner and observer when criteria has 'My'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3723


lyttam:
My my my!

PR #2600, opened 2020-09-11
BOAC-3656, disable create/edit note when BOA session expires, with messaging

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3656

This UI change does not perfectly match design in https://jira.ets.berkeley.edu/jira/browse/BOAC-3628 but it serves its purpose. QA and UX can state their opinions after merge.

Use `?w=1`

In screenshot below, notice that buttons are gone and alert is shown. Plus, all form elements are disabled.

image


paulkerschen:
Uh oh, indeed! Looks good but where’s the plane icon?

PR #2609, opened 2020-10-12
BOAC-3744, vue-bootstrap and other package updates; fix styles and bootstrap breakage

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3744

* Bootstrap enhancements: https://bootstrap-vue.org/docs/reference/changelog
* eslint rule to enforce "do not mutate props" (which we've done in a few places, thus the TODOs)


src/components/sidebar/SearchForm.vue
```diff
@@ -48,7 +48,8 @@
           </div>
           <b-btn
             id="search-options-panel-toggle"
-            v-b-toggle="'search-options-panel'"
+            aria-controls="search-options-panel"
+            :aria-expanded="showSearchOptions"
             class="pr-0 pt-0 search-options-panel-toggle"
             variant="link"
             @click="toggleSearchOptions">
```
johncrossman, line 55: This was the bootstrap breakage. We shouldn't use `v-model` _and_ `v-b-toggle`.

PR #2622, opened 2020-10-14
BOAC-3688 Refocus on completed search

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3688

### DIFF
--- a/src/views/Search.vue
+++ b/src/views/Search.vue
@@ -2,7 +2,7 @@
   <div class="m-3">
     <Spinner :is-plural="true" alert-prefix="Search results" />
     <div v-if="loading || (results.totalStudentCount || results.totalCourseCount || results.totalAdmitCount || size(results.notes) || size(results.appointments))">
-      <h1 id="page-header" class="sr-only sr-only-focusable">Search Results</h1>
+      <h1 id="page-header" class="sr-only" tabindex="0">Search Results</h1>
     </div>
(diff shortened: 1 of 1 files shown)


src/views/Search.vue
```diff
@@ -2,15 +2,15 @@
   <div class="m-3">
     <Spinner :is-plural="true" alert-prefix="Search results" />
     <div v-if="loading || (results.totalStudentCount || results.totalCourseCount || results.totalAdmitCount || size(results.notes) || size(results.appointments))">
-      <h1 id="page-header" class="sr-only sr-only-focusable">Search Results</h1>
+      <h1 id="page-header" class="sr-only" tabindex="0">Search Results</h1>
```
paulkerschen, line 5: In my screenreader testing, `sr-only-focusable` didn’t make the element programmatically focusable, as I was expecting. Instead, it appears that you still need to do it with a tabindex; and once you’ve done that, the behavior of `sr-only-focusable` is to make the element appear to _sighted_ users when it gets focus. I assume we don’t want that.

src/views/Search.vue
```diff
@@ -76,7 +76,7 @@
         :render-primary-header="!results.totalStudentCount && !!results.totalCourseCount && !size(results.notes)" />
     </div>
     <div v-if="!loading && size(results.notes)" class="pt-4">
-      <h3
+      <h2
```
paulkerschen, line 79: These headers should be `h2`, I think, since they sit next to (not under) student and class results in the hierarchy.

lyttam:
Check #2 seems stuck in the queue. I'll go ahead and merge this anyway.

PR #2624, opened 2020-10-14
BOAC-3746, buildspec: 'nvm use' before vue-cli build, in case wrong node version

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3746

I want to rule out shifting node versions as a cause.

### DIFF
--- a/buildspec.yml
+++ b/buildspec.yml
@@ -15,7 +15,8 @@ phases:
 
   build:
     commands:
-      - npm install -g @vue/cli
+      - node -v
+      - nvm use
       - npm run build-vue
 
   post_build:


buildspec.yml
```diff
@@ -15,7 +15,8 @@ phases:
 
   build:
     commands:
-      - npm install -g @vue/cli
+      - node -v
+      - nvm use
```
johncrossman, line 19: I removed the `install @vue/cli` because it shouldn't be necessary, and has risk in not matching version in our package.json. `npm install` (prior to build phase) will load dev dependencies, including vue-cli.

PR #2637, opened 2020-10-16
BOAC-3694 Voice graduated and inactive icons when not obvious from context

paulkerschen:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3694

### DIFF
--- a/src/components/student/StudentAvatar.vue
+++ b/src/components/student/StudentAvatar.vue
@@ -10,7 +10,7 @@
       @error="avatarError" />
     <div
       v-if="alertCount"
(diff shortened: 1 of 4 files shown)


src/components/student/StudentAvatar.vue
```diff
@@ -10,7 +10,7 @@
       @error="avatarError" />
     <div
       v-if="alertCount"
-      class="inactive-info-icon student-avatar-alert-count">
```
paulkerschen, line 13: I don’t know how this class snuck in here; it doesn’t belong.

johncrossman:
One of these things just doesn't belong here...

PR #2643, opened 2020-10-19
BOAC-3703: associates form labels with rich text editors

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3703

### DIFF
--- a/src/components/admin/EditServiceAnnouncement.vue
+++ b/src/components/admin/EditServiceAnnouncement.vue
@@ -22,8 +22,8 @@
           id="textarea-update-service-announcement"
           :initial-value="originalText"
           :disabled="isSaving"
-          :on-value-update="onEditorUpdate"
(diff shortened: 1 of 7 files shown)


src/components/admin/EditServiceAnnouncement.vue
```diff
@@ -22,8 +22,8 @@
           id="textarea-update-service-announcement"
           :initial-value="originalText"
           :disabled="isSaving"
-          :on-value-update="onEditorUpdate"
-          aria-label="Service announcement input" />
+          :label="'Service alert input'"
```
johncrossman, line 25: @lyttam - can the above (and those below) be simplified to
```
label="Service alert input"
```

lyttam: @johncrossman oops, yes. Amended. "We Simplify"

PR #2645, opened 2020-10-20
Added Transfer status to the CSV Export

cesarvh:
JIRA: https://jira.ets.berkeley.edu/jira/browse/BOAC-3546


cesarvh:
Debated with whether to fill in the null value with "No", but opted not to since the JIRA said it could just be a null value. Let me know if this needs to  change

lyttam:
@cesarvh I think 'Yes' or null makes the most sense, though one could run it by the service lead if one were so inclined. From a code standpoint this looks great.

PR #2646, opened 2020-10-20
NOJIRA, buildspec does not need sudo

johncrossman:


johncrossman:
Debugging the pipeline.  cc: @paulkerschen @lyttam

PR #2650, opened 2020-10-20
BOAC-1575, do not put aria-live on h1, h2, etc. NVDA will not recognize them as headers

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1575


paulkerschen:
@johncrossman Out `putFocusNextTick` method requires the target to have a `tabIndex` value. Does tabIndex on a header make it unrecognizable to NVDA? If so then we might need to put tabIndex on a surrounding container element.

johncrossman:
@paulkerschen - I'll have a look. Thanks.

PR #2652, opened 2020-10-21
BOAC-3713, replace 'b-form-select' with more accessible 'b-dropdown', styled

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3713

I'm done trying to shoe-horn accessibility into `b-form-select'`. This is a better approach, methinks.

image


paulkerschen:
b-form-select, the component that was just a little too helpful.

PR #2660, opened 2020-10-23
BOAC-3473, less sr confusion: use vue-a11y/announcer; no longer rely on Spinner for srAlerts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3473

Our home-cooked srAlert feature relies on vuex store and mixins. By replacing that with `@vue-a11y/announcer` we get less code and a couple nice features (eg, politeness control). Also, having the Spinner component manage "page has loaded" alerts is not elegant and often results in double-voicing alerts.

### DIFF
--- a/src/components/cohort/CohortHistory.vue
+++ b/src/components/cohort/CohortHistory.vue
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <SectionSpinner :loading="loading" name="History" />
+    <SectionSpinner :loading="loading" />
     <div v-if="!loading && totalEventsCount > itemsPerPage">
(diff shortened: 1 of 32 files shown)


src/components/cohort/CohortHistory.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <div>
-    <SectionSpinner :loading="loading" name="History" />
+    <SectionSpinner :loading="loading" />
```
johncrossman, line 3: The same goes for `SectionSpinner`.

lyttam:
Nice!

PR #2681, opened 2020-10-28
BOAC-1463, Course page: toggle tabs or page-size and expect proper sr alerts

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1463

This PR also benefits sighted users with a cleaner UI.


paulkerschen:
Everyone wins!

PR #2691, opened 2020-10-30
BOAC-3678, screen-reader users can dismiss/restore BOA service announcement

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3678

This button is sr-only but it's in the tab-order. Savvy sighted users can get at it.


paulkerschen:
This will definitely make the departure lounge a more relaxing place.

PR #2693, opened 2020-11-01
BOAC-3619: Student page now displays graduated minors

cesarvh:
Some  things to note:
- The data  for minors owners is  not there, so I can't  display it. How would this work  for dual degrees  + minors since they would be different colleges?
- Hyperlinks don't seem to be there for graduated students
- Had to hard code the "UG"
- I added some logic in  case people have multiple  minors

image

### DIFF
--- a/src/components/student/profile/StudentProfileHeader.vue
+++ b/src/components/student/profile/StudentProfileHeader.vue
@@ -92,7 +92,8 @@
               :active="true" />
           </div>
           <div v-if="plansMinorPartitionedByStatus[0].length" id="student-bio-minors" class="mb-3">
-            <h3 class="student-profile-section-header">Minor</h3>
+            <h3 v-if="plansMinorPartitionedByStatus.length > 1 " class="student-profile-section-header">Minors</h3>
+            <h3 v-if="plansMinorPartitionedByStatus.length === 1 " class="student-profile-section-header">Minor</h3>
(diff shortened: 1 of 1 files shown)


src/components/student/profile/StudentProfileHeader.vue
```diff
@@ -92,7 +92,8 @@
               :active="true" />
           </div>
           <div v-if="plansMinorPartitionedByStatus[0].length" id="student-bio-minors" class="mb-3">
-            <h3 class="student-profile-section-header">Minor</h3>
+            <h3 v-if="plansMinorPartitionedByStatus.length > 1 " class="student-profile-section-header">Minors</h3>
+            <h3 v-if="plansMinorPartitionedByStatus.length === 1 " class="student-profile-section-header">Minor</h3>
```
johncrossman, line 96: Looks good @cesarvh!  Minor amend needed: remove trailing spaces in `v-if` clause and extra line break below.

src/components/student/profile/StudentProfileHeader.vue
```diff
@@ -200,6 +211,7 @@ export default {
   },
   data: () => ({
     degreePlans: [],
+    minorPlans: [],
```
johncrossman, line 214: @cesarvh - Fyi, we like to keep lists of imports and var declarations alphabetized.

PR #2694, opened 2020-11-03
BOAC-1420, Context mixin for srAlert call

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-1420

We were getting "no such alertScreenReader() method" error in console.


scripts/db/migrate/2019/20190924-BOAC-2620/post_deploy_01_gpa_json_in_cohort_filter.py
```diff
@@ -34,7 +34,7 @@
 
 
 @scriptify.in_app  # noqa: C901
-def main(app):
+def main(app):  # noqa: C901
```
johncrossman, line 37: I updated flake8 to match Nessie. Oddly, Travis insisted on having `noqa` on both method declaration and the annotation.

PR #2724, opened 2020-11-20
BOAC-2909: Added cohort filter for students with holds

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2909


boac/merged/cohort_filter_options.py
```diff
@@ -90,6 +90,7 @@ def get_all_filter_categories(self):
                 _boolean_filter('midpointDeficient', 'Midpoint Deficient Grade'),
                 _boolean_filter('transfer', 'Transfer Student'),
                 _filter('unitRanges', 'Units Completed', options=unit_range_options),
+                _boolean_filter('studentHolds', 'Holds'),
```
johncrossman, line 93: @cesarvh This all looks good. Nice work! One amendment needed: The "Holds" option above should be placed in alpha order, just before the "Intended Major" option.

cesarvh: @johncrossman Done!

PR #2729, opened 2020-12-10
BOAC-3827: Divides student enrl terms into expandable sections by year

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3827


johncrossman:
Mmmm, kebabCase.

PR #2731, opened 2020-12-11
BOAC-3596, UX of alerts export, with v-calendar update

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3596

Date picker with smart min/max logic for date range.

### DIFF
--- a/src/components/admin/AlertsLogExport.vue
+++ b/src/components/admin/AlertsLogExport.vue
@@ -1,11 +1,89 @@
 <template>
   <div>
-    AlertsLogExport
+    <div class="text-muted">Date Range</div>
+    <b-form @submit="onSubmit">
+      <b-form-group>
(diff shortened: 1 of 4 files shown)


src/components/admin/AlertsLogExport.vue
```diff
@@ -1,11 +1,89 @@
...
+    minDate: new Date('01/01/2014'),
+    toDate: undefined
+  }),
+  methods: {
+    dateFormat(value) {
+      const parsed = Date.parse(value)
+      return isNaN(parsed) ? null : this.$options.filters.moment(parsed, this.dateInputFormat)
+    },
+    onSubmit() {
+      console.log('Download CSV')
```
johncrossman, line 79: Wiring this method to the API call will be final step.

paulkerschen:
I’m sold! I’ll pick that date!

PR #2737, opened 2020-12-14
BOAC-3802, search form accessibility: Inline button, source order change

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3802

* The diff might be more readable with `?w`
* `bootstrap-vue` fixed an oddity with form-checkbox. I updated our usages of that component.


src/components/cohort/FilterRow.vue
```diff
@@ -17,6 +17,7 @@
         id="new-filter-button"
         :aria-labelledby="`new-filter-${index}-label`"
         class="h-100"
+        menu-class="min-width-100"
         no-caret
         toggle-class="dd-override"
         variant="link"
```
johncrossman, line 23: List of b-dropdown-items should be at least as wide as the b-dropdown element it lives in.

PR #2745, opened 2020-12-17
BOAC-3831: Added sort academic year functionality to student classes ui

cesarvh:
JIRA: https://jira.ets.berkeley.edu/jira/browse/BOAC-3831

Design: https://jira.ets.berkeley.edu/jira/browse/BOAC-3734

### DIFF
--- a/src/components/student/profile/StudentClasses.vue
+++ b/src/components/student/profile/StudentClasses.vue
@@ -1,6 +1,24 @@
 <template>
   <div id="student-terms-container" class="m-3">
-    <h2 class="student-section-header">Classes</h2>
+    <div>
+      <div class="d-inline mr-3">
+        <h2 class=" d-inline-block student-section-header">Classes </h2>
+      </div>
+      <div class="d-inline">
+        <b-link
+          id="sort-academic-year"
+          @click="setOrder"
+        > 
+          Sort academic year
+          <span v-if="currentOrder === 'asc' ">
+            <font-awesome icon="long-arrow-alt-down" />
+          </span>
+          <span v-if="currentOrder === 'desc' ">
+            <font-awesome icon="long-arrow-alt-up" />
+          </span>
+        </b-link>
+      </div>
+    </div>
     <div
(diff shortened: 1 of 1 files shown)


src/components/student/profile/StudentClasses.vue
```diff
@@ -1,6 +1,24 @@
...
+        > 
+          <span />
+          Sort academic year
+          <span v-if="currentOrder === 'asc' ">
+            <font-awesome icon="long-arrow-alt-down" />
+          </span>
+          <span v-if="currentOrder === 'desc' ">
+            <font-awesome icon="long-arrow-alt-up" />
+          </span>
+        </b-link>
```
johncrossman, line 19: @cesarvh - PR looks good. I have a few comments/requests:
- Do not use `style` attribute. Very often you can use a [Bootstrap CSS util](https://bootstrap-vue.org/docs/reference/utility-classes). Instead of `style="display:inline"` you would use `class="d-inline"`.  There is also `d-inline-block`.  
- Remove the stray `<span />`
- Put an id attribute on the `b-link` and the QA dept will thank you.
- In the `setOrder()` method below, use `alertScreenReader(...)` to notify screenreader of order change. (You'll see examples in other components.)

Thanks!

src/components/student/profile/StudentClasses.vue
```diff
@@ -312,15 +330,24 @@ export default {
...
+    currentOrder: undefined
+  }),
+  created () {
+    this.currentOrder = 'desc'
+  },
   computed: {
     enrollmentTermsByYear: function() {
-      return this.$_.groupBy(this.student.enrollmentTerms, 'academicYear')
-    }
+      let orderedTerms = this.$_.orderBy(this.student.enrollmentTerms, 'termId', this.currentOrder)
```
lyttam, line 341: @cesarvh I think we want to preserve the order of the terms within each academicYear (fall, spring, summer) and sort by the academicYear itself. Should require simply switching the second argument to `academicYear`.

lyttam:
Thanks @cesarvh , nice work!

PR #2746, opened 2020-12-17
BOAC-3596, bootstrap input wrapped in v-date-picker -> breakage

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3596

For reasons of accessibility, we might want to drop `v-calendar` and use Bootstrap's https://bootstrap-vue.org/docs/components/calendar, which calls out accessibility and keyboard-only features.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -22902,9 +22902,9 @@
       "dev": true
     },
     "v-calendar": {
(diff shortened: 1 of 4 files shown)


paulkerschen:
No bees in the vees.

PR #2755, opened 2020-12-22
BOAC-3823: Fixed intended majors export

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3823

PR #2760, opened 2021-01-07
BOAC-3830: Adds class detail section to new classes view

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3830


johncrossman:
Thanks @lyttam  Let's give it a go.

PR #2766, opened 2021-01-09
BOAC-3852, if only one filter-option-group then flatten to simple array of options

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3852

See recent comments in da Jira.


paulkerschen:
Sold!

PR #2768, opened 2021-01-11
BOAC-3743, sensible and legit ids for filter select options

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3743


paulkerschen:
2L2Q.

PR #2772, opened 2021-01-12
BOAC-3861, when edit cohort filter, flatten opt-groups to find selected option

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3861

Includes a bit of cleanup to slim down the bloated `FilterRow` component.

### DIFF
--- a/src/components/cohort/FilterRow.vue
+++ b/src/components/cohort/FilterRow.vue
@@ -16,7 +16,7 @@
       class="filter-row-column-01 mt-1 pr-2"
     >
       <FilterSelect
-        :filter-row-index="filterRowIndex"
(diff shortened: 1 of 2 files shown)


src/components/cohort/FilterRow.vue
```diff
@@ -16,7 +16,7 @@
       class="filter-row-column-01 mt-1 pr-2"
     >
       <FilterSelect
-        :filter-row-index="filterRowIndex"
+        :filter-row-index="position"
```
johncrossman, line 19: `filterRowIndex` was computed value based on `position`. With this PR, `position` has the proper default and `filterRowIndex` was dropped.

PR #2776, opened 2021-01-13
BOAC-3828, BOAC-3842, BOAC-3837: Displays relevant empty terms

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3828
https://jira.ets.berkeley.edu/jira/browse/BOAC-3842
https://jira.ets.berkeley.edu/jira/browse/BOAC-3837

For every academic year in which the student is enrolled in at least one term, all three terms (Fall, Spring, and Summer) will be shown.  The current academic year will be expanded by default.

### DIFF
--- a/boac/merged/student.py
+++ b/boac/merged/student.py
@@ -785,28 +785,14 @@ def _merge_enrollment_terms(profile, enrollment_results, academic_standing=None)
         else:
             # Omit dropped sections for non-current terms.
             term.pop('droppedSections', None)
+            if term_id < current_term_id():
+                # Filter out the now-empty term if all classes were dropped.
+                if not term.get('enrollments'):
+                    continue
+                _omit_zombie_waitlisted_enrollments(term)
 
-            # Filter out the now-empty term if all classes were dropped.
-            enrollments = term.get('enrollments')
-            if not enrollments:
-                continue
-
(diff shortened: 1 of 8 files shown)


boac/merged/student.py
```diff
@@ -785,28 +785,14 @@ def _merge_enrollment_terms(profile, enrollment_results, academic_standing=None)
         else:
             # Omit dropped sections for non-current terms.
             term.pop('droppedSections', None)
+            if term_id < current_term_id():
+                # Filter out the now-empty term if all classes were dropped.
+                if not term.get('enrollments'):
+                    continue
+                _omit_zombie_waitlisted_enrollments(term)
```
johncrossman, line 792: @lyttam - minor observation: the above can be simplified. 
```
if term.get(...): 
  _omit_zombie...()
```

lyttam: @johncrossman Good suggestion; however we would still need the `continue` to break out of the `for` loop if there are no enrollments. I have a mild preference for the existing code over the following alternative because the former seems more readable to me. If you have stronger feelings, I'll make the change.

```if term.get('enrollments'):
    _omit_zombie_waitlisted_enrollments(term)
else:
    continue```

PR #2778, opened 2021-01-13
BOAC-3841: flips the academic year sort-by icon

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3841


paulkerschen:
ascending-descending

PR #2779, opened 2021-01-13
BOAC-3881, ModalHeader gets optional container-class prop

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3881

E.g., if you don't want bottom-border.


paulkerschen:
Actually, @johncrossman, having just merged, I'm suddenly wondering if Vue does out of the box what you did with your container-class prop. What happens if you just put `class="border-bottom-0"` on the ModalHeader without explicit handling? I think HTML attributes on a component might get evaluated onto the top element in that component by default.

johncrossman:
@paulkerschen - I'll experiment.

PR #2784, opened 2021-01-15
BOAC-3885, search-input re-uses StudentPersonalDetails transition CSS

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3885

Disabled search input gets custom bg-color, as in previous impl.

### DIFF
--- a/src/components/search/InputAutocomplete.vue
+++ b/src/components/search/InputAutocomplete.vue
@@ -3,6 +3,8 @@
     :id="id"
     :aria-labelledby="ariaLabelledby"
     :aria-required="required"
+    :base-class="disabled ? 'disabled' : 'autocomplete'"
+    :disabled="disabled"
     :placeholder="placeholder"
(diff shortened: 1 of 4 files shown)


src/components/search/InputAutocomplete.vue
```diff
@@ -84,4 +90,16 @@ export default {
...
+  background-color: #ddd;
+  background-image: none;
+  border: 2px solid #ccc;
+  border-radius: 4px;
+	box-sizing: border-box;
+  color: #333;
+  font-size: 16px;
+  height: 45px;
+  padding: 0 10px 0 10px;
+	width: 100%;
```
lyttam, line 103: @johncrossman Extra indent on lines 98 and 103?

johncrossman: Good catch @lyttam  Apparently, tabs snuck in. I've amended.

PR #2786, opened 2021-01-15
BOAC-3819, no more concurrent jobs in travis.yml; use tox --parallel

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3819

Travis concurrency is not free and `tox --parallel` is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,23 +12,14 @@ install:
   - npm install
   - pip3 install tox
 
-jobs:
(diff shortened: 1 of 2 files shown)


johncrossman:
Previous builds have taken 10-11 min. Let's see how this one does.

johncrossman:
7 min 28 sec. We pay less for faster builds.

PR #2789, opened 2021-01-21
BOAC-3771, screen-reader speaks error when create curated group w/ bulk SIDs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3771


paulkerschen:
I’m sorry, Dave, I can’t bulk add those SIDs.

PR #2790, opened 2021-01-21
BOAC-3809, break up AcademicTimeline component prior to accessibility  fix

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3809

`AcademicTimeline.vue` was nearing 1000 lines, surpassing maintainability. Let's break it up into logical chunks.


paulkerschen:
Nothing to be ashamed of, AcademicTimeline, most components go up a pants size as they get older.

PR #2797, opened 2021-01-25
BOAC-3889: Added CPN to EPN filter

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3889

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -1096,7 +1096,8 @@ def get_students_query(     # noqa
         query_bindings.update({'coe_advisor_ldap_uids': coe_advisor_ldap_uids})
     if coe_epn:
         query_tables += ' JOIN student.student_enrollment_terms e ON e.sid=sas.sid '
-        query_filter += ' AND e.enrollment_term LIKE \'%gradingBasis": "EPN"%\' AND e.term_id = ANY(:coe_epn)'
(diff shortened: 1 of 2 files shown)


boac/externals/data_loch.py
```diff
@@ -1096,7 +1096,8 @@ def get_students_query(     # noqa
         query_bindings.update({'coe_advisor_ldap_uids': coe_advisor_ldap_uids})
     if coe_epn:
         query_tables += ' JOIN student.student_enrollment_terms e ON e.sid=sas.sid '
-        query_filter += ' AND e.enrollment_term LIKE \'%gradingBasis": "EPN"%\' AND e.term_id = ANY(:coe_epn)'
+        query_filter += ' AND e.enrollment_term LIKE \'%gradingBasis": "EPN"%\' '
+        query_filter += ' OR e.enrollment_term LIKE \'%gradingBasis": "CPN"%\' AND e.term_id = ANY(:coe_epn)'
```
paulkerschen, line 1100: @cesarvh How should precedence work with this query? It’s not always easy to predict how multiple statements combined with AND and OR will get evaluated with SQL, so you might want to add parentheses around the parts that should get evaluated first.

cesarvh: Ah, I was afraid this might be the case. Thanks Paul!

PR #2803, opened 2021-01-26
BOAC-3899: CSV export now checks for students without intended majors

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3899

PR #2805, opened 2021-01-27
BOAC-3908: Student page now displays GPA/Units as m-dash instead of 0…

cesarvh:
….000/0


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3908

PR #2812, opened 2021-01-28
BOAC-3914, FDR is for admins _and_ directors 

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3914


paulkerschen:
Every good director deserves FDR.

PR #2833, opened 2021-03-02
BOAC-3930: Adds degree check feature flag and view

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3930


johncrossman:
@lyttam - The PR looks good. Travis is unhappy and it's not clear why. In https://jira.ets.berkeley.edu/jira/browse/BOAC-3929 I anticipated a certain problem but I'm not sure it matches this error.

PR #2837, opened 2021-03-04
BOAC-3926: Restructured admitted student page

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3926

PR #2838, opened 2021-03-04
BOAC-3857:Sort by academic year displays only if > 1 year, not > 1 term

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3857

pkerschen:
Thanks @cesarvh!

PR #2847, opened 2021-03-11
BOAC-3943: Adjusted header font sizes + padding of cells

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3943

PR #2856, opened 2021-03-30
BOAC-3961: Models a unit requirement belonging to a degree template

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3961


scripts/db/schema.sql
```diff
@@ -374,6 +374,32 @@ ALTER TABLE ONLY degree_progress_templates
...
+    CACHE 1;
+ALTER SEQUENCE degree_progress_unit_requirements_id_seq OWNER TO boac;
+ALTER SEQUENCE degree_progress_unit_requirements_id_seq OWNED BY degree_progress_unit_requirements.id;
+ALTER TABLE ONLY degree_progress_unit_requirements ALTER COLUMN id SET DEFAULT nextval('degree_progress_unit_requirements_id_seq'::regclass);
+ALTER TABLE ONLY degree_progress_unit_requirements
+    ADD CONSTRAINT degree_progress_unit_requirements_pkey PRIMARY KEY (id);
+ALTER TABLE ONLY degree_progress_unit_requirements
+    ADD CONSTRAINT degree_progress_unit_requirements_template_id_fkey FOREIGN KEY (template_id) REFERENCES degree_progress_templates(id) ON DELETE CASCADE;
+ALTER TABLE ONLY degree_progress_unit_requirements
+    ADD CONSTRAINT degree_progress_unit_requirements_name_template_id_unique_const UNIQUE (name, template_id);
```
johncrossman, line 398: @lyttam - Do we need `degree_progress_unit_requirements_name_template_id_unique_const` in the db-modify script?

boac/models/degree_progress_unit_requirement.py
```diff
@@ -0,0 +1,66 @@
...
+
+
+class DegreeProgressUnitRequirement(db.Model):
+    __tablename__ = 'degree_progress_unit_requirements'
+
+    id = db.Column(db.Integer, nullable=False, primary_key=True)  # noqa: A003
+    min_units = db.Column(db.Integer, nullable=False)
+    name = db.Column(db.String(255), nullable=False)
+    template_id = db.Column(db.Integer, db.ForeignKey('degree_progress_templates.id'), nullable=False)
+    template = db.relationship('DegreeProgressTemplate', back_populates='unit_requirements')
```
johncrossman, line 37: @lyttam - Curious: why no `created_at` and `updated_at` in this table?

johncrossman:
Looks good. Posted a couple questions.

boac/models/degree_progress_unit_requirement.py
lyttam: @johncrossman The minimum requirement is to keep track of those in the parent model `DegreeProgressTemplate` so that when a unit requirement is edited, the timestamp will be reflected in `template.updated_at`.  Though I am concerned that's going to be harder to implement than each model having its own `updated_at`.

scripts/db/schema.sql
lyttam: You're right; I forgot to add it there.

boac/models/degree_progress_unit_requirement.py
lyttam: I think we'll want those two columns, plus `created_by` and `updated_by` since unit requirements can be added and edited independent of the parent template. Adding them now.

lyttam:
@johncrossman Amended.

PR #2858, opened 2021-04-02
BOAC-3935: Added API for initializing degree

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3935

### DIFF
--- a/boac/models/degree_progress_template.py
+++ b/boac/models/degree_progress_template.py
@@ -87,3 +87,12 @@ class DegreeProgressTemplate(Base):
             cls.deleted_at == None,  # noqa: E711
         )
         return cls.query.filter(criterion).order_by(cls.created_at).all()
+
+    def to_api_json(self):
+        return {
+            'id': self.id,
+            'degree_name': self.degree_name,
+            'advisor_dept_codes': self.advisor_dept_codes,
+            'created_by': self.created_by,
+            'updated_by': self.updated_by,
+        }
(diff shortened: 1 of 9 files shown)


cesarvh:
Hey Anne, I made a few changes to how the Degree page is laid out. Instead of creating a separate view, I'm using conditional rendering to render different components depending on if the degree exists or if its being made.

lyttam:
That's great @cesarvh, your single-view changes should merge nicely with what I'm currently working on for unit requirements.

boac/models/degree_progress_unit_requirement.py
```diff
@@ -25,8 +25,6 @@
 
 from boac import db, std_commit
 from boac.models.base import Base
-from boac.models.degree_progress_template import DegreeProgressTemplate
```
lyttam, line 28: Did you mean to remove this import? The class is referenced on line 72.

boac/models/degree_progress_template.py
```diff
@@ -24,11 +24,11 @@
 """
 
 from boac import db, std_commit
+from boac.models.degree_progress_unit_requirement import DegreeProgressUnitRequirement
```
lyttam, line 27: Did you mean to add this import? The class itself isn't referenced, just the string name on line 43.

src/api/cohort.ts
```diff
@@ -16,7 +16,8 @@ export function createCohort(
     })
     .then(response => {
       const cohort = response.data
-      store.commit('currentUserExtras/cohortCreated', cohort)
+      // TO DO: Store the degrees once we're able to retrieve them
+      // store.commit('currentUserExtras/cohortCreated', cohort)
```
lyttam, line 20: Did you mean to put this in degree.ts instead?

src/views/CreateDegreeTemplate.vue
```diff
@@ -0,0 +1,68 @@
...
+      this.degreeId = this.$_.get(this.$route, 'params.id') || ''
+    } else {
+      this.$router.push({path: '/404'})
+    }
+  }
+}
+</script>
+
+<style>
+#label-of-create-degree-input {
```
lyttam, line 64: I think we prefer a CSS class name here, instead of an ID.

PR #2863, opened 2021-04-06
BOAC-3966, protect degree-progress features per user.canEdit / user.canRead

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3966

With reliable server-side auth logic we can simplify client-side.


pauline2k:
can_read

PR #2865, opened 2021-04-06
BOAC-3985: Lost exclamation alert icon next to grades is back!

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3985

johncrossman:
Thanks @cesarvh

PR #2870, opened 2021-04-07
BOAC-3964: API for creating a unit requirement

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3964

### DIFF
--- a/boac/api/degree_progress_controller.py
+++ b/boac/api/degree_progress_controller.py
@@ -23,12 +23,13 @@ SOFTWARE AND ACCOMPANYING DOCUMENTATION, IF ANY, PROVIDED HEREUNDER IS PROVIDED
 ENHANCEMENTS, OR MODIFICATIONS.
 """
 
-from boac.api.errors import BadRequestError
+from boac.api.errors import BadRequestError, ResourceNotFoundError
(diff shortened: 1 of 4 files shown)


boac/api/degree_progress_controller.py
```diff
@@ -46,3 +47,25 @@ def create_degree():
         advisor_dept_codes=dept_codes_where_advising(current_user),
     )
     return tolerant_jsonify(degree.to_api_json())
+
+
+@app.route('/api/degree_progress/<template_id>/unit_requirement', methods=['POST'])
```
johncrossman, line 52: @lyttam  - What do you think about `/api/degree/...` as the API prefix?  That would match the other, and it's concise.

johncrossman:
Makes sense. Posted one question.

boac/api/degree_progress_controller.py
lyttam: @johncrossman I don't have strong feelings either way. `/api/degree_progress/...` matches the controller name, but concise is nice. Should we change the controller name to match, i.e. `degree_controller.py`?

johncrossman: IMO, the controller name can remain as is. Thanks @lyttam

PR #2875, opened 2021-04-07
BOAC-4015 Adjust spacing in curated group header

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4015


johncrossman:
Travis never ran here but the QA PR is green so...

PR #2879, opened 2021-04-07
BOAC-4012, rename degree on /degrees page (list view)

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4012

Design Jira is linked to BOAC-4012


johncrossman:
cc: @lyttam @pauline2k @cesarvh

PR #2887, opened 2021-04-08
BOAC-4018, proper putFocus after CRUD operatoins on /degrees

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4018

And removed an extraneous `getTemplate` on server-side.

### DIFF
--- a/boac/api/degree_progress_controller.py
+++ b/boac/api/degree_progress_controller.py
@@ -122,7 +122,6 @@ def update_degree_template(template_id):
 def _validate_template_upsert(name, template_id=None):
     if not name:
         raise BadRequestError('\'name\' is required.')
-    template = _get_degree_template(template_id) if template_id else None
     # Name must be unique across non-deleted templates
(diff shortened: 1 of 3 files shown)


boac/api/degree_progress_controller.py
```diff
@@ -122,7 +122,6 @@ def update_degree_template(template_id):
 def _validate_template_upsert(name, template_id=None):
     if not name:
         raise BadRequestError('\'name\' is required.')
-    template = _get_degree_template(template_id) if template_id else None
```
lyttam, line 125: @johncrossman I believe I left this here because `_get_degree_template` is doing some validation. It does feel like a code smell though.

johncrossman: @lyttam - The line just after this one has `template = DegreeProgressTemplate.find_by_name` which is what's used for the validation. However, `template_id` is used below.

lyttam: Makes sense, thanks!

PR #2889, opened 2021-04-09
BOAC-4000, degree_requirement_categories: table, model

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4000


boac/models/degree_progress_category.py
```diff
@@ -0,0 +1,146 @@
...
+class DegreeProgressCategory(Base):
+    __tablename__ = 'degree_progress_categories'
+
+    id = db.Column(db.Integer, nullable=False, primary_key=True)  # noqa: A003
+    category_type = db.Column(degree_progress_category_type, nullable=False)
+    description = db.Column(db.Text)
+    name = db.Column(db.String(255), nullable=False)
+    parent_category_id = db.Column(db.Integer, db.ForeignKey('degree_progress_categories.id'))
+    position = db.Column(db.Integer, nullable=False)
+    template_id = db.Column(db.Integer, db.ForeignKey('degree_progress_templates.id'), nullable=False)
```
johncrossman, line 50: @lyttam  - Feedback welcome on (1) use of `degree_progress_category_type` enum, and (2) column name 'position'.

johncrossman:
@lyttam - I need to amend this PR to add db-modify script. I'll ping you when it's ready.

boac/models/degree_progress_category.py
lyttam: @johncrossman I like the column name 'position'.

When we met to review the schema, someone pointed out that an enum isn't necessary because the presence of a `parent_category_id` indicates a subcategory. However, it looks like you're planning to have course requirements in this table alongside categories and subcategories. If we do that, then technically the presence of `units` could be used to differentiate courses from categories - but maybe the enum would make the distinction more clear.

johncrossman:
Ready for review.

johncrossman:
@lyttam  - Amendment is in.

boac/models/degree_progress_category.py
```diff
@@ -0,0 +1,153 @@
...
+                    template_id={self.template_id},
+                    created_at={self.created_at},
+                    updated_at={self.updated_at}>"""
+
+    @classmethod
+    def create(
+            cls,
+            category_type,
+            name,
+            parent_category_id,
```
lyttam, line 88: @johncrossman `parent_category_id` should be an optional parameter too, right? Default to None when a top-level category is being created.

johncrossman: Yes, good catch.

johncrossman: I'll amend once more when your review is done.

lyttam:
@johncrossman Looks great, just the one comment about `parent_category_id`.

johncrossman:
@lyttam  Amended and green.

PR #2890, opened 2021-04-09
BOAC-4001, /degree/:id page pulls categories and organizes per 'position'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4002

Includes minor fix in db-migrate script. (It is this version of the script that was run on `boa-dev`.)


lyttam:
Thanks for fixing all those broken Promises.

PR #2896, opened 2021-04-12
BOAC-4027, degree progress: input 'units' and 'reqt fulfillments' when creating a course

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4027

### DIFF
--- a/src/components/degree/AddDegreeCategory.vue
+++ b/src/components/degree/AddDegreeCategory.vue
@@ -38,48 +38,72 @@
       </div>
       <div v-if="selectedCategoryType === 'Course'">
         <div class="font-weight-500 my-2">
-          Requirement Fulfillment
+          Units
         </div>
         <div>
-          <b-select
-            :id="`column-${position}-requirement-fulfillment-select`"
+          <b-form-input
+            :id="`column-${position}-course-units-input`"
+            v-model="courseUnits"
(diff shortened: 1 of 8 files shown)


src/components/degree/AddDegreeCategory.vue
```diff
@@ -38,48 +38,72 @@
...
+            v-model="courseUnits"
+            class="course-units-input"
             :disabled="isSaving"
-            @change="onChangeRequirementFulfillment"
+            maxlength="2"
+            trim
+          />
+          <span v-if="!isValidCourseUnits" class="has-error faint-text font-size-12">
+            Number required
+          </span>
```
lyttam, line 54: @johncrossman In EditUnitRequirements.vue, I used `type="number"` on the units input field so that non-number characters can't be entered. Wondering if you would like to use that attribute here, or if it's a good idea at all (I haven't tested it for accessibility yet).

lyttam: (On devices without a physical keyboard, I think it will bring up a number pad virtual keyboard instead of the full querty virtual keyboard)

johncrossman: Thanks @lyttam  That's a good tip. I'll amend this PR if it works locally.

PR #2898, opened 2021-04-13
BOAC-4023, categories.to_api_json struct suitable to front-end rendering needs

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4023

IMO, more intuitive and cleaner.


pfarestveit:
Looks gorgeous to the QA Dept, @johncrossman

PR #2903, opened 2021-04-13
BOAC-4025, table.td need its own flavor of truncate-with-ellipsis

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4025

See https://stackoverflow.com/questions/9789723/css-text-overflow-in-a-table-cell


pauline2k:
Educational!

PR #2904, opened 2021-04-14
BOAC-4030: Fixes element id on unit reqt cancel button

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4030


src/components/degree/EditUnitRequirement.vue
```diff
@@ -100,10 +100,12 @@ export default {
       this.reset()
     },
     cancel() {
+      this.$announcer.polite('Canceled.')
```
pauline2k, line 103: The announcer speaks American English.

PR #2923, opened 2021-04-16
BOAC-4011: Initial version of printable degree

cesarvh:
Design: https://jira.ets.berkeley.edu/jira/browse/BOAC-3880

JIRA: https://jira.ets.berkeley.edu/jira/browse/BOAC-4011

### DIFF
--- a/src/router.ts
+++ b/src/router.ts
@@ -21,6 +21,7 @@ import Home from '@/views/Home.vue'
 import Login from '@/layouts/Login.vue'
 import NotFound from '@/views/NotFound.vue'
 import PassengerManifest from '@/views/PassengerManifest.vue'
+import PrintableDegreeTemplate from '@/views/degree/PrintableDegreeTemplate.vue'
 import Profile from '@/views/Profile.vue'
 import Router from 'vue-router'
 import Search from '@/views/Search.vue'
@@ -279,6 +280,13 @@ const router = new Router({
           meta: {
             title: 'Student Degree',
           },
+        },
+        {
+          path: '/degree/print/:id',
+          component: PrintableDegreeTemplate,
+          meta: {
+            title: 'Print Degree Template',
+          }
         }
       ]
     },
--- a/src/views/degree/ManageDegreeChecks.vue
+++ b/src/views/degree/ManageDegreeChecks.vue
@@ -109,15 +109,14 @@
             <div class="align-right w-100">
               <div v-if="row.item.id !== $_.get(templateForEdit, 'id')" class="d-flex">
                 <div>
-                  <b-btn
-                    :id="`degree-check-${row.index}-print-btn`"
+                  <router-link
+                    :id="`degree-check-${row.item.id}-print-link`"
                     class="p-1"
                     :disabled="isBusy"
-                    variant="link"
-                    @click="print(row.item.id)"
+                    :to="`degree/print/${row.item.id}`"
                   >
                     Print
-                  </b-btn>
+                  </router-link>
                 </div>
                 <div v-if="$currentUser.canEditDegreeProgress">
                   <span class="separator">|</span>
--- a/src/views/degree/PrintableDegreeTemplate.vue
+++ b/src/views/degree/PrintableDegreeTemplate.vue
@@ -1,0 +1,126 @@
+<template>
+  <div class="ml-3 mr-3 mt-3">
+    <Spinner />
+    <div v-if="!loading">
+      <b-container>
+        <b-row>
+          <b-col class="pl-0">
+            <b-row>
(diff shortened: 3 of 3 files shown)


src/router.ts
```diff
@@ -279,7 +280,14 @@ const router = new Router({
           meta: {
             title: 'Student Degree',
           },
-        }
+        },
+        {
+          path: 'degree/print/:id',
```
johncrossman, line 285: @cesarvh - Leading slash needed. And trailing commas in `meta` (below) are unnecessary.

src/views/degree/ManageDegreeChecks.vue
```diff
@@ -109,15 +109,13 @@
...
                     class="p-1"
                     :disabled="isBusy"
-                    variant="link"
-                    @click="print(row.item.id)"
-                  >
-                    Print
-                  </b-btn>
+                    :to="`degree/print/${row.item.id}`"
+                    v-html="`Print`"
+                  />
```
johncrossman, line 118: @cesarvh - We should use `v-html` only when necessary. In this case, it ain't necessary. :)

src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -0,0 +1,143 @@
...
+    Spinner,
+  },
+  mixins: [Context, DegreeEditSession, Loading, Util],
+  mounted() {
+    const id = this.toInt(this.$_.get(this.$route, 'params.id'))
+    this.init(id).then(() => {
+      this.setPageTitle(this.degreeName)
+      this.loaded(this.getLoadedAlert())
+      this.putFocusNextTick('add-unit-requirement')
+    })
```
johncrossman, line 60: @cesarvh - For this page you should not need `DegreeEditSession` because this is a read-only view. I recommend you replace `init()` above with standard API call: `getDegreeTemplate()`

johncrossman: @cesarvh – Also, I don't think `add-unit-requirement` is an element id on this page.

src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -0,0 +1,143 @@
...
+      }
+    ],
+  }),
+  methods: {
+    getLoadedAlert() {
+      if (!this.templateId) {
+        return 'Create degree page has loaded'
+      } else {
+        return `Degree ${this.degreeName || ''} has loaded`
+      }
```
johncrossman, line 86: @cesarvh - `templateId` should never be null in `/print` view, right?  I think a screen-reader alert such as, "Printable {degree.name} has loaded", would be sufficient.

src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -0,0 +1,143 @@
...
+
+#print-unit-requirements-table > thead {
+  font-size: 8px;
+}
+
+#print-unit-requirements-table {
+  font-size: 8px;
+}
+
+</style>
```
johncrossman, line 142: @cesarvh - A minor concern: I don't think you need two line breaks between style entries.

cesarvh:
@johncrossman  Fixed!

src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -0,0 +1,137 @@
...
+  mixins: [Context, Loading, Util],
+  mounted() {
+    const id = this.toInt(this.$_.get(this.$route, 'params.id'))
+
+    getDegreeTemplate(id).then(data => {
+      this.templateId = data.id
+      this.degreeName = data.name
+      this.unitRequirements = data.unitRequirements
+      this.sid = data.sid
+      this.loaded(this.getLoadedAlert())
```
johncrossman, line 62: @cesarvh - Thanks for ditching `DegreeEditSession`. Now I recommend having a single `template` variable (instead of multiple: name, id, sid, etc.).  This page will be rendering notes and categories soon and we don't want to add a new var for each.

src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -0,0 +1,137 @@
...
+      }
+    ],
+    templateId: undefined,
+    degreeName: undefined,
+    unitRequirements: undefined,
+    sid: undefined
+  }),
+  methods: {
+    getLoadedAlert() {
+      `Printable ${this.degreeName} has loaded`
```
johncrossman, line 89: @cesarvh - This is a one-liner string that probably doesn't need to be in a method.

johncrossman:
Thank you @cesarvh !

PR #2928, opened 2021-04-19
BOAC-4058: Added categories, subcategories, courses to print view

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4058

### DIFF
--- a/src/views/degree/PrintableDegreeTemplate.vue
+++ b/src/views/degree/PrintableDegreeTemplate.vue
@@ -2,13 +2,14 @@
   <div class="ml-3 mr-3 mt-3">
     <Spinner />
     <div v-if="!loading">
-      <b-container>
+      <b-container fluid>
         <b-row>
           <b-col class="pl-0">
             <b-row>
               Student Information
             </b-row>
           </b-col>
+          <b-col />
           <b-col>
             <div class="unofficial-label-pill">UNOFFICIAL DEGREE PROGRESS REPORT </div>
             <h1 class="print-degree-name pt-2">{{ template.name }}</h1>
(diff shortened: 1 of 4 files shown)


src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -82,6 +136,14 @@ export default {
 </script>
 
 <style>
```
lyttam, line 138: @cesarvh If you add the `scoped` attribute, the styles will apply only to this component (they are global by default).  e.g. `<style scoped>`. I think that would be desirable since the printable template needs its own set of styles, right?

lyttam:
Sorry, I merged #2929 and caused a conflict.

src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -32,7 +33,56 @@
...
+            <template>
+              <div>
+                <div
+                  v-for="category in $_.filter(template.categories, c => c.position === position && $_.isNil(c.parentCategoryId))"
+                  :key="category.id"
+                >
+                  <Category
+                    v-if="category.id"
+                    :category="category"
+                    :on-click-edit="edit"
```
johncrossman, line 52: @cesarvh - In printable view, you should not set `on-click-edit`

cesarvh:
@lyttam Went ahead and added the scoped attribute. If I add it, on my local machine all of the style goes away, but it seems that just might be the loader being buggy, according to stackoverflow. Once its up on dev we can make sure that the styling is working as desired

lyttam:
Thanks, looks good.

PR #2930, opened 2021-04-19
BOAC-4033: Allows category course units to be a numerical range

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4033

I opted not to add validation to ensure the 'min' is less than the 'max' - if a user decides to input '4-3', for example, that's on them.

### DIFF
--- a/boac/models/degree_progress_category.py
+++ b/boac/models/degree_progress_category.py
@@ -27,7 +27,8 @@ from boac import db, std_commit
 from boac.models.base import Base
 from boac.models.degree_progress_course_unit_requirement import DegreeProgressCourseUnitRequirement
 from dateutil.tz import tzutc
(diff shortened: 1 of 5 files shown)


johncrossman:
@lyttam - What do you think of using Postgres builtin range type? https://medium.com/dataseries/range-types-in-postgresql-and-gist-indexes-788db23346c5   That would allow data integrity checks and help with the comparisons we'll need to do https://jira.ets.berkeley.edu/jira/browse/BOAC-3872   Thoughts?

lyttam:
Good idea @johncrossman, I'll amend.

lyttam:
@johncrossman I pushed some changes. Ready for another look.

johncrossman:
Thank you @lyttam  Looks good. I'll merge when Travis gives the green.

PR #2934, opened 2021-04-20
BOAC-4077, UI tweaks to work-in-progress /unassigned/courses

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4077

### DIFF
--- a/src/components/degree/student/UnassignedCourses.vue
+++ b/src/components/degree/student/UnassignedCourses.vue
@@ -24,18 +24,20 @@
         </b-thead>
         <b-tbody>
           <b-tr v-for="course in courses" :key="course.id">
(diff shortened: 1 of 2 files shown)


src/components/degree/student/UnassignedCourses.vue
```diff
@@ -109,7 +111,7 @@ export default {
       this.putFocusNextTick('name-input')
     },
     isEditing(course) {
-      return course.id === this.$_.get(this.courseForEdit, 'id')
+      return false && course.id === this.$_.get(this.courseForEdit, 'id')
```
lyttam, line 114: @johncrossman `isEditing` will always return false. Intentional?

johncrossman: @lyttam - I disabled that for the moment. :) Good catch but you can ignore.

PR #2940, opened 2021-04-21
BOAC-4042: Degree note table, model, and API

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4042

### DIFF
--- a/scripts/db/schema.sql
+++ b/scripts/db/schema.sql
@@ -442,6 +442,25 @@ ALTER TABLE ONLY degree_progress_courses
 
 --
 
+CREATE TABLE degree_progress_notes (
+  template_id integer NOT NULL,
+  body text NOT NULL,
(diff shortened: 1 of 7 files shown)


scripts/db/schema.sql
```diff
@@ -442,6 +442,25 @@ ALTER TABLE ONLY degree_progress_courses
 
 --
 
+CREATE TABLE degree_progress_notes (
+  template_id integer NOT NULL,
+  body text,
```
johncrossman, line 447: @lyttam - Should `body` be non-nullable? Empty string allowed, of course.

johncrossman:
@lyttam - Looks good. Posted one question.

scripts/db/schema.sql
lyttam: @johncrossman I think that makes sense. I will amend.

PR #2944, opened 2021-04-22
BOAC-4096, Degree Progress: better use of space in courses-table of column/category section

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4096


pfarestveit:
Thanks @johncrossman , we'll see what Teena thinks.

PR #2953, opened 2021-04-23
BOAC-4104, remove 'responsive' from b-table so assign-course dropdown can overflow

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4104

The Jira has more info.


pauline2k:
Tricky!

PR #2955, opened 2021-04-26
BOAC-3871, category-to-course relationship becomes many-to-many

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4053

From [designs](https://jira.ets.berkeley.edu/jira/browse/BOAC-3871): "There are times when an individual course may fulfill multiple requirements. For example, American History can also count towards a Humanities/Social Sciences requirement."

The order of the mappings will matter. The first mapping (see created_at date) is the primary and subsequent mappings are "clones" that are deleted (see trash icon) as opposed to "unassigned".


pauline2k:
@johncrossman - The design states that: "The number of units, GPA, and notes are carried over, however units and notes may be edited after the course has been cloned." I was interpreting that to mean that a copied course would be represented as an entirely new degree_progress_course row, with initial values copied over from the previous row. Would the many-to-many relationship allow different values for units and notes in course entries under different categories?

johncrossman:
@pauline2k - Good catch. Let me review. Hold off on merge.

johncrossman:
@pauline2k - Yes, thank you. This will not serve that reqt. I think we need to alter the primary-key definition on `degree_progress_courses` by adding standard serial `id` and then enforce uniqueness on (section_id, sid, term_id, category_id) tuple. Closing this PR.

PR #2964, opened 2021-04-27
BOAC-4058: Fixed (sub)category font sizes

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4058

johncrossman:
@cesarvh - In PR #2928 you add "scoped" to an existing 'degree-progress-category' style which is why we now see breakage. You also added a new 'degree-progress-category' style which this PR has moved to `boac-global`.  I don't think this will fix the problem but it's a bit hard to track the changes since we have two different style definitions with the same name, one scoped and one not.

johncrossman:
@cesarvh - If you remove `degree-progress-category` and `degree-progress-subcategory` from `PrintableDegreeTemplate.vue` then I'll merge this. Thanks!

PR #2973, opened 2021-04-29
BOAC-4053, /api/degree response packages assigned and unassigned courses

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4053

PR includes:
* 'Add Course' menu populated
* simplify the degree-edit-session store a bit
* removed dedicated `/unassigned` endpoint

### DIFF
--- a/src/components/degree/CoursesTable.vue
+++ b/src/components/degree/CoursesTable.vue
@@ -1,9 +1,9 @@
 <template>
   <div>
-    <div v-if="$_.isEmpty(courses)" class="no-data-text">
+    <div v-if="$_.isEmpty(items)" class="no-data-text">
       No courses
     </div>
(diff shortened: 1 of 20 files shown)


src/components/degree/CoursesTable.vue
```diff
@@ -1,9 +1,9 @@
 <template>
   <div>
-    <div v-if="$_.isEmpty(courses)" class="no-data-text">
+    <div v-if="$_.isEmpty(items)" class="no-data-text">
```
johncrossman, line 3: I renamed the 'courses' prop to 'items' because (1) these are course-requirement objects, not courses and (2) degree-edit-session now has a `courses` property which is a collection of assigned and unassigned courses, which will be useful in other components.

lyttam:
Seems sensible!

PR #2979, opened 2021-05-03
BOAC-4061, /course/copy with test coverage

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4061

Elaborate rules verified in test coverage.


pauline2k:
And that’s how you make copies.

PR #2981, opened 2021-05-03
BOAC-3981, 'Future' terms first in EGT cohort filter options

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3981


pauline2k:
If man is still alive…

PR #3000, opened 2021-05-06
BOAC-4148, custom UnitsInput component; API has rangeLower/upper instead of stringified

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4148
https://jira.ets.berkeley.edu/jira/browse/BOAC-4149

Consolidate validation and range handling.


johncrossman:
BOA hits PR 3000

image

lyttam:
Thanks for the fix.

PR #3005, opened 2021-05-07
BOAC-4164, speed up page load of /degrees and /degree/history with slimmer queries

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4164

Degree template/check objects are bloated by categories, courses, etc. The listing pages only need name, id, and not much else.


cesarvh:
Looks good :) Nice cleanup!

PR #3006, opened 2021-05-07
BOAC-4109, BOAC-4110: Adds cohort and group menus to batch degree check form

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4109 - display total students count
https://jira.ets.berkeley.edu/jira/browse/BOAC-4110

### DIFF
--- a/src/components/util/Autocomplete.vue
+++ b/src/components/util/Autocomplete.vue
@@ -25,7 +25,7 @@
         <b-button
           :id="`${id}-add-button`"
           class="btn btn-primary-color-override"
-          :disabled="addButtonLoading || addButtonDisabled(query, selectedSuggestion)"
+          :disabled="addButtonLoading || isLoading || addButtonDisabled(query, selectedSuggestion)"
           @click="onAddButton"
           @keyup.enter="onAddButton"
         >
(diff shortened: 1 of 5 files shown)


src/components/util/Autocomplete.vue
```diff
@@ -246,8 +248,11 @@ export default {
     onAddButton() {
       this.addButtonLoading = true
       this.addSelection(this.selectedSuggestion || this.query).then(() => {
-        this.addButtonLoading = false
         this.closeSuggestions()
+      }).catch(() => {
+        this.$nextTick(() => this.isOpen = false)
+      }).finally(() => {
+        this.addButtonLoading = false
```
johncrossman, line 255: @lyttam - Just curious, what type of error/exception are we catching here?

johncrossman:
Looks good to me @lyttam  I posted one question.

src/components/util/Autocomplete.vue
lyttam: @johncrossman If you enter some text and hit the Add button, but the text fails validation (`validateSids` returns false), then we want to close the suggestions without clearing the text from the input field. My thought was that clearing the text removes the context for the error message that displays below.

johncrossman: Thanks @lyttam

PR #3012, opened 2021-05-10
BOAC-4055, drag assigned course to 'Unassigned' to unassign using native Drag API

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4055

Because this page is complex, using native drag-and-drop API – instead of a third-party Vue.js lib – will keep our code cleaner and give us more control.

### DIFF
--- a/src/components/degree/CoursesTable.vue
+++ b/src/components/degree/CoursesTable.vue
@@ -25,7 +25,12 @@
             <b-tr
               :id="`course-${bundle.category.id}-table-row-${index}`"
               :key="`tr-${index}`"
-              class="font-size-16"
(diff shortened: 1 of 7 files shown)


src/components/degree/CoursesTable.vue
```diff
@@ -197,8 +202,7 @@ export default {
   },
   data: () => ({
     bundleForDelete: undefined,
-    bundleForEdit: undefined,
-    isAddingCourse: false
+    bundleForEdit: undefined
```
johncrossman, line 205: `isAddingCourse` is not used.

PR #3018, opened 2021-05-11
BOAC-4172, drag assigned course to a course requirement elsewhere in degree check

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4172

The styles-during-drag are a work in progress. Also, console debug logging in local-dev env only.

### DIFF
--- a/src/mixins/DegreeEditSession.vue
+++ b/src/mixins/DegreeEditSession.vue
@@ -30,6 +30,7 @@ export default {
       'degreeNote',
       'disableButtons',
       'includeNotesWhenPrint',
+      'isUserDragging',
       'sid',
       'templateId',
       'unitRequirements',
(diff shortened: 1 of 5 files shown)


src/mixins/DegreeEditSession.vue
```diff
@@ -71,14 +73,6 @@ export default {
...
         return category.courseRequirements
       }
-    },
-    canAssignCourseToCategory(category, course) {
-      if (!category || !course) {
-        return false
-      }
-      return (category.categoryType === 'Course Requirement' && !!category.courseIds.length)
-        || (category.categoryType === 'Category' && !!category.subcategories.length)
-        || category.courseIds.includes(course.id)
```
johncrossman, line 81: The above was a small goof in previous PR. The bool expression above is for when-to-disable-option, the opposite of "allow". This bit was reverted and drag/drop restrictions are encapsulated in the 'store'.

PR #3022, opened 2021-05-12
BOAC-4183, verify valid float before putting numeric-range in db

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4183


lyttam:
Thanks! Will merge when green.

PR #3023, opened 2021-05-12
BOAC-4111: Degree check batch creation

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4111

More work is needed to improve performance. I'm planning to switch the `clone` function to bulk SQL inserts, similar to how batch note creation works.

### DIFF
--- a/src/views/degree/BatchDegreeCheck.vue
+++ b/src/views/degree/BatchDegreeCheck.vue
@@ -41,7 +41,7 @@
         <span>Type a name, individual Student Identification (SID), or paste a list of SID numbers below. Example: 9999999990, 9999999991</span>
       </label>
     </div>
-    <div class="mb-2">
+    <div class="mb-3">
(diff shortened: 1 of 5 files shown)


src/views/degree/BatchDegreeCheck.vue
```diff
@@ -228,10 +248,12 @@ export default {
       this.templateId = templateId
     },
     removeCohort(cohort) {
+      console.log(`removing ${cohort.id}`)
       this.$_.remove(this.addedCohorts, c => c.id === cohort.id),
       this.recalculateStudentCount(this.addedSids, this.addedCohorts, this.addedCuratedGroups)
     },
     removeCuratedGroup(curatedGroup) {
+      console.log(`removing ${curatedGroup.id}`)
```
johncrossman, line 256: @lyttam - Replace console.logging above with screenreader alerts?

PR #3031, opened 2021-05-14
BOAC-4112: Alerts user to students already assigned degree check

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4112

### DIFF
--- a/boac/api/degree_progress_student_controller.py
+++ b/boac/api/degree_progress_student_controller.py
@@ -26,7 +26,7 @@ ENHANCEMENTS, OR MODIFICATIONS.
 from boac.api.degree_progress_api_utils import clone_degree_template, create_batch_degree_checks
 from boac.api.errors import BadRequestError, ResourceNotFoundError
 from boac.api.util import can_edit_degree_progress, can_read_degree_progress
-from boac.externals.data_loch import get_sid_by_uid
+from boac.externals.data_loch import get_basic_student_data, get_sid_by_uid
 from boac.lib.http import tolerant_jsonify
 from boac.lib.util import get as get_param, is_int, to_int_or_none
 from boac.models.degree_progress_category import DegreeProgressCategory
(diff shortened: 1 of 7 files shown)


boac/api/degree_progress_student_controller.py
```diff
@@ -166,6 +166,30 @@ def get_degree_checks(uid):
...
+@can_edit_degree_progress
+def get_students(template_id):
+    params = request.get_json()
+    sids = get_param(params, 'sids')
+    if not sids:
+        raise BadRequestError('sids is required.')
+    template = DegreeProgressTemplate.find_by_id(template_id)
+    if not template:
+        raise ResourceNotFoundError('Degree template not found')
+    degree_checks = DegreeProgressTemplate.find_by_name(template.degree_name, student_sids=sids)
```
johncrossman, line 179: @lyttam - This won't work if the template is renamed, right? Will we change the logic when student degree-checks have foreign-key ref back to template?  If yes, maybe a `TODO:` line above would remind us later.

lyttam: You're right @johncrossman - but maybe that's ok? What if the template gets changed (including the name) and the advisor wants all their students to get the new version?

lyttam: Either way, matching by the degree name doesn't seem ideal. I will add a TODO regardless.

johncrossman: @lyttam - I think "advisor wants all their students to get the new version" was ruled out as a requirement. We can circle back to this TODO soon.

lyttam: @johncrossman Thanks, I amended with a TODO comment.

johncrossman:
Thanks @lyttam  Waiting on Travis.

johncrossman:
`lint-py` complains.

lyttam:
Fixed the lint-py issue, @johncrossman

PR #3032, opened 2021-05-17
BOAC-4132, degree progress: 'student' prop not necessary cuz 'sid' is in vuex session

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4132


johncrossman:
Travis did not waking up, even after PR amendment. I'll investigate.

PR #3034, opened 2021-05-18
BOAC-4139: Adds revision indicator to student degree check

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4139


johncrossman:
Looks good. Thanks @lyttam

PR #3045, opened 2021-05-26
BOAC-4162: Skips error page when error should be displayed inline

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4162

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -63,21 +63,28 @@ router.afterEach(writeHistory)
 
 const apiBaseUrl = process.env.VUE_APP_API_BASE_URL
 
+const isHandledInComponent = error => {
+  const errorUrl = _.get(error, 'response.config.url')
+  return errorUrl && errorUrl.includes('/api/user/create_or_update')
+}
+
 // Axios
 const axiosErrorHandler = error => {
   const errorStatus = _.get(error, 'response.status')
   if (_.get(Vue.prototype.$currentUser, 'isAuthenticated')) {
     if (errorStatus === 404) {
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -89,6 +96,8 @@ const axiosErrorHandler = error => {
   }
 }
 
+Vue.prototype.$errorHandler = axiosErrorHandler
```
johncrossman, line 99: @lyttam - Just curious... What is purpose of having `$errorHandler` globally available?

pauline2k: @johncrossman It might be helpful to look at https://github.com/ets-berkeley-edu/junction/pull/7914 and https://github.com/ets-berkeley-edu/junction/pull/7944 for equivalent work we had to do in the Junction Vue rewrite.

lyttam: @johncrossman Oops, it doesn't need to be global. I will amend.

pauline2k: The `$errorHandler` would need to be globally available only if and only if there were API calls to `/api/user/create_or_update` that we wanted individual components _not_ to handle themselves, but instead pass back to default handling. That's the purpose of the pattern in ets-berkeley-edu/junction#7944. If `/api/user/create_or_update` calls will always be handled in the component, no need to make the handler global.

lyttam: Thank you @pauline2k, that's good to know. So far I don't think we've run into that situation in BOA.

johncrossman:
We'll bring that line back when necessary.

PR #3051, opened 2021-05-27
BOAC-4204: Expands EPN/CPN cohort filter to all advisors and range of terms

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4204


pauline2k:
@lyttam In the past we've had to do migration scripts on release to adapt to changed cohort filter keys. See for example the files in https://github.com/ets-berkeley-edu/boac/tree/master/scripts/db/migrate/2019/20190716-BOAC-2256.

PR #3053, opened 2021-05-27
BOAC-4126: Added support for blur mode on print view

cesarvh:
It wasn't until I was about to submit this that I realized it was my external monitor's settings that made the text basically invisible 🤦‍♂️


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4126

PR #3084, opened 2021-06-09
BOAC-4198, copied-course can be moved and, in the process, its placeholder category is deleted

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4198

The db.migrate script adds a new category_type to identify “copied” courses. Read the Jira for more detail.


scripts/db/migrate/2021/20210608-BOAC-4198/pre_deploy_degree_progress_category_type_add_value.sql
```diff
@@ -0,0 +1,2 @@
+
+ALTER TYPE degree_progress_category_types ADD VALUE 'Placeholder: Course Copy';
```
johncrossman, line 2: Release instructions updated.  (Note: script has no begin/end because "ALTER TYPE ... ADD" cannot run inside a transaction block.)

PR #3086, opened 2021-06-09
BOAC-4196: Allows cats, subcats to be linked to unit reqts; BOAC-4250

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4196

Also fixes a bug with creating/updating unit requirements in the template builder:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4250

### DIFF
--- a/src/components/degree/EditCategory.vue
+++ b/src/components/degree/EditCategory.vue
@@ -56,12 +56,13 @@
           :units-upper="unitsUpper"
         />
       </div>
-      <div v-if="selectedCategoryType === 'Course Requirement' && unitRequirements.length">
+      <div v-if="unitRequirements.length">
         <div class="font-weight-500 pb-1">
           Requirement Fulfillment
         </div>
         <div class="mb-3">
           <SelectUnitFulfillment
+            :ref="`column-${position}-unit-requirement-select`"
             :disable="isSaving"
             :initial-unit-requirements="selectedUnitRequirements"
             :on-unit-requirements-change="onUnitRequirementsChange"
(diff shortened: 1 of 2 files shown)


src/components/degree/EditCategory.vue
```diff
@@ -249,9 +250,18 @@ export default {
...
-      this.$announcer.polite(option ? `${this.selectedParentCategory} selected` : 'Unselected')
+      this.$announcer.polite(option ? `${this.selectedParentCategory.name} selected` : 'Unselected')
+      const existingUnitRequirements = this.selectedUnitRequirements
+      const parentUnitRequirements = this.$_.get(this.selectedParentCategory, 'unitRequirements')
+
       if (option) {
+        const inheritedUnitRequirements = this.$_.differenceBy(parentUnitRequirements, existingUnitRequirements, 'id')
+        this.$_.each(inheritedUnitRequirements, unitRequirement => {
+          this.$refs[`column-${option.position}-unit-requirement-select`].onChangeUnitRequirement(unitRequirement)
+        })
```
johncrossman, line 261: @lyttam - Is this the fix to BOAC-4250?  I.e., refresh what's in view when changes are made to unit-requirements? If so, I'd recommend using `$_refresh()` in `degree-edit-session.ts` instead.  We could add that refresh call in the `updateUnitRequirement` (vuex action).

lyttam: That's right @johncrossman . Good tip; I will amend.

lyttam: Correction: the change to UnitRequirements.vue line 245 is the fix for BOAC-4250. But I think your suggestion still makes sense.

johncrossman: @lyttam - It's unclear what line 260 is doing. Can you give short explanation?

lyttam: @johncrossman When the requirement location (parent category) is selected, that line triggers the change event on the SelectUnitFulfillment component for each unit reqt belonging to the parent. The effect is the inherited unit reqt(s) appear as though the user had selected them.

johncrossman:
Thanks @lyttam

PR #3100, opened 2021-06-15
BOAC-4235, tamer scroll when dragging degree-check courses

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4235

Event listeners track drag and scroll. Auto-scroll only if the drag is in lower region (< 100) of viewport.


pauline2k:
Next: websockets!

PR #3110, opened 2021-06-17
BOAC-2426: Requirements are no longer clickable in the Academic timeline

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-2426

### DIFF
--- a/src/components/student/profile/AcademicTimelineTable.vue
+++ b/src/components/student/profile/AcademicTimelineTable.vue
@@ -140,11 +140,11 @@
               :id="`timeline-tab-${activeTab}-message-${index}`"
               :aria-pressed="$_.includes(openMessages, message.transientId)"
               :class="{
-                'align-top message-open': $_.includes(openMessages, message.transientId),
+                'align-top message-open': $_.includes(openMessages, message.transientId) && message.type !== 'requirement' ,
                 'truncate': !$_.includes(openMessages, message.transientId),
                 'img-blur': $currentUser.inDemoMode && ['appointment', 'note'].includes(message.type)
               }"
-              role="button"
+              :role="message.type === 'requirement' ? '' : 'button'"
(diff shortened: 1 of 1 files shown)


src/components/student/profile/AcademicTimelineTable.vue
```diff
@@ -140,11 +140,11 @@
               :id="`timeline-tab-${activeTab}-message-${index}`"
               :aria-pressed="$_.includes(openMessages, message.transientId)"
               :class="{
-                'align-top message-open': $_.includes(openMessages, message.transientId),
+                'align-top message-open': $_.includes(openMessages, message.transientId) && message.type !== 'requirement' ,
                 'truncate': !$_.includes(openMessages, message.transientId),
                 'img-blur': $currentUser.inDemoMode && ['appointment', 'note'].includes(message.type)
               }"
-              role="button"
+              :role="message.type === 'requirement' ? 'textbox' : 'button'"
```
johncrossman, line 147: @cesarvh - What does this mean for accessibility? Would screen-reader user perceive this as text input?  And I'm curious how `textbox` makes it clickable. :)

pauline2k: @johncrossman "Requirements are no longer clickable" isn't a description of a bug, it's actually the desired behavior, so it's true that we don't want a `button` role. But `textbox` implies free-form text input, which doesn't seem quite right in this case. If this row isn't supposed to be expandable, does it need an aria role at all?

@cesarvh - is it enough to change the aria role and CSS on this without altering the event handling? What happens now when the `open` function is called on a click or keypress for a requirement? It seems like we would still get a screenreader alert, even if the "open" status is now visually indistinguishable from being closed.

cesarvh: @pauline2k Thanks for the clarification :)

pauline2k:
Thanks @cesarvh!

PR #3112, opened 2021-06-18
BOAC-4278, PrintableDegreeTemplate tweaks: CSS, concise course name

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4278

### DIFF
--- a/src/components/degree/CoursesTable.vue
+++ b/src/components/degree/CoursesTable.vue
@@ -12,11 +12,11 @@
             <b-th v-if="assignedCourseCount && canEdit" class="th-course-assignment-menu">
               <span class="sr-only">Options to re-assign course</span>
             </b-th>
-            <b-th class="pl-0" :class="{'font-size-10': printable}">Course</b-th>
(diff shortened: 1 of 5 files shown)


src/components/degree/CoursesTable.vue
```diff
@@ -341,6 +346,19 @@ export default {
...
     },
+    getBundleName(course, category) {
+      let name = (course || category).name
+      if (course && this.printable) {
+        this.$_.each(['COL', 'DIS', 'FLD', 'GRP', 'IND', 'LAB', 'LEC', 'SEM'], format => {
+          const trimmed = name.replace(new RegExp(` ${format} [0-9]+$`), '')
+          if (trimmed !== name) {
+            name = trimmed
+            return false
+          }
```
johncrossman, line 357: Because `course.name` is coming from the flattened data in `degree_progress_courses` table, this runtime parsing is the best we can do. It won't catch all formats but that's okay.

PR #3117, opened 2021-06-22
BOAC-4290, /print degree page must tolerate null student

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4290


pauline2k:
Null student welcome here.

PR #3123, opened 2021-06-23
BOAC-4252, course cannot share category w/ copy but it can move within its own category

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4252

I renamed `getCourses()` to describe its actual purpose.

### DIFF
--- a/src/components/degree/CoursesTable.vue
+++ b/src/components/degree/CoursesTable.vue
@@ -27,7 +27,7 @@
               :key="`tr-${index}`"
               :class="{
                 'cursor-grab': isDraggable(bundle),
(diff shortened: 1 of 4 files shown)


src/components/degree/CoursesTable.vue
```diff
@@ -27,7 +27,7 @@
               :key="`tr-${index}`"
               :class="{
                 'cursor-grab': isDraggable(bundle),
-                'drop-zone-on': isDroppable(bundle.category) && draggingContext.target === bundle.category.id,
+                'drop-zone-on': isDroppable(bundle.category),
```
johncrossman, line 30: `draggingContext.target === bundle.category.id` check is redundant. It's also in the `isDroppable(...)` function.

PR #3130, opened 2021-06-24
BOAC-4307, do not offer save/rename if name.trim() is empty

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4307


pauline2k:
One more bomb defused.

PR #3131, opened 2021-06-25
BOAC-4243: Tracks batch create progress via caching

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4243


pauline2k:
Looks good to me, @lyttam!

PR #3132, opened 2021-06-25
BOAC-4308, if degree entry of type 'Category' has course-requirements then it cannot have a Subcategory

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4308

### DIFF
--- a/src/components/degree/EditCategory.vue
+++ b/src/components/degree/EditCategory.vue
@@ -106,7 +106,7 @@
               :id="`column-${position}-parent-select-option-${category.name}`"
               :key="category.id"
               :aria-label="`${category.categoryType} ${category.name}`"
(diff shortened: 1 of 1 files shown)


src/components/degree/EditCategory.vue
```diff
@@ -225,6 +225,13 @@ export default {
       this.selectedCategoryType = null
       this.afterCancel()
     },
+    disableLocationOption(option) {
+      const optionType = option.categoryType
+      const selectedType = this.selectedCategoryType
+      return (selectedType === 'Subcategory' && optionType === 'Subcategory')
+        || (selectedType === 'Subcategory' && optionType === 'Category' && this.getItemsForCoursesTable(option).length > 0)
```
johncrossman, line 232: The one line above is the only new logic in this PR.

PR #3168, opened 2021-07-06
BOAC-4292, CSS tweaks on /degree/batch

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4292

### DIFF
--- a/src/views/degree/BatchDegreeCheck.vue
+++ b/src/views/degree/BatchDegreeCheck.vue
@@ -40,12 +40,12 @@
       <div class="w-75">
         <label
           for="degree-check-add-student-input"
-          class="font-size-14 input-label text mt-2"
+          class="input-label text mt-1"
         >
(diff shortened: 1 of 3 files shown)


src/views/degree/BatchDegreeCheck.vue
```diff
@@ -89,26 +90,24 @@
...
-      <div class="mb-3">
+      <div class="mb-2">
         <BatchAddStudentSet
           v-if="myCuratedGroups && myCuratedGroups.length"
           class="w-75"
           :add-object="addCuratedGroup"
           :disabled="isSaving"
           :is-curated-groups-mode="true"
           :remove-object="removeCuratedGroup"
-          :target="'degree check'"
```
johncrossman, line 111: `target` prop was removed because it served no purpose. Also, its value of 'degree check' was putting spaces in element IDs. I will need to update Teena code, too.  cc: @pfarestveit

PR #3177, opened 2021-07-07
BOAC-4336, if zero templates then no link to 'Batch degree checks'

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4336


johncrossman:
It seems that stricter flake8 rules call `lint-py` failure. This PR did not touch py files but I'll fix anyway.

PR #3181, opened 2021-07-08
BOAC-4337, screen readers skip input placeholder; add the same text to sr-only label

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4337

According to https://www.w3.org/WAI/tutorials/forms/instructions/#placeholder-text, "placeholder text is not broadly supported across assistive technologies."

### DIFF
--- a/src/components/sidebar/SearchForm.vue
+++ b/src/components/sidebar/SearchForm.vue
@@ -269,6 +269,7 @@
         Search for students, courses, or notes.
         {{ searchInputRequired ? 'Input is required.' : '' }}
         {{ searchHistory.length ? 'Expect auto-suggest of previous searches.' : '' }}
+        (Type / to put focus in the search input field.)
       </span>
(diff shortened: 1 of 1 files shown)


src/components/sidebar/SearchForm.vue
```diff
@@ -269,6 +269,7 @@
         Search for students, courses, or notes.
         {{ searchInputRequired ? 'Input is required.' : '' }}
         {{ searchHistory.length ? 'Expect auto-suggest of previous searches.' : '' }}
+        (Type / to put focus in the search input field.)
```
johncrossman, line 272: Interesting info on screen readers w.r.t punctuation and typographic symbols:  https://www.deque.com/blog/dont-screen-readers-read-whats-screen-part-1-punctuation-typographic-symbols/

pauline2k: I was hoping the screen reader would say "virgule."

PR #3186, opened 2021-07-16
BOAC-3897, AcademicTimeline search: client-side index for filter-as-you-type

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3897

Much faster, too.


pauline2k:
Zippy!

PR #3191, opened 2021-07-21
BOAC-4352, vue and vue-router minor version bump

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4352

I can only reproduce the bug on Chrome, and I cannot reproduce locally. These minor upgrades are unlikely to fix but they can't hurt – and it gives us fresh package-lock deps.


johncrossman:
Bizarre failure. Closing non-essential PR.

PR #3192, opened 2021-07-21
BOAC-4352, minor version bumps for vue, vue-router and vue-template-compiler

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4352

The `vue-template-compiler` upgrade gets us past the error seen in PR #3191.


pauline2k:
Hard to object!

PR #3202, opened 2021-07-28
BOAC-3937, color-coded select; b-tr text respects course.accentColor

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-3937

### DIFF
--- a/src/components/degree/student/AccentColorSelect.vue
+++ b/src/components/degree/student/AccentColorSelect.vue
@@ -1,0 +1,114 @@
+<template>
+  <div>
+    <div>
+      <label
+        for="color-code-select"
+        class="font-weight-bolder"
+      >
+        Color Code
+      </label>
+    </div>
+    <div>
+      <b-dropdown
+        id="color-code-select"
+        v-model="selected"
+        block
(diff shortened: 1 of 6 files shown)


src/components/degree/student/AccentColorSelect.vue
```diff
@@ -0,0 +1,114 @@
...
+    setSelected(value) {
+      this.selected = value
+      this.onChange(this.selected)
+      this.$announcer.polite(`${this.selected} selected`)
+    }
+  }
+}
+</script>
+
+<style>
```
lyttam, line 92: @johncrossman Should these styles be `scoped`?

johncrossman: Yes indeed. Amending.

johncrossman: @lyttam - Actually, it cannot be `scoped` because the style is applied to a bootstrap component. No amendment needed.

lyttam: Ah, makes sense.

PR #3205, opened 2021-07-28
BOAC-4371, front-end adjusts to latest course uniqueness constraint

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4371
https://jira.ets.berkeley.edu/jira/browse/BOAC-4370


src/components/degree/student/UnassignedCourses.vue
```diff
@@ -249,6 +251,9 @@ table {
   border-collapse: separate;
   border-spacing: 0 0.05em;
 }
+.btn-container {
+  min-width: 20px;
+}
```
johncrossman, line 256: Unassigned list may have two buttons (edt, delete) in far right column. When user begins drag we don't want to disappear that column and cause UI jerkiness.

PR #3241, opened 2021-08-16
BOAC-4363: YCBM data now correctly displayed in UI (advisor, cancellation)

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4363


pauline2k:
Thanks, @cesarvh, let's get it on dev and see how it looks.

PR #3243, opened 2021-08-17
BOAC-4418: YCBM appointment title now appearing correctly

cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4418


lyttam:
Looks good!

PR #3246, opened 2021-08-18
BOAC-4421: Removed Arrived @ for YCBM appts

cesarvh:


cesarvh:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4421

PR #3259, opened 2021-08-24
BOAC-4409, after assign junk/unassigned course, put focus on component header

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4409

Give and take w.r.t accessibility. After course assignment, the 'Unassigned' header is the only feasible focus target. Thus, the h3 gets tabindex="-1".  Same for 'Free Electives'.


pauline2k:
Oh, tabindex.

PR #3269, opened 2021-09-02
BOAC-4429, you say canceled, I say cancelled, let's call the whole thing off

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4429

Only changed what the customer sees.


lyttam:
Cancel culture strikes again.

PR #3274, opened 2021-09-15
BOAC-4436 Add class number to course heading display

pauline2k:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4436


johncrossman:
LGTM. Do we need a git amend to wake up Travis?  cc: @pauline2k

PR #3278, opened 2021-09-15
BOAC-4472: Hides grading basis details on non grading basis change eForms

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4472

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -665,8 +665,9 @@ def get_sis_advising_appointments(sid):
 def get_sis_late_drop_eforms(sid):
     sql = f"""
         SELECT
-            id, course_display_name, course_title, created_at, eform_id, eform_status, eform_type, grading_basis_code,
-            grading_basis_description, requested_action, requested_grading_basis_code, requested_grading_basis_description,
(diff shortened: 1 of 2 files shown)


boac/externals/data_loch.py
```diff
@@ -665,8 +665,9 @@ def get_sis_advising_appointments(sid):
 def get_sis_late_drop_eforms(sid):
     sql = f"""
         SELECT
-            id, course_display_name, course_title, created_at, eform_id, eform_status, eform_type, grading_basis_code,
-            grading_basis_description, requested_action, requested_grading_basis_code, requested_grading_basis_description,
+            id, course_display_name, course_title, created_at, eform_id, eform_status, eform_type,
+            NULLIF(grading_basis_description, ' ') AS grading_basis_description, requested_action,
+            NULLIF(requested_grading_basis_description, ' ') AS requested_grading_basis_description,
```
johncrossman, line 670: @lyttam - are these values always a single space char?  As opposed to empty string.

lyttam: @johncrossman They are always a single space it seems - there are no empty strings in the table.

PR #3280, opened 2021-09-16
BOAC-4423, curated bulk SIDs: if invalid.length <= magic then list 'em out

johncrossman:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4423

Auto-fix is only available if SIDs.length > magic.


pauline2k:
00100450279_R1GMQ4i

PR #3293, opened 2021-10-14
BOAC-4450: API for Campus Requirements

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4450


pauline2k:
Thanks @lyttam, looks good to me.

PR #3313, opened 2021-10-22
BOAC-4493: Switches over to new terms RDS schema

lyttam:
https://jira.ets.berkeley.edu/jira/browse/BOAC-4493

QA PR #3312


pauline2k:
Oh, very sneaky. Thanks for tracking this one down, @lyttam.

PR #3333, opened 2021-11-02
BOAC-4487 For want of a pipe character the deploy was lost

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4487


johncrossman:
image

PR #3366, opened 2021-11-10
BOAC-4524, old-school YAML for cloudwatch config in 00_ami.config

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4524

Our use of JSON-based config might have screwy syntax?!


pauline2k:
Maybe it does!

PR #3371, opened 2021-11-15
BOAC-4537, private notes: apply a template and privacy selector remains

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4537

Also, removed useless quick links (eg, Schedulers) from Passenger Manifest.


johncrossman:
cc: @pauline2k @lyttam

PR #3394, opened 2021-12-02
BOAC-4532, unit fulfillment(s) can be added to a course iff it is assigned

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4532

### DIFF
--- a/src/components/degree/student/EditCourse.vue
+++ b/src/components/degree/student/EditCourse.vue
@@ -63,6 +63,19 @@
         :on-change="value => accentColor = value"
       />
     </div>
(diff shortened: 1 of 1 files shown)


src/components/degree/student/EditCourse.vue
```diff
@@ -63,6 +63,19 @@
         :on-change="value => accentColor = value"
       />
     </div>
+    <div v-if="course.categoryId || $_.size(selectedUnitRequirements)">
```
johncrossman, line 66: I added `... || $_.size(selectedUnitRequirements)` in case we have a bug (unexpected unit-requirement values).  In case of such a bug, we want those values to show in UI and not live silently, under the hood.

PR #3396, opened 2021-12-03
BOAC-4556: caches cohort filter options nightly

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4556


boac/merged/cohort_filter_options.py
```diff
@@ -81,24 +81,24 @@ def get_filter_option_groups(self):
                     'Academic Standing',
                     options=academic_standing_options(min_term_id=sis_term_id_for_name(f'Fall {current_year - 5}')),
                 ),
-                _filter('colleges', 'College', options=colleges),
-                _filter('enteringTerms', 'Entering Term', options=entering_terms),
-                _filter('epnCpnGradingTerms', 'EPN/CPN Grading Option', options=grading_terms),
-                _filter('expectedGradTerms', 'Expected Graduation Term', options=grad_terms),
+                _filter('colleges', 'College', options=colleges()),
```
lyttam, line 84: I had to change these function references to function calls in order for the caching to work.

tests/test_api/test_auth_controller.py
```diff
@@ -115,6 +115,8 @@ def test_known_user_with_correct_password_logs_in(self, app, client):
     def test_user_expired_according_to_calnet(self, app, client):
         """Fails if user has no record in LDAP."""
         with override_config(app, 'DEVELOPER_AUTH_ENABLED', True):
+            from boac.models import json_cache
+            json_cache.clear('%')
```
lyttam, line 119: Mysteriously, this test started failing once I made the changes to test_cohort_utils.py, below. Adding these two lines fixed it.

lyttam:
I thought the simplest caching option would be to use the existing `JsonCache`, but the weird behavior in the unit tests makes me wonder if I should have used Flask-Caching instead.

tests/test_api/test_auth_controller.py
pauline2k: db session persistence in pytest remains sadly mysterious to me too.

pauline2k:
Caching is always going to introduce some weirdness, and the nice thing about JsonCache is that entries are so easily inspectable in the db. Let's give this a try as is!

PR #3405, opened 2021-12-07
BOAC-4377, padding for 'satisfied note'

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4377


pauline2k:
Padded with peanuts, this note satisfies.

PR #3407, opened 2021-12-07
BOAC-4377, printable degree check: no faint text, landscape, improved sizing

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4377

### DIFF
--- a/src/views/degree/PrintableDegreeTemplate.vue
+++ b/src/views/degree/PrintableDegreeTemplate.vue
@@ -2,16 +2,16 @@
   <div class="m-4">
     <Spinner />
     <b-container v-if="!loading" fluid>
-      <b-row>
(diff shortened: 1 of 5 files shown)


src/views/degree/PrintableDegreeTemplate.vue
```diff
@@ -162,6 +158,11 @@ export default {
 </script>
 
 <style scoped>
+@media print{
+  @page {
+    size: landscape;
+  }
+}
```
johncrossman, line 165: According to https://developer.mozilla.org/en-US/docs/Web/CSS/@page/size, CSS descriptor `@page.size.landscape` is supported by Chrome, Edge and Opera. Not yet supported on Firefox and Safari.

PR #3418, opened 2021-12-09
NS-1432 NS-1433 Remove a whole mess of special handling for historical tables

pauline2k:
https://jira-secure.berkeley.edu/browse/NS-1432
https://jira-secure.berkeley.edu/browse/NS-1433


johncrossman:
20K+ lines of code removed. Bravo!

pauline2k:
Most of those lines were in package-lock.json apparently but I'll take the praise 😸

PR #3419, opened 2021-12-09
BOAC-4568, toss out current-user-extras mixin; admit-student gets group selector

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4568


pauline2k:
It's extra trashed!

PR #3427, opened 2021-12-10
BOAC-4565, UI cleanup on /cohort and /curated; AdmitStudentRow gets 'curate' controls

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4565


pauline2k:
If we're putting `curate` controls in place, I expect to see `vicar` and `sexton` as well.

PR #3435, opened 2021-12-13
BOAC-4581, use proper label per curated_group.domain

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4581

### DIFF
--- a/src/components/sidebar/CuratedGroups.vue
+++ b/src/components/sidebar/CuratedGroups.vue
@@ -2,12 +2,12 @@
   <div>
     <div class="d-flex justify-content-between mb-1 sidebar-row-link">
       <div class="ml-1" :class="headerClass">
-        {{ headerText }}
+        {{ domain === 'admitted_students' ? 'CE3 Groups' : 'Curated Groups' }}
       </div>
       <div class="ml-1 mr-2">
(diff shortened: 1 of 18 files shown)


src/components/sidebar/CuratedGroups.vue
```diff
@@ -24,7 +24,7 @@
       <div class="ml-1 truncate-with-ellipsis">
         <NavLink
           :id="`sidebar-curated-group-${index}`"
-          :aria-label="'Curated group ' + group.name + ' has ' + group.totalStudentCount + ' students'"
+          :aria-label="`${$_.capitalize(domainLabel(false))} ${group.name} has ${group.totalStudentCount} students.`"
```
lyttam, line 27: @johncrossman  Why `_.capitalize()` here when you could pass `capitalize=true` to `domainLabel()`?

johncrossman: @lyttam - Good eye. `_.capitalize` only modifies first character of the string. The util param will cause all words to capitalize. Subtle.

lyttam: Ah, thanks!

PR #3453, opened 2021-12-16
BOAC-4593, wrong offset value when get_admitted_students_by_sids()

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4593


boac/api/curated_group_controller.py
```diff
@@ -222,7 +222,7 @@ def _curated_group_with_complete_student_profiles(
     if curated_group.domain == 'admitted_students':
         api_json['students'] = get_admitted_students_by_sids(
             limit=limit,
-            offset=offset,
+            offset=0,
```
johncrossman, line 225: The list of sids was queried using the proper offset. This is the follow up query which gets the chunked set of sids.

PR #3458, opened 2021-12-21
BOAC-4558 Drop SQLAlchemy abstraction layer from loch queries

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4558

We can't update to psycopg3 yet, it seems, because SQLAlchemy doesn't yet know how to adapt certain object types.

### DIFF
--- a/boac/externals/data_loch.py
+++ b/boac/externals/data_loch.py
@@ -31,35 +31,27 @@ from boac.lib.berkeley import previous_term_id, sis_term_id_for_name
 from boac.lib.mockingdata import fixture
 from boac.lib.util import join_if_present, tolerant_remove
 from flask import current_app as app
-import sqlalchemy
-from sqlalchemy import create_engine
-from sqlalchemy.sql import text
-
-
-# Lazy init to support testing.
-data_loch_db_rds = None
+import psycopg2
+import psycopg2.extras
 
 
 def safe_execute_rds(string, **kwargs):
-    global data_loch_db_rds
-    if data_loch_db_rds is None:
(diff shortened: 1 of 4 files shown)


boac/externals/data_loch.py
```diff
@@ -169,14 +161,14 @@ def get_enrolled_primary_sections_for_parsed_code(term_id, subject_area, catalog
         'catalog_id': f'{catalog_id}%',
     }
     if subject_area:
-        subject_area_clause = 'AND sis_subject_area_compressed LIKE :subject_area'
+        subject_area_clause = 'AND sis_subject_area_compressed LIKE %(subject_area)s'
```
pauline2k, line 164: @johncrossman @lyttam Note the bulk of this diff, which involves changing the syntax for passing in parameters, but only when using raw psycopg to talk to Nessie RDS. When talking to BOA RDS we still use SQLAlchemy and the colon syntax stands.

boac/externals/data_loch.py
```diff
@@ -843,7 +835,7 @@ def get_colleges():
     sql = f"""SELECT DISTINCT maj.college
         FROM {student_schema()}.student_profile_index spi
         JOIN {student_schema()}.student_majors maj ON maj.sid = spi.sid AND spi.academic_career_status = 'active'
-        WHERE maj.college NOT LIKE 'Graduate%'
+        WHERE maj.college NOT LIKE 'Graduate%%'
```
pauline2k, line 838: Note also, in the rare case where 1) you're using raw psycopg and 2) a percent pattern is part of your hardcoded SQL string (not passed in as a parameter), you have to double up the percent.

PR #3459, opened 2021-12-21
BOAC-2216: Removes unused student profile elements currentTerm, termGpa, academicStanding

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-2216

`termGpa` and `academicStanding` are not used on the course page, and `currentTerm` isn't used on any page.


pauline2k:
Thanks for lightening the load!

PR #3488, opened 2022-01-18
BOAC-4618, Degree Progress permissions: automate=TRUE is default for CoE

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4618


scripts/db/migrate/2022/20220106-BOAC-4314/add_column_automate_degree_progress_permission.sql
```diff
@@ -1,12 +1,15 @@
 BEGIN;
 
-ALTER TABLE authorized_users ADD COLUMN automate_degree_progress_permission BOOLEAN DEFAULT FALSE;
+ALTER TABLE authorized_users
+ADD COLUMN IF NOT EXISTS automate_degree_progress_permission BOOLEAN DEFAULT FALSE;
```
johncrossman, line 4: "IF Not Exists" so script can be run again in boa-dev.

PR #3504, opened 2022-02-10
BOAC-4636 Open degree checks with link, not button

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4634


pauline2k:
Never mind what Slack says, Travis is green at the source.

PR #3518, opened 2022-03-10
BOAC-4638: Added units in progress to CSV export

cesarvh:
https://jira-secure.berkeley.edu/browse/BOAC-4638


johncrossman:
Good news! We have a legit py-linter error.  Amend the PR to fix and I bet we'll get a green Travis.

pauline2k:
@cesarvh Looks like lint-py is complaining about expecting blank lines in a couple places.

PR #3521, opened 2022-03-14
BOAC-4650: Student page - suppress non-SIS, Canvas-only enrollments

cesarvh:
https://jira-secure.berkeley.edu/browse/BOAC-4650

### DIFF
--- a/src/views/Student.vue
+++ b/src/views/Student.vue
@@ -113,22 +113,7 @@ export default {
         next()
       }
     },
-    decorateCourse(course) {
-      // course_code is often valuable (eg, 'ECON 1 - LEC 001'), occasionally not (eg, CCN). Use it per strict criteria:
-      const useCourseCode = /^[A-Z].*[A-Za-z]{3} \d/.test(course.courseCode)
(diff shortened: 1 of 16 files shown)


src/views/Student.vue
```diff
@@ -113,22 +113,7 @@ export default {
...
-        canvasSites: [course]
-      })
-    },
     parseEnrollmentTerm(term) {
-      // Merge in unmatched canvas sites
-      const unmatched = this.$_.map(
-        term.unmatchedCanvasSites,
-        this.decorateCourse
-      )
-      term.enrollments = this.$_.concat(term.enrollments, unmatched)
```
johncrossman, line 131: @cesarvh - Should we have `term.enrollments = []` in place of the (removed) line above? The line just below will iterate on the value.

johncrossman: Ignore that previous comment.  I see that the line above is extending the existing `term.enrollments`

pauline2k:
Looks good to me! Thanks @cesarvh

PR #3523, opened 2022-03-28
BOAC-3952: Added academic division cohort filter

cesarvh:
Can't currently test this on my local machine since Nessie hasn't updated. I'll leave the PR here so I don't forget, and make sure it works once Nessie updates the tables.

### DIFF
--- a/tests/test_api/test_cohort_controller.py
+++ b/tests/test_api/test_cohort_controller.py
@@ -1126,6 +1126,21 @@ class TestCohortPerFilters:
         assert students[0]['sid'] == '3141592653'
         assert students[1]['sid'] == '2718281828'
 
+    def test_filter_division(self, client, coe_advisor_login):
+        api_json = self._api_get_students_per_filters(
+            client,
+            {
+                'filters': [
+                    {'key': 'academicDivisions', 'value': 'L&S Arts & Humanities Division'},
+                ],
+            },
+        )
+
+        students = api_json['students']
+        assert len(students) == 2
+        assert students[0]['sid'] == '11667051'
+        assert students[1]['sid'] == '3456789012'
+
     def test_filter_entering_term(self, client, coe_advisor_login):
         api_json = self._api_get_students_per_filters(
             client,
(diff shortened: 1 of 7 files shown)


cesarvh:
https://jira-secure.berkeley.edu/browse/BOAC-3952

pauline2k:
@cesarvh - Application code looks good. To get test fixtures to match what we now expect from nessie, you'll need to update the student_majors table in fixtures/loch.sql (and it would also be good to add a test in test_api/test_cohort_controller to verify you can get records for the new filter type).

#2359, which added the colleges filter, has patterns for both of those.

tests/test_api/test_cohort_controller.py
```diff
@@ -1126,6 +1126,23 @@ def test_filter_degree_term(self, client, coe_advisor_login):
...
+            {
+                'filters': [
+                    {'key': 'academicDivisions', 'value': 'L&S Arts & Humanities Division'},
+                ],
+            },
+        )
+
+        students = api_json['students']
+        print (students)
+        assert len(students) == 3
```
pauline2k: @cesarvh Three students in the results is a totally reasonable expectation given the test data you just put in, but in fact BOA is following correct (tangled) logic in only returning two. If you look up at the student.student_profile_index test data in loch.sql, you'll see that UID 2718281828 has a 'completed' academic status. By default, BOA cohorts only return students with 'active' status unless there's a filter specifically requesting otherwise.

cesarvh: Ah. I see. Thanks!

tests/test_api/test_user_controller.py
```diff
@@ -716,6 +716,7 @@ def test_user_search_by_last_name(self, client, fake_auth):
         calnet_users = list(calnet.get_calnet_users_for_uids(app, ['1081940']).values())
         last_name = calnet_users[0]['lastName']
         api_json = self._api_users_autocomplete(client, f' {last_name[:3]}  ')
+        print(api_json)
```
pauline2k: @cesarvh All looks good to me (the test failure on user search may have been a fluke based on randomized values) - we should just take this debug line back out before merge.

cesarvh: Ah, I was wondering what was going on there. Just pushed the change to remove the debug line

PR #3528, opened 2022-04-19
BOAC-4682: Changed null in progress units to 0 from {}

cesarvh:
https://jira-secure.berkeley.edu/browse/BOAC-4682


pauline2k:
Thanks, @cesarvh

PR #3559, opened 2022-09-09
BOAC-4737, add column is_ignored to degree_progress_categories; and end-to-end wiring

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4737

Release instructions updated.


pauline2k:
It's the end-to-end wiring you just can't ignore!

PR #3561, opened 2022-09-12
BOAC-4742, when provisioning student degree check, auto-create course when is_transfer_course

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4742


pauline2k:
Royal purple for transfer courses.

PR #3562, opened 2022-09-13
BOAC-4738, degree progress: show in-progress courses

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4738

Use `?w=1` for a simpler diff.


johncrossman:
cc: @lyttam @pauline2k

PR #3565, opened 2022-09-13
BOAC-4745, disable save/update note button immediately on-click

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4745


johncrossman:
cc: @lyttam @pauline2k

PR #3574, opened 2022-09-22
BOAC-4722: removes hard-coded 'L&S' from eForms

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4722

### DIFF
--- a/src/components/note/AdvisingNote.vue
+++ b/src/components/note/AdvisingNote.vue
@@ -8,7 +8,7 @@
         advisor {{ author.name }}<span v-if="note.topics && $_.size(note.topics)">: {{ oxfordJoin(note.topics) }}</span>
       </span>
       <span v-if="!note.subject && !$_.size(note.message) && !note.category && note.eForm" :id="`note-${note.id}-subject`">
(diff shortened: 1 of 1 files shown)


src/components/note/AdvisingNote.vue
```diff
@@ -8,7 +8,7 @@
         advisor {{ author.name }}<span v-if="note.topics && $_.size(note.topics)">: {{ oxfordJoin(note.topics) }}</span>
       </span>
       <span v-if="!note.subject && !$_.size(note.message) && !note.category && note.eForm" :id="`note-${note.id}-subject`">
-        eForm: L&amp;S Late Change of Schedule Request &mdash; {{ note.eForm.status }}
+        eForm: Late Change of Schedule Request &mdash; {{ note.eForm.status }}
```
johncrossman, line 11: @lyttam - do we want `:id="note-${note.id}-subject"`?  Should the suffix be "eform" instead?

lyttam: @johncrossman I think we did it that way because BOA treats eForms as notes, albeit with different fields.

PR #3583, opened 2022-09-29
BOAC-4673, identify course copies in UnassignedCourses w/ icon

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4673


pauline2k:
I would rather have had the icon be a Xerox machine, but software development is full of tradeoffs.

PR #3597, opened 2022-10-19
BOAC-4767, enrollment grading-basis dictates incomplete-grade description

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4767

See the Jira for full range of grading-basis codes.

### DIFF
--- a/src/mixins/Berkeley.vue
+++ b/src/mixins/Berkeley.vue
@@ -140,7 +140,16 @@ export default {
             }
             switch(statusCode) {
             case 'I':
-              const prefix = 'Incomplete grade scheduled to become a failing grade'
+              const gradingBasis = _.toUpper(section.gradingBasis)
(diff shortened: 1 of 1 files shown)


src/mixins/Berkeley.vue
```diff
@@ -140,7 +140,16 @@ export default {
             }
             switch(statusCode) {
             case 'I':
-              const prefix = 'Incomplete grade scheduled to become a failing grade'
+              const gradingBasis = _.toUpper(section.gradingBasis)
+              let outcome = 'failing grade'
```
pauline2k: @johncrossman Minor, but does the default string want to be `'a failing grade'`?

johncrossman: Nice catch @pauline2k. Amended.

PR #3616, opened 2023-01-03
BOAC-4820, squash Y2023 bug in BOA pytests

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4820


pauline2k:
In the year 2023, if BOA is alive to see…

PR #3623, opened 2023-01-24
BOAC-4759 Cohort filter for incomplete grades

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4759


tests/test_api/test_auth_controller.py
```diff
@@ -24,10 +24,10 @@
 """
 
 import json
+from unittest import mock
 
 from boac.models.user_login import UserLogin
 import cas
-import mock
```
pauline2k, line 30: My local machine wanted this import change before tests would pass. No idea why it hasn't been bothering Travis.

PR #3624, opened 2023-01-25
BOAC-4759 Alphabetize academic cohort filters

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4759


pauline2k:
We seem to be hitting another of these unrelated order-dependent test failures. Retrying Travis.

PR #3625, opened 2023-01-25
BOAC-4797 Front-end display for course requirements

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4797


johncrossman:
PR looks good. One lint-vue problem to fix. (The pytest fail is probably transient.)

PR #3626, opened 2023-02-02
BOAC-4812, minor py dependency updates

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4812

Why not.


johncrossman:
Transient Travis fail. Kicking off a retry.

PR #3638, opened 2023-02-17
BOAC-4838, fix search-history; disable search submit when searching

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4838

A PR I mean to get in yesterday. (Vacay violation.)


pauline2k:
Vacay violation but Travis approves.

PR #3654, opened 2023-03-17
BOAC-4867 Fix bug building date range SQL expression

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4867


johncrossman:
Glorious SQL @pauline2k!

PR #3659, opened 2023-04-07
BOAC-4875, bump BOA to Node v16, as required by CodeBuild upgrade

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4875


johncrossman:
More discussion needed.

PR #3664, opened 2023-04-13
BOAC-4880, more efficient approach to draft_notes schema support

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4880

Instead of `note_drafts` table, we add a column to the notes table: `is_draft_for_sids`.  This approach gives us more compatibility w/ existing code. For example, existing notes-CRUD code and the display in academic-timeline on /student page.  I am not using a boolean `is_draft` column because we need a way to map a single note to multiple SIDs and this felt like the least messy approach.

Also, `notes_fts_index` will ignore draft notes.


src/views/DraftNotes.vue
```diff
@@ -78,69 +59,120 @@
...
+      this.editingDraftNoteId = null
     },
     afterSaveDraftNote(data) {
-      const existing = this.$_.find(this.myNoteDrafts, ['id', this.editingNoteDraftId])
+      const existing = this.$_.find(this.myDraftNotes, ['id', this.editingDraftNoteId])
       Object.assign(existing, data)
-      this.editingNoteDraftId = null
+      this.editingDraftNoteId = null
+    },
+    openDeleteTopicModal(draftNote) {
```
lyttam: @johncrossman Should this be `openDeleteDraftNoteModal` instead?

johncrossman: Yes, I can amend.

johncrossman: Amended and waiting on Travis.

PR #3669, opened 2023-04-17
BOAC-4885, draft notes are complicated and horrible

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4885


pauline2k:
Can you check how Pinterest does it?

PR #3672, opened 2023-04-19
BOAC-4888, front-end: refactor edit-note transactions; edit-batch-note-modal for editing drafts

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4888

A work in progress. When editing a draft note, the user has the option of "publishing" it to one or more students via edit-batch-note modal. *Note:* Our schema limits drafts to zero or one SIDs but using a draft to create 2+ notes allows for unlimited SIDs so the modal must accommodate.

### DIFF
--- a/src/components/note/EditBatchNoteModal.vue
+++ b/src/components/note/EditBatchNoteModal.vue
@@ -1,5 +1,5 @@
 <template>
-  <div role="dialog">
+  <div v-if="mode" role="dialog">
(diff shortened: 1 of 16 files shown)


src/components/note/EditBatchNoteModal.vue
johncrossman, line 1: Use `?w=1` to get a more manageable diff.

PR #3684, opened 2023-04-25
BOAC-4893 USA-style date format in cohort range filters

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4893


pfarestveit:
American date formats are dumb.

PR #3719, opened 2023-05-03
BOAC-4916: excludes admits from batch note cohort and group selectors

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4916


pauline2k:
Reject!

PR #3723, opened 2023-05-05
BOAC-4933, show clippy icon if draft note has attachment

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4933

And a few minor/random fixes.


pauline2k:
Screen Shot 2023-05-05 at 12 10 44 PM

PR #3732, opened 2023-05-25
BOAC-4878: changes date range cohort filter to a date picker

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4878


pauline2k:
Date picker! Date picker!

PR #3738, opened 2023-05-31
BOAC-4942: hacks date range picker to be more accessible

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4942


johncrossman:
Hacks 'r us!

PR #3773, opened 2023-08-23
BOAC-4979, GA4 config must be injected earlier in the code-build cycle

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4979

Use `?w=1`

Fyi, I have put a GA4 configuration how-to in the 'BOA Ops Manual' doc.


pauline2k:
Thanks for the opsing!

PR #3785, opened 2023-10-09
BOAC-4982 Academic Career cohort filter

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-4982

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,22 +2,43 @@ sudo: required
 language: python
 python: "3.8"
 
+addons:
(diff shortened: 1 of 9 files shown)


pauline2k:
Travis is unhappy because it's still running tests under way obsolete Postgres 9, which doesn't include the `starts_with` function. I'm trying to work in an upgrade recipe.

pauline2k:
The Postgres 13 magic is accomplished.

PR #3803, opened 2024-01-19
BOAC-5017, no more piped date-parsing filters in template code; drop vue-moment

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5017

* `vue-moment` is what provided the filtering syntax which is not supported in Vue 3.
* Many modified Vue templates are drop-in appt related and will soon be removed


pauline2k:
It's a very consistent PR.

PR #3807, opened 2024-01-24
BOAC-5020, begin replace Vue.prototype.currentUser with vuex-based approach

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5020


pauline2k:
Get ready for pinia!

PR #3808, opened 2024-01-24
BOAC-5033: deletes (non-legacy) appointments functionality

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5033


johncrossman:
Nice work @lyttam.  How satisfying to see ~8000 lines of code dropped. BOA is feeling nimble again.

PR #3809, opened 2024-01-24
BOAC-5020, currentUser is now available via 'context' mixin

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5020

From the Vue 3 docs, "While mixins are still supported in Vue 3, the Composition API is generally considered to be the preferred approach for code reuse. The Composition API is more flexible, readable, and scalable than mixins, and it is also easier to debug."


pauline2k:
BOA just has a lot of front end. It's time to say it.

PR #3812, opened 2024-01-24
BOAC-5025, move util functions (eg, putFocusNextTick) off Vue.prototype

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5025


johncrossman:
@pauline2k or @lyttam - When you a sec, please review. Thx

PR #3825, opened 2024-01-31
BOAC-5042, initial refactor of store:curated-group

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5042

In the Vue 3 + Pinia world, mixins are discouraged. Thus, consider our design pattern of having mixins dispatch to Vuex actions to be deprecated. Alternatives:
* Invoke Vuex actions directly from components using `store.commit`. (In this PR, `init` function is an example.)
* Move the action function to a dedicated typescript file and import into your component. (In this PR, `goToCuratedGroup` is an example.)


johncrossman:
@lyttam - When you have a moment...

PR #3831, opened 2024-02-02
BOAC-5023: replaces vue-a11y announcer with homegrown version

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5023


johncrossman:
Thanks @lyttam !

PR #3839, opened 2024-02-07
BOAC-5010: sorts degree check courses alphabetically

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5010


src/components/degree/CoursesTable.vue
```diff
@@ -507,6 +508,7 @@ export default {
         this.alertScreenReader('Note hidden')
       }
     },
+    isAlertGrade,
```
lyttam, line 511: This fixes an unrelated error.

PR #3850, opened 2024-02-14
BOAC-5060, degree-progress: callback in CoursesTable needs object

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5060


johncrossman:
@lyttam - when you have a moment...

PR #3869, opened 2024-02-23
BOAC-5078: restores Util mixin

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5078

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -1,0 +1,59 @@
+import _ from 'lodash'
+import {nextTick} from 'vue'
+
+export function oxfordJoin(arr, zeroString) {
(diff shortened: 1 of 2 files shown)


boac/__init__.py
```diff
@@ -29,7 +29,7 @@
 from flask_sqlalchemy import SQLAlchemy
 from sqlalchemy.exc import SQLAlchemyError
 
-__version__ = '6.0'
+__version__ = '5.23'
```
johncrossman: @lyttam - A version rollback sneaked into the PR. :)

lyttam: oops! Amended.

PR #3873, opened 2024-02-23
BOAC-5082: restores DevAuth and BOAC-5089: Context mixin

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5082
https://jira-secure.berkeley.edu/browse/BOAC-5089

### DIFF
--- a/src/mixins/Context.vue
+++ b/src/mixins/Context.vue
@@ -1,0 +1,32 @@
+<script>
+import {mapActions, mapState} from 'pinia'
+import {useContextStore} from '@/stores/context'
+
+export default {
+  name: 'Context',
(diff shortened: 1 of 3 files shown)


src/mixins/Context.vue
```diff
@@ -0,0 +1,33 @@
+<script>
+import {mapActions} from 'pinia'
+import {mapState} from 'pinia'
```
johncrossman: @lyttam - Very minor and not a merge blocker: the two lines above can be combined into one?

lyttam: @johncrossman Oh right! Amended.

PR #3877, opened 2024-02-26
BOAC-5092: restores notes API and store

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5092

Also, adds a tsconfig (to resolve some import errors).

### DIFF
--- a/src/stores/note-edit-session/utils.ts
+++ b/src/stores/note-edit-session/utils.ts
@@ -1,0 +1,136 @@
+import {get, isString, map, trim} from 'lodash'
+import {deleteNote, removeAttachment, updateNote} from '@/api/notes'
+import {getDistinctSids} from '@/api/student'
+import {NoteEditSessionModel, NoteRecipients} from '@/store/note-edit-session/index'
+import {useContextStore} from '@/stores/context'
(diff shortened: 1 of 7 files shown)


lyttam:
@pauline2k would you mind reviewing?

src/stores/note-edit-session/utils.ts
```diff
@@ -0,0 +1,136 @@
...
+    }
+  })
+}
+
+export function isAutoSaveMode(mode: string | undefined): boolean {
+  return mode ? ['createBatch', 'createNote', 'editDraft'].includes(mode) : false
+}
+
+export function onVisibilityChange(): void {
+  const visibility: 'hidden' | 'visible' = document.visibilityState
```
pauline2k, line 37: @lyttam is a single pipe operator correct here or will it try to evaluate as a bitwise OR? (I'm not up on the details of TypeScript union types.)

pauline2k: ok, I have now read up on union types with string literals and see what's going on. :)

lyttam: @pauline2k  me too! This does seem to be correct because it was working (as far as I know) in the context of Vue 2.

PR #3901, opened 2024-03-01
BOAC-5016, you won't have the Matrix view to kick around anymore

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5016

image


config/default.py
```diff
@@ -110,8 +110,6 @@
 DATA_LOCH_STUDENT_SCHEMA = 'student'
 DATA_LOCH_TERMS_SCHEMA = 'terms'
 
-DISABLE_MATRIX_VIEW_THRESHOLD = 800
-
```
johncrossman, line 114: We do not override this default config so no release instruction necessary.

PR #3904, opened 2024-03-05
BOAC-5125, deal with Vite deprecation warning

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5125

_package.json_ changes:
* added `"type": "module"` (per Vite docs)
* removed extraneous entries (eg, private=false)


johncrossman:
Fixing...

johncrossman:
@pauline2k @lyttam - Travis is now happy.

PR #3905, opened 2024-03-05
BOAC-5104: /home curated groups, overrides font

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5104


johncrossman:
Nice! Thx @lyttam

PR #3906, opened 2024-03-06
BOAC-5121: vuetifies note templates menu

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5121


johncrossman:
Thanks for adding the notes to README file.

PR #3909, opened 2024-03-06
BOAC-5119: vuetifies RenameTemplateModal

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5119

### DIFF
--- a/src/plugins/vuetify.ts
+++ b/src/plugins/vuetify.ts
@@ -1,7 +1,7 @@
 import 'vuetify/styles'
 import {aliases, mdi} from 'vuetify/iconsets/mdi-svg'
 import {createVuetify} from 'vuetify'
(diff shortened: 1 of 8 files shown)


src/plugins/vuetify.ts
```diff
@@ -1,7 +1,7 @@
 import 'vuetify/styles'
 import {aliases, mdi} from 'vuetify/iconsets/mdi-svg'
 import {createVuetify} from 'vuetify'
-import {Resize} from 'vuetify/directives'
+import {Intersect, Resize} from 'vuetify/directives'
```
lyttam, line 4: Added this to resolve `[Vue warn]: Failed to resolve directive: intersect`. Apparently the `<v-text-field>` uses it.

PR #3915, opened 2024-03-08
BOAC-5128, tweak plugins/vuetify.ts for faster vue-build

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5128

See https://next.vuetifyjs.com/en/features/sass-variables/


johncrossman:
One fix needed

PR #3928, opened 2024-03-12
[vue2] BOAC-5143, revert stray h1 style causing 'BOA' shrinkage

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5143

This change was made in [commit 3c350338](https://github.com/ets-berkeley-edu/boac/commit/3c35033839c88d1724cafa21964d9e40a87dc804), in context of `/cohort` view work. Perhaps @lyttam can verify that this revert is okay.

### DIFF
--- a/src/assets/styles/boac-global.css
+++ b/src/assets/styles/boac-global.css
@@ -1,8 +1,3 @@
-h1 {
-  font-size: 24px !important;
-  font-weight: 700 !important;
-  margin: 0 0 10px 0 !important;
-}
 select {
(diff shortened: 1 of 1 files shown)


lyttam:
@johncrossman It will make all the other h1 headers bigger, but it's not the end of the world. Alternatively you could make sure the style only applies where it's needed. I'm not sure how it was working before.

EDIT: never mind, I checked and the font size of page headers is 24px in production.

PR #3949, opened 2024-03-15
[vue2] BOAC-5159, /degree_check: revert sorting in CourseTable

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5159


pauline2k:
Next time we'll try the Dewey decimal system!

PR #3957, opened 2024-03-18
boac-5130 - Profile View Complete

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5130

### DIFF
--- a/src/components/admin/DemoModeToggle.vue
+++ b/src/components/admin/DemoModeToggle.vue
@@ -10,23 +10,26 @@
           class="avatar student-avatar-large"
         />
       </div>
-      <div class="pl-2 pt-2">
+      <div class="pl-3 ">
         <div v-if="!_isNil(currentUser.inDemoMode)">
-          <b-form-checkbox
+          <v-checkbox
             v-if="!isToggling"
             id="toggle-demo-mode"
             v-model="inDemoMode"
             :disabled="isToggling"
+            :label="`${inDemoMode ? 'On' : 'Off'}`"
+            :aria-label="`Demo mode is ${inDemoMode ? 'On' : 'Off'}`"
             @change="toggle"
           >
-            <span class="sr-only">Demo mode is </span>{{ inDemoMode ? 'On' : 'Off' }}
-          </b-form-checkbox>
+          </v-checkbox>
           <div v-if="isToggling" class="demo-mode-label">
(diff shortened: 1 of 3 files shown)


src/components/admin/DemoModeToggle.vue
```diff
@@ -1,3 +1,5 @@
+Demo
```
pauline2k: @mohamalnadi Are these title rows outside the template a new convention for Vue 3 files, or were they meant to be temporary?

src/components/admin/DemoModeToggle.vue
```diff
@@ -10,23 +12,26 @@
...
             v-if="!isToggling"
             id="toggle-demo-mode"
             v-model="inDemoMode"
             :disabled="isToggling"
+            :label="`${inDemoMode ? 'On' : 'Off'}`"
             @change="toggle"
           >
-            <span class="sr-only">Demo mode is </span>{{ inDemoMode ? 'On' : 'Off' }}
-          </b-form-checkbox>
+            <!-- <span class="sr-only">Demo mode is </span>{{ inDemoMode ? 'On' : 'Off' }} -->
```
pauline2k: Commented-out screenreader text: intentional?

pauline2k:
Thanks @mohamalnadi! Looks good to me, a couple questions on specific lines. (Maybe @lyttam can take a look too since she's been closer to this work.)

src/components/admin/DemoModeToggle.vue
lyttam: I think we want either a `<label>` element separate from the `<v-checkbox>` _or_ `label` and `aria-label` properties on the `<v-checkbox>`. In the past I've noticed that using both a separate label and label properties causes the screen reader to voice it twice.

If you go with the second option, the `aria-label` property can be set to `Demo mode is ${inDemoMode ? 'On' : 'Off'}`.

mohamalnadi:
@pauline2k @lyttam - Made the new changes! Thank you! 

I am not sure where the word "Demo" came from outside of the template, I may have added it accidentally?

For the Demo On/Off label, I added an aria-label. The reason I removed the span is because in Vuetify, the nested span text aligned to the left of the checkbox, but you would have to use a label to have the text on the right of the checkbox.

lyttam:
Looks good, thanks!

PR #3979, opened 2024-04-02
BOAC-5145, elegant-date-picker with v-text-field slotted in

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5145

Proper date-picker in advanced search modal. We have aliased the https://vcalendar.io components with `elegant-date-picker`, etc. (Also sneaking in minor package updates.)


pauline2k:
It's the elegantest!

PR #4003, opened 2024-04-11
BOAC-5176, proper redirect after CAS login

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5176


johncrossman:
@pauline2k @lyttam - When you have a moment...

PR #4005, opened 2024-04-11
BOAC-5188, first pass at AcademicTimelineTable UX

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5188

### DIFF
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -96,6 +96,52 @@
   font-weight: bold;
   line-height: 1.2;
 }
+.pill {
+  background-color: #fff;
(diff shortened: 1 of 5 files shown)


pauline2k:
Pills!

src/assets/styles/global.css
```diff
@@ -96,6 +96,52 @@
   font-weight: bold;
   line-height: 1.2;
 }
+.pill {
```
lyttam, line 99: @johncrossman I don't think we have to keep these `pill` classes around. Vuetify's `v-chip` seems cleaner and more standardized.

PR #4010, opened 2024-04-12
BOAC-5176, responsive /login UX

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5176


src/layouts/Login.vue
```diff
@@ -89,92 +73,73 @@ export default {
...
+  color: #3b7ea5;
+  font-weight: 200;
+  font-size: 81px;
+  letter-spacing: 8px;
+  padding-top: 14px;
+}
+.airplane {
   background-color: #fff;
-  border: 10px solid #0275d8;
+  border: 10px solid #3b7ea5;
```
lyttam: @johncrossman You could replace `#3b7ea5` with `rgb(var(--v-theme-primary))`.

johncrossman: Thanks @lyttam. I'll amend.

lyttam: Same thing on lines 115 and 138.

PR #4025, opened 2024-04-24
BOAC-5206 - Fix Service announcement not showing up on log in

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5206

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -7,6 +7,7 @@ import {get, trim} from 'lodash'
 import {initializeAxios} from './utils'
 import {registerPlugins} from '@/plugins'
 import {useContextStore} from '@/stores/context'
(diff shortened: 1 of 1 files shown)


src/main.ts
```diff
@@ -7,6 +7,7 @@ import {get, trim} from 'lodash'
 import {initializeAxios} from './utils'
 import {registerPlugins} from '@/plugins'
 import {useContextStore} from '@/stores/context'
+import {getServiceAnnouncement} from './api/config'
```
johncrossman: @mohamalnadi - I recommend we use `@` to alias the `src` directory. Thus, the line above should have `@/api/config`

mohamalnadi: Updated to use `@`. thank you!

PR #4032, opened 2024-04-26
BOAC-5219, /cohort UX wrt filters w/ date and int ranges

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5219


mohamalnadi (approved):
thank you!

PR #4038, opened 2024-04-30
BOAC-5235, finish up tabbed search results page

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5235

### DIFF
--- a/src/components/admin/EditUserProfileModal.vue
+++ b/src/components/admin/EditUserProfileModal.vue
@@ -45,7 +45,7 @@
             <span class="font-weight-bolder">Error:</span> {{ error }}
           </div>
           <div v-if="!isExistingUser" class="align-items-center mb-3 ml-4 mt-3">
-            <label for="uid-input" class="sr-only">U I D </label>
+            <label for="uid-input" class="sr-only">U I D<span aria-hidden="true">&nbsp;</span></label>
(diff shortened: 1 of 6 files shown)


src/components/admin/EditUserProfileModal.vue
```diff
@@ -45,7 +45,7 @@
             <span class="font-weight-bolder">Error:</span> {{ error }}
           </div>
           <div v-if="!isExistingUser" class="align-items-center mb-3 ml-4 mt-3">
-            <label for="uid-input" class="sr-only">U I D </label>
+            <label for="uid-input" class="sr-only">U I D&nbsp;</label>
```
pauline2k: @johncrossman Out of curiosity, what does a non-breaking space do for the screen reader?

johncrossman: @pauline2k - In 3 minutes of internet research, I found differing opinions. I added the non-breaking space to fix broken copy/paste for these IDs. I'm going to try `white-space: nowrap` instead.

johncrossman: @pauline2k - I'll amend this PR. Actually, the `nowrap` style does not achieve the same fix. I'll wrap the non-breaking space with aria-hidden="true"

PR #4054, opened 2024-05-10
BOAC-5251, proper page title; plane-go-round spinner

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5251


pfarestveit:
Pan-Am-747-over-San-Francisco

PR #4058, opened 2024-05-14
BOAC-5228, /cohort: disable cohort filter option if already added

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5228


johncrossman:
@mohamalnadi or @pfarestveit - Ready review, if you have time.

PR #4069, opened 2024-05-21
BOAC-5257, date-picker v-calendar has both widget and text-input for manual entry

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5257


johncrossman:
@pauline2k - PR review when you have a moment.

PR #4077, opened 2024-05-24
BOAC-5275, get the pagination IDs right cuz people are people

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5275


pauline2k:
I can't understand what makes a `<span>` hate another `<span>`.

PR #4082, opened 2024-06-04
BOAC-5276, responsive search results (SortableStudents)

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5276


pauline2k:
If the component's importable, the students are sortable.

PR #4091, opened 2024-06-12
BOAC-5189, v-dialog: take advantage of Vue3 v-model per component

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5189


pfarestveit:
this.isMerging

PR #4101, opened 2024-06-25
BOAC-5271, /course page; response layout w/o page reload

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5271

Further understanding of Vue3 route (re)mounting means slight refactor in `loadingStart()` approach. More to come...


johncrossman:
@pauline2k - You're too complex.  (C901 'provision_course_site' is too complex)

PR #4107, opened 2024-06-27
BOAC-5326, use vuetify v-date-input instead of v-calendar

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5326


johncrossman:
@mohamalnadi or @pauline2k - when you have a minute, please review. Thx!

PR #4161, opened 2024-07-24
BOAC-5386, restore row and col colors on Pax Manifest

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5386

### DIFF
--- a/src/views/PassengerManifest.vue
+++ b/src/views/PassengerManifest.vue
@@ -1,29 +1,23 @@
 <template>
   <div v-if="!loading" class="default-margins">
-    <div class="list-group">
-      <div class="align-items-baseline d-flex mt-2">
-        <div class="pr-2">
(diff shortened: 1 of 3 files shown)


src/views/PassengerManifest.vue
```diff
@@ -1,29 +1,23 @@
...
-        </div>
-        <div class="flex-grow-1 text-right">
-          <EditUserProfileModal
-            :after-update-user="afterCreateUser"
-            :departments="departments"
-          />
-        </div>
+    <div class="align-center d-flex py-2">
+      <div class="pr-2">
+        <v-icon :icon="mdiContacts" :style="{color: '#3b7ea5'}" size="x-large" />
```
lyttam, line 5: @johncrossman I think you can use `color="primary"` (or maybe it's `class`) instead of `:style="{color: '#3b7ea5'}` here.

johncrossman: @lyttam - Indeed. Thank you! (That code was existing. I'll fix in my next PR.)

PR #4168, opened 2024-07-26
BOAC-5360: refreshes academic timeline when note created

lyttam:
I also finished the conversion of AcademicTimeline and AcademicTimelineTable to Vue3.


johncrossman:
Nice move to Composition API.

PR #4197, opened 2024-08-06
BOAC-5403: treats draft note pustules

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5403

I went ahead and also replaced all remaining instances of bootstrap classes `p-n` and `m-n` with `pa-n` and `ma-n` and fixed some other styles.

### DIFF
--- a/src/components/note/AdvisingNoteAttachments.vue
+++ b/src/components/note/AdvisingNoteAttachments.vue
@@ -73,14 +73,17 @@
           variant="outlined"
         >
           <span class="truncate-with-ellipsis">{{ attachment.displayName }}</span>
-          <template #close>
+          <template #append>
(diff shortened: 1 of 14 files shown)


src/components/note/AdvisingNoteAttachments.vue
```diff
@@ -73,14 +73,17 @@
           variant="outlined"
         >
           <span class="truncate-with-ellipsis">{{ attachment.displayName }}</span>
-          <template #close>
+          <template #append>
```
lyttam, line 76: This fixes the issue where we had a button nested inside a button.

johncrossman: Nice!

PR #4199, opened 2024-08-06
boac-5310 + boac-5361 + other bugs I found related to Curate groups

mohamalnadi:
Jiras: 

- https://jira-secure.berkeley.edu/browse/BOAC-5310
- https://jira-secure.berkeley.edu/browse/BOAC-5361

+ also fixes 1 or 2 curated group bugs I found on the way that I do not think we have Jiras for

### DIFF
--- a/src/api/curated.ts
+++ b/src/api/curated.ts
@@ -36,9 +36,9 @@ export function addStudentsToCuratedGroup(curatedGroupId: number, sids: string[]
 
 export function createCuratedGroup(domain: string, name: string, sids: string[]) {
   const url: string = `${utils.apiBaseUrl()}/api/curated_group/create`
-  return axios.post(url, {domain, name, sids}).then(function(data) {
-    $_onCreate(data)
-    return data
+  return axios.post(url, {domain, name, sids}).then(function(response) {
+    $_onCreate(response)
+    return response.data
   })
 }
 
--- a/src/views/CreateCuratedGroup.vue
+++ b/src/views/CreateCuratedGroup.vue
@@ -54,7 +54,7 @@ export default {
     },
     create(name) {
       this.showCreateModal = false
-      createCuratedGroup(this.domain, name, this.sids)
+      return createCuratedGroup(this.domain, name, this.sids)
         .then(group => {
           alertScreenReader(`Curated group '${name}' created. It has ${this.sids.length} students.`)
           this.isSaving = false
--- a/src/components/curated/dropdown/CuratedGroupSelector.vue
+++ b/src/components/curated/dropdown/CuratedGroupSelector.vue
@@ -63,8 +63,6 @@
               color="primary"
               density="compact"
               hide-details
-              @click="curatedGroupCheckboxClick(group)"
-              @keyup.enter="curatedGroupCheckboxClick(group)"
             >
               <template #label>
                 <span class="ml-2">
--- a/src/components/student/profile/TimelineDate.vue
+++ b/src/components/student/profile/TimelineDate.vue
@@ -29,6 +29,6 @@ const props = defineProps({
   }
 })
 const date = typeof props.date === 'string' ? new Date(props.date) : props.date
-const adjustedDate = date ? DateTime.fromJSDate(date).setZone(useContextStore().config.timezone) : null
+const adjustedDate = date ? DateTime.fromISO(props.date).setZone(useContextStore().config.timezone) : null
 const dateFormat = adjustedDate && adjustedDate.year === DateTime.now().year ? 'MMM d' : 'MMM d, yyyy'
 </script>


src/components/curated/CreateCuratedGroupModal.vue
```diff
@@ -123,7 +123,7 @@ export default {
         putFocusNextTick('create-cohort-input')
       } else {
         this.isSaving = true
-        this.create(this.name).then(() => {
+        this.create(this.name)?.then(() => {
```
johncrossman: @mohamalnadi - This question-mark seems out of place. What's the intention? That `create()` function will always return a `Promise`.

src/lib/cohort.ts
```diff
@@ -16,7 +16,7 @@ const isExistingName = (name: string, id?: number) => {
     each(cohorts, existing => {
       if (
         (!id || id !== existing.id) &&
-        name.toUpperCase() === existing.name.toUpperCase()
+        name?.toUpperCase() === existing?.name?.toUpperCase()
```
johncrossman: @mohamalnadi - We should be very careful with this null-tolerant approach. Sometimes a "variable undefined" error is what immediately highlights a problem. If the code fails silently due to the null-tolerance of that question-mark then finding the bug might take more time.

I would also argue that the logic becomes much harder to wrap your head around. The expression `name?.toUpperCase() === existing?.name?.toUpperCase()` has more than a few permutations of logic.

johncrossman:
@mohamalnadi - If you don't mind, please rethink the usage of null-tolerant `?`. A more verbose boolean expression might be better than a shorthanded expression having multiple `?` usages.

johncrossman:
@mohamalnadi - I'm suggesting you amend this PR to eliminate the null tolerance. Do you disagree?

mohamalnadi:
@johncrossman - Yes I am working on tracing back these lines of code to see why they are null. The reason I added the null tolerance is because I was getting errors in the browser console that they are undefined, which I wanted to eliminate. But as you said, there may be an underlying bug with the fact that they are undefined, so I am trying to see why that is the case.

src/lib/cohort.ts
mohamalnadi: It took me a while, but I finally figured out what the issue was. In some cases, the `existing` variable is directly a curatedgroup object and sometimes it seems to be wrapped in a `response` object. The new code I added should identify that and fail safe properly. 

The image below shows the difference: 
image

src/components/curated/CreateCuratedGroupModal.vue
mohamalnadi: Fixed this as well, the function that was passed in as a prop did not have a `return`.

mohamalnadi:
@johncrossman - please see above changes, thank you for your feedback!

src/lib/cohort.ts
johncrossman: Ah, well done @mohamalnadi. You identified the bug. I recommend a change at a lower level: [Line 41 in src/api/curated.ts](https://github.com/ets-berkeley-edu/boac/blob/dev/src/api/curated.ts#L41). As you say, a response object is being returned. The `createCuratedGroup` function at line 41 of `api/curated.ts` should be changed to return `response.data`. Once that change is made, you can change `CreateCuratedGroup` back to what it was: expecting a `curatedGroup` and not a response object. Make sense? Thank you for taking the time to investigate!

src/components/curated/CreateCuratedGroupModal.vue
johncrossman: Nicely done @mohamalnadi. You fixed the problem at its source.

mohamalnadi:
@johncrossman - made the changes, thank you for showing where the response is coming from!

src/api/curated.ts
```diff
@@ -38,7 +38,7 @@ export function createCuratedGroup(domain: string, name: string, sids: string[])
   const url: string = `${utils.apiBaseUrl()}/api/curated_group/create`
   return axios.post(url, {domain, name, sids}).then(function(data) {
     $_onCreate(data)
-    return data
+    return data.data
```
johncrossman: @mohamalnadi - Looks good. This bug was probably caused by an axios-related refactor I made months ago.... So, my bad! (For future reference, our variable naming convention should be `response` such that the function returns `response.data`.)

src/views/CreateCuratedGroup.vue
```diff
@@ -54,11 +54,11 @@ export default {
     },
     create(name) {
       this.showCreateModal = false
-      createCuratedGroup(this.domain, name, this.sids)
+      return createCuratedGroup(this.domain, name, this.sids)
         .then(group => {
           alertScreenReader(`Curated group '${name}' created. It has ${this.sids.length} students.`)
           this.isSaving = false
-          this.$router.push(`/curated/${group.id}`)
+          this.$router.push(`/curated/${group.data.id}`)
```
johncrossman: @mohamalnadi - Given your change in `src/api/curated.ts`, the above line can go back to being `group.id`. Once that change is made then I'll merge. 😄

mohamalnadi:
@johncrossman - made the changes, thank you!

PR #4204, opened 2024-08-08
BOAC-5393, /course page: featured student present and on top

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5393

Bonus: IDs on all `tr` and `td` elements.


pfarestveit:
A cornucopia of element ids!  Thanks @johncrossman

PR #4221, opened 2024-08-13
BOAC-5286, BOAC-5287. BOAC-5306, BOAC-5398 - various fixes for notes and passenger manifest

mohamalnadi:
Jiras:

- https://jira-secure.berkeley.edu/browse/BOAC-5286
- https://jira-secure.berkeley.edu/browse/BOAC-5287
- https://jira-secure.berkeley.edu/browse/BOAC-5306
- https://jira-secure.berkeley.edu/browse/BOAC-5398

### DIFF
--- a/src/components/admin/Users.vue
+++ b/src/components/admin/Users.vue
@@ -511,9 +511,11 @@ export default {
   },
   methods: {
     handleSort(sortBy) {
-      this.sortBy = sortBy[0].key
-      this.sortDesc = sortBy[0].order === 'asc' ? false : true
-      this.usersProvider() // Method to fetch users with new sorting parameters
+      if (sortBy.length) {
+        this.sortBy = sortBy[0].key
+        this.sortDesc = sortBy[0].order === 'asc' ? false : true
+        this.usersProvider() // Method to fetch users with new sorting parameters
+      }
     },
     clickColumn(slotData) {
       const indexExpanded = this.expanded.findIndex(i => i === slotData)
@@ -846,5 +848,9 @@ export default {
 .name {
   position: relative;
   display: inline-block;
+}
+
+:deep(.v-table > .v-table__wrapper > table > thead > tr > th) {
+  height: 0px !important;
 }
 </style>
--- a/src/components/note/AdvisingNote.vue
+++ b/src/components/note/AdvisingNote.vue
@@ -1,6 +1,10 @@
 <template>
   <div :id="`note-${note.id}-outer`" class="advising-note-outer w-100">
-    <div :id="`note-${note.id}-is-closed`" :class="{'note-snippet-when-closed truncate-with-ellipsis': !isOpen}" aria-label="Advising note">
+    <div :id="`note-${note.id}-is-closed`"
+         class="ellipses-for-long-text"
+         :class="{'note-snippet-when-closed truncate-with-ellipsis': !isOpen}"
+         aria-label="Advising note"
+    >
       <span v-if="note.isDraft" :id="`note-${note.id}-is-draft`">
         <span class="pr-2">
           <v-chip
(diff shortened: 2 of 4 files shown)


src/components/admin/Users.vue
```diff
@@ -511,9 +511,11 @@ export default {
   },
   methods: {
     handleSort(sortBy) {
-      this.sortBy = sortBy[0].key
-      this.sortDesc = sortBy[0].order === 'asc' ? false : true
-      this.usersProvider() // Method to fetch users with new sorting parameters
+      if (sortBy.length) {
+        this.sortBy = sortBy[0]?.key
+        this.sortDesc = sortBy[0]?.order === 'asc' ? false : true
+        this.usersProvider() // Method to fetch users with new sorting parameters
```
johncrossman, line 517: @mohamalnadi - I'm concerned with the null-checks here. Are we getting a non-empty `sortBy` array containing nulls? If so, perhaps we have a problem upstream from this code. Is the input to this function coming from `v-data-table`? As before, we should be very cautious with null-tolerant question-marks.

src/components/admin/Users.vue
```diff
@@ -847,4 +849,8 @@ export default {
   position: relative;
   display: inline-block;
 }
+
+:deep(.v-table > .v-table__wrapper > table > thead > tr > th) {
+  height: 0px !important;
+}
```
johncrossman, line 855: @mohamalnadi - Formatting of `th` elements (table-header cells in `v-data-table`) can be done using the `header-props` component prop. Elaborate CSS is often necessary but, in this case, I'd suggest using `header-props`.

src/components/note/AdvisingNote.vue
```diff
@@ -309,4 +313,10 @@ loadAuthorDetails()
   text-overflow: ellipsis;
   white-space: nowrap;
 }
+.ellipses-for-long-text {
+  width: 100%;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
```
johncrossman: @mohamalnadi - Minor indentation fix needed above. Thx.

src/components/search/AdvancedSearchModal.vue
```diff
@@ -317,6 +319,17 @@ const cancel = () => {
   setTimeout(reset, 100)
 }
 
+const focusOnDate = (type) => {
+  if (type === 'from') {
+    const element = document.getElementById('search-options-note-filters-last-updated-from')
+    element.focus()
+  } else if (type === 'to') {
+    const element = document.getElementById('search-options-note-filters-last-updated-to')
+    element.focus()
```
johncrossman: @mohamalnadi - Thank you for considering accessibility here. Fyi, we have `putFocusNextTick` (utility function) that could be used in place of the raw script above.

johncrossman:
Thank you @mohamalnadi. I've added a few comments/questions. Let me know if anything is unclear.

src/components/admin/Users.vue
mohamalnadi: @johncrossman - yes, I just removed the null-checks, they technically do not do anything as that part of the code only runs if sortBy has a length of more than 0.

src/components/admin/Users.vue
mohamalnadi: This CSS is only relevant when the table is in a loading state. I do not think header-props will work in this case.

src/components/note/AdvisingNote.vue
mohamalnadi: fixed!

src/components/search/AdvancedSearchModal.vue
mohamalnadi: replaced with `putFocusNextTick `, thank you!!

mohamalnadi:
@johncrossman - please see my updated changes and comments, thank you for you feedback!

src/components/admin/Users.vue
johncrossman: Got it. Thanks @mohamalnadi

johncrossman:
Thank you @mohamalnadi

PR #4240, opened 2024-08-16
boac-5401 - Draft Node Edit Modal Should react to ESC

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5401


pauline2k:
Looks good!

PR #4243, opened 2024-08-16
BOAC-5496, fix logic of degreeStore.isUserDragging: course

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5496


pfarestveit:
Sounds like a drag.

PR #4245, opened 2024-08-16
BOAC-5477: updates focus mgmt, aria tags in all  modals

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5477

I wound up fixing some focus management in the cohort header, curated group header, and degree template as well.


johncrossman:
Very nice. Thank you @lyttam  (And Bea updates, too.)

PR #4249, opened 2024-08-19
BOAC-5483: replaces overlays with dialogs, standardizes styles

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5483


johncrossman:
Standardization across all modals. I like it! Thanks @lyttam

PR #4270, opened 2024-08-23
BOAC-5517 / BOAC-5527, admit cohorts do not have opt-groups in primary dropdown; enable 'Add' for ranges

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5517
https://jira-secure.berkeley.edu/browse/BOAC-5527


lyttam:
Tricky!

PR #4272, opened 2024-08-23
boac-5499 - UI for searching for notes and appointments with your department

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5499

### DIFF
--- a/src/components/search/AdvancedSearchModal.vue
+++ b/src/components/search/AdvancedSearchModal.vue
@@ -109,6 +109,13 @@
                         value="you"
                         @change="() => searchStore.setAuthor(null)"
                       />
+                      <v-radio
+                        id="search-options-note-filters-posted-by-your department"
+                        :disabled="searchStore.isSearching"
+                        :ischecked="searchStore.postedBy === 'yourDepartment'"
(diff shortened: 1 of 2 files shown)


src/components/search/AdvancedSearchModal.vue
```diff
@@ -368,6 +375,13 @@ const search = () => {
       } else if (searchStore.author) {
         query.advisorCsid = searchStore.author.sid
         query.advisorUid = searchStore.author.uid
+      } else if (searchStore.postedBy === 'yourDepartment') {
+        if (currentUser.departments && currentUser.departments.length > 0) {
+          query.departmentCodes = []
+          currentUser.departments.forEach((department) => {
+            query.departmentCodes.push(department.code)
```
pauline2k: @mohamalnadi logic looks good, but these three lines could be more tersely expressed as a `map` one-liner:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map

mohamalnadi: Done! thank you for your feedback!

PR #4286, opened 2024-08-27
BOAC-5537, much better page-load time for 'Draft Notes' page

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5537

On localhost w/ 900 draft notes, page-load time went from 24s to <1s.


lyttam:
Very nice!

PR #4288, opened 2024-08-27
BOAC-5522, delete draft-note modal tolerates empty subject and/or student

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5522


pauline2k:
So tolerant!

PR #4294, opened 2024-08-28
BOAC-5512, alignment on class page with multiple course sites

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5512

Review w/ `?w=1`


lyttam:
Paging Travis.

PR #4295, opened 2024-08-28
BOAC-5523: replaces date picker, sets text field defaults

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5523


pauline2k:
So long, `<v-date-input>`! Welcome back, `<date-picker>`!

PR #4302, opened 2024-08-29
BOAC-5547, permalink from search takes you to /student note, open and in view

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5547

No need for third-party package. And when we scroll to permalink we do not want default put-focus-on-h1.


src/components/student/profile/AcademicTimelineTable.vue
```diff
@@ -474,25 +473,20 @@ const init = () => {
...
+        return true
       }
+    })
+    if (obj) {
+      isShowingAll.value = true
+      nextTick(() => {
+        open(obj, true)
+        scrollToPermalink(permalink.messageType, permalink.messageId)
+      })
     }
```
johncrossman, line 490: Use `?w=1` for simpler diff.

PR #4307, opened 2024-09-03
BOAC-5563, both mixins/Util and mixins/Context are retired

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5563

And more than a few components cut over to the Composition API. No more mixins for BOA.


pauline2k:
It's more than a feeling, more than a few!

PR #4319, opened 2024-09-04
BOAC-5558: restores keyboard navigability to autocomplete suggestions

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5558

I applied the same fixes to other instances of Autocomplete. Includes a few other changes like replacing the deprecated Axios cancel token with an AbortController.


johncrossman:
@lyttam - Thank you!

PR #4322, opened 2024-09-04
boac-5501: Query Notes from sis with user department(s)

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5501


pauline2k:
Looks good!

PR #4331, opened 2024-09-05
[QA] BOAC-5172 Safeguard Nessie against horrendously long search strings

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-5172


johncrossman:
10 names is all you get. Looks good. Waiting on T.

PR #4349, opened 2024-09-06
BOAC-5571, max-width for select degree-template menu

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5571

### DIFF
--- a/src/views/degree/StudentDegreeCreate.vue
+++ b/src/views/degree/StudentDegreeCreate.vue
@@ -17,13 +17,12 @@
             id="degree-template-select"
             v-model="selectedOption"
             aria-label="Select a degree template"
-            class="select-menu"
+            class="select-menu select-menu-max-width"
             :disabled="isSaving"
           >
             <option
               id="degree-check-select-option-null"
               :value="null"
-              @select="onChangeSelect"
             >
               Choose...
             </option>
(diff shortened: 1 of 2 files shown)


src/views/degree/BatchDegreeCheck.vue
```diff
@@ -146,7 +146,7 @@
         <ProgressButton
           id="batch-degree-check-save"
           :action="save"
-          class="ms-auto mr-2"
```
lyttam: @johncrossman I think we need to keep this margin so there is space between the buttons (you don't notice the need for space until visual focus is on the `variant="text"` button, causing the outline to show). I've been putting class `ml-2` on the Cancel button in each case so that if the buttons wrap, they won't appear out of vertical alignment.

src/views/degree/StudentDegreeCreate.vue
```diff
@@ -17,48 +17,44 @@
...
+              :title="option.name"
               :value="option"
-              @select="onChangeSelect"
             >
               {{ option.name }}
             </option>
           </select>
         </div>
         <div class="align-center d-flex mt-6">
-          <div class="mr-1">
```
lyttam: Similar thing here; since you're removing the unnecessary div wrappers from each button, I would add `class="ml-2"` to the Cancel button to keep some space between them.

johncrossman:
Thank you @lyttam. I reverted those two changes.

johncrossman:
@lyttam - Amended. Fyi, these buttons don't wrap in narrow viewport. And if they did, the margin-left would be zero for both.

PR #4352, opened 2024-09-10
BOAC-4953: adds progress indicator, other SR text for note attachments

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-4953

Also, the screen reader alert was completely missing from the page (though visible in the footer).

### DIFF
--- a/src/components/note/AdvisingNoteAttachments.vue
+++ b/src/components/note/AdvisingNoteAttachments.vue
@@ -3,10 +3,12 @@
     <slot name="label"></slot>
     <v-file-input
       v-if="!readOnly"
-      :id="`${idPrefix}choose-file-for-note-attachment`"
+      :id="inputId"
       ref="attachmentFileInput"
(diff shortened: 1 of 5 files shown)


src/components/note/AdvisingNoteAttachments.vue
```diff
@@ -166,21 +168,50 @@ const modelProxy = ref(props.note || noteStore.model)
 const attachmentLimitReached = computed(() => {
   return size(model.value.attachments) >= contextStore.config.maxAttachmentsPerNote
 })
+const inputId = computed(() => {
+  return `${props.idPrefix}choose-file-for-note-attachment`
+})
+const progressBarId = computed(() => {
+  return `${props.idPrefix}note-attachment-progress`
+})
```
johncrossman, line 176: @lyttam  - I don't think the two vars above need to be `computed`. They won't change. Not a merge blocker tho.

PR #4361, opened 2024-09-11
BOAC-5417: addresses accessibility in Pax Manifest

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5417


johncrossman:
Thank you @lyttam. Looks good.

PR #4365, opened 2024-09-12
BOAC-5578 - Update "Invalid" message for Units in Degree Check

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5578


mohamalnadi:
@johncrossman @lyttam - can you please review when you get the chance? thank you!

PR #4369, opened 2024-09-12
BOAC-5531, /search notes: if loch query is not necessary, we still need total_count from the loch

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5531


johncrossman:
@lyttam @mohamalnadi - When you have a sec, would you mind reviewing / merging?

mohamalnadi (approved):

PR #4374, opened 2024-09-13
BOAC-5245: consolidates colors in student profile, classes, degree checks

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5245

FYI, I'm changing most of the variants of grey text to `text-medium-emphasis`. This will make it much easier to change all instances if we need to darken the color for contrast reasons.


johncrossman:
Very nice.

PR #4378, opened 2024-09-13
BOAC-5607, python null should not put 'None' in service announcement text

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5607


pauline2k:
We don't want to advertise our vacancy.

PR #4395, opened 2024-09-18
boac-5598 - Passenger Manifest: editing a user's role doesn't update table view dynamically

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5598


lyttam:
Travis build failed on installing NPM. I re-ran it and it passed.

PR #4404, opened 2024-09-19
BOAC-5628, curated-group used by cohort when not owned by current-user

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5628

This error occurs when user-A views a curated-group owned by user-B and that curated-group is used by a cohort of user-B. Phew!


pauline2k:
🙃

PR #4405, opened 2024-09-19
boac-5583 - update test_users to have first and last names instead of being randomly generated

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5583

This should hopefully fix the occasional test failures. Thank you @lyttam for looking into it and walking me through this!


johncrossman:
@mohamalnadi - Nice choice of test data. Linter is unhappy. Otherwise, looks good.

mohamalnadi:
@johncrossman oops, fixed and yup, agreed! the names were all from @lyttam

PR #4408, opened 2024-09-19
BOAC-3816: turns student table sortable headers into buttons

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-3816

I also put aria tags on the tablist and tabpanels so the screen reader associates each panel with its tab.

### DIFF
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -277,6 +277,9 @@
 .student-teams-container {
   margin-top: 5px;
 }
+.table-striped table tbody tr:nth-of-type(odd) {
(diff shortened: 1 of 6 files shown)


src/assets/styles/global.css
```diff
@@ -309,6 +312,18 @@
     width: 100%;
   }
 }
+.v-table-hidden-row-override table tbody tr[style="height: 0px; border: 0px;"] {
+  display: none;
```
lyttam, line 316: Vuetify tables add this empty first row that's only visually hidden. This should fix the issue where JAWS reads the wrong row count (it still counts the header as a row, but that might just be a JAWS issue).

PR #4418, opened 2024-09-23
BOAC-3809: fixes acad timeline accessibility bugs, labels student page regions

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-3809

The extra, sr-only Edit and Delete buttons in timeline notes were confusing to JAWS somehow. It thought every element after those buttons was a delete button (e.g. tabbing to the advisor name link and typing Enter would bring up the delete confirm modal).

I don't think the invisible buttons are necessary anymore. JAWS can find the visible ones just fine.


pauline2k:
Can't argue with fewer elements on the page.

PR #4421, opened 2024-09-24
BOAC-5648: ensures 1 h1 per page; fixes axe findings

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5648

Axe Devtools was complaining about missing aria labels and content not contained in regions.


johncrossman:
Thank you @lyttam!

PR #4424, opened 2024-09-24
BOAC-5646, less complexity in AdvisingNoteAttachments

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5646

### DIFF
--- a/src/views/DraftNotes.vue
+++ b/src/views/DraftNotes.vue
@@ -106,34 +106,34 @@
         {{ currentUser.isAdmin ? 'No' : 'You have no' }} saved drafts.
       </div>
     </div>
+    <AreYouSureModal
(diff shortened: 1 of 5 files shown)


src/views/DraftNotes.vue
```diff
@@ -106,34 +106,34 @@
...
-  />
+    </AreYouSureModal>
+    <EditBatchNoteModal
+      v-model="isEditDialogOpen"
+      initial-mode="editDraft"
+      :note-id="get(selectedNote, 'id')"
+      :on-close="afterEditDraft"
+      :sid="get(selectedNote, 'sid')"
+    />
+  </div>
```
johncrossman, line 136: The diff above is only an indentation change.

PR #4430, opened 2024-09-25
BOAC-5630: workaround for date picker keyboard interaction skipping months

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5630


johncrossman:
Nothing wrong with a little workaround.

PR #4465, opened 2024-10-03
BOAC-5673: makes student profile units & GPA more responsive

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5673


mohamalnadi:
The build failed originally, re-ran it and looks good!

PR #4481, opened 2024-10-07
BOAC-5674: prevents focus from landing on nonexistent pagination

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-5674

### DIFF
--- a/src/views/Cohort.vue
+++ b/src/views/Cohort.vue
@@ -184,8 +184,6 @@ const getFocusElementId = isBackButtonToCohort => {
     if (matches) {
       focusId = `student-${matches[0]}`
     }
-  } else if (cohortStore.pagination.currentPage) {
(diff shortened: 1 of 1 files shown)


src/views/Cohort.vue
```diff
@@ -184,8 +184,6 @@ const getFocusElementId = isBackButtonToCohort => {
     if (matches) {
       focusId = `student-${matches[0]}`
     }
-  } else if (cohortStore.pagination.currentPage) {
-    focusId = `pagination-page-${cohortStore.pagination.currentPage}`
```
lyttam, line 188: If pagination is on the page, its own focus management will supersede that of Cohort. vue so we don't need to try to do it here. There are scenarios where the pagination component is not present: if the number of students in the cohort falls below the threshold, or if we're on the Create a Cohort page.

PR #4493, opened 2024-10-14
boac-5686 - Note privacy is missing, and existing privacy isn't respected

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5686


pauline2k:
Looks like a fix to me. Thanks @mohamalnadi!

PR #4545, opened 2024-11-06
BOAC-5732, filter topics on Flight Deck

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5732


src/components/topics/ManageTopics.vue
```diff
@@ -74,6 +76,7 @@
               <v-btn
                 v-if="item.deletedAt"
                 v-bind="props"
+                :id="`undelete-topic-${normalizeId(item.topic)}`"
```
johncrossman, line 79: IDs on buttons, for BEA.

PR #4547, opened 2024-11-06
boac-5731 - Remove BOA Environment text from the footer alert

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5731


johncrossman:
@mohamalnadi - Random thought: we could leave the text and put a `sr-alert` class on the text so it's only available to screen-readers?!

PR #4618, opened 2024-12-20
BOAC-4968, 'Academic standing' option added to Export List (CSV)

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4968

### DIFF
--- a/boac/api/util.py
+++ b/boac/api/util.py
@@ -29,7 +29,7 @@ import json
 
 from boac.api.errors import BadRequestError, ResourceNotFoundError
 from boac.externals.data_loch import get_admitted_students_by_sids, get_sis_holds, get_student_profiles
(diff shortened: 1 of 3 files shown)


boac/api/util.py
```diff
@@ -426,41 +426,38 @@ def _response_with_students_csv_download(sids, fieldnames, benchmark, term_id):
     term_id_last = previous_term_id(current_term_id())
     term_id_previous = previous_term_id(term_id_last)
     getters = {
```
johncrossman, line 428: The following 40 lines were ordered by key, thus the noisy diff. The `academic_standing` entry is the only new entry.

boac/api/util.py
```diff
@@ -516,9 +513,20 @@ def _row_for_csv(result):
     )
```
johncrossman, line 515: The only new method below is `_get_academic_standing`.

PR #4625, opened 2024-12-27
BOAC-5782 - Remove ellipses from some course codes + increase "large" screen threshold

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5782

QA PR: https://github.com/ets-berkeley-edu/boac/pull/4626


mohamalnadi:
I changed the "large" screen threshold from 1280px to 1400px as, in some cases where the course code is more than 13 characters, the columns in the classes tables were bunching on top of each other. This change fixes this issue. This way we can have longer course codes without ellipses and not have the columns be on top of each other.

PR #4628, opened 2025-01-03
BOAC-4859, degree check link from cohort/group listing page

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-4859

Includes BEA update.


pfarestveit:
Thanks for the BEA update @johncrossman !

PR #4637, opened 2025-01-07
BOAC-5795, /degree_progress: respect line breaks in category descriptions

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5795

`<pre>` tags do not play nice with `v-html`

### DIFF
--- a/src/views/degree/ManageDegreeChecks.vue
+++ b/src/views/degree/ManageDegreeChecks.vue
@@ -157,14 +157,15 @@
                 @click="cancelEdit"
               />
             </div>
-            <div v-if="item.id !== get(templateForEdit, 'id')" class="align-center d-flex flex-wrap justify-end">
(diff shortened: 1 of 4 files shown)


src/views/degree/ManageDegreeChecks.vue
```diff
@@ -200,10 +203,11 @@
                 <span class="text-disabled" role="separator">|</span>
                 <v-btn
                   :id="`degree-check-${item.id}-delete-btn`"
+                  class="font-size-14 pl-1"
                   color="primary"
                   :disabled="isBusy"
+                  size="x-sm"
```
lyttam, line 209: @johncrossman Does this need to be `"x-small"` or is `x-sm` an alias?

johncrossman: @lyttam - Yes, `x-sm` is an alias.

PR #4647, opened 2025-01-09
boac-5789 - Degree check link now looks like a link vs. a button

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5789

### DIFF
--- a/src/components/student/StudentRowBioColumn.vue
+++ b/src/components/student/StudentRowBioColumn.vue
@@ -126,8 +126,9 @@
       <v-btn
         :id="`degree-checks-of-${student.sid}`"
         color="primary"
-        size="sm"
         variant="text"
+        size="sm"
+        tag="a"
         @click="() => goToStudentDegreeChecksByUID(student.uid)"
       >
         <div class="align-center d-flex text-anchor">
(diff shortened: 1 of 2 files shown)


lyttam:
@mohamalnadi Sorry, the ticket wasn't really clear. The element doesn't need to change visually. The underlying HTML tag just needs to be `<a>` instead of `<button>` so that screen readers identify it as a link. I think you can accomplish this just by giving the `<v-btn>` an `href` attribute instead of the `@click` handler.  See https://vuetifyjs.com/en/components/buttons/#props-href

mohamalnadi:
@lyttam - no worries! So, the `goToStudentDegreeChecksByUID` method makes an endpoint call which we use to get the path to route to in a new tab, so href would technically not work since the path is calculated on click of the v-btn. I can change it so that the path is calculated on initialization (`onMounted`) of the component, but that would take some time and I am not sure if worth it for the time being, but I can definitely do it if need be. For now, I added an `a` tag in v-btn, not sure if that would help with JAWS identifying it as a link.

lyttam:
@mohamalnadi That approach might not work if the `<a>` is nested inside a `<button>`, but if you instead add `tag="a"` to the `<v-btn>` component that should change the button to an anchor without changing how it looks.

mohamalnadi:
@lyttam - thank you, this is good to know! Didn't realize you can just add a tag like that on a `v-btn`. I just made the change.

PR #4660, opened 2025-01-13
BOAC-5825 Precision placement for expanded course tabs

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-5825


pauline2k:
Travis is unable to check out code for a run right now, but my local tox comes out green for go.

PR #4661, opened 2025-01-13
[QA] BOAC-5825 Precision placement for expanded course tabs

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-5825

QA version of #4660.


pauline2k:
As with the dev PR, this commit is locally tox-approved even though Travis can't verify.

PR #4726, opened 2025-01-28
BOAC-5859, initial work on PAM /dashboard (auth, front-end route, view)

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5859


johncrossman:
@mohamalnadi or @lyttam - PR needs review (when you have time)

PR #4749, opened 2025-02-04
BOAC-5895, new university_dept: School of Public Health (KTPUB)

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5895


johncrossman:
Release instructions updated.

PR #4752, opened 2025-02-05
boac-5842 - Complete migration scripts and schema work for peer advising

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5842


pauline2k:
@mohamalnadi SQL looks great, thanks for this! Looks like an extra merge commit has snuck into your history - let us know if you'd like to confer on the best way to clean that up.

mohamalnadi:
@pauline2k thank you! Yes I am having some trouble cleaning up that extra commit. Maybe we can meet after stand up?

PR #4760, opened 2025-02-07
BOAC-5876, use defineModel in EditUserProfileModal; refactor to structure of user.departments

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5876

Type checking `EditUserProfileModal` helps enforce structure of `user` model-object.


src/api/user.ts
```diff
@@ -2,13 +2,14 @@ import axios from 'axios'
 import utils from '@/api/api-utils'
 import {get, isNil} from 'lodash'
 import {useContextStore} from '@/stores/context'
+import {BoaUser} from '@/lib/utils'
```
johncrossman, line 5: Much of the diff is due to renaming of 'CurrentUser' type.

PR #4763, opened 2025-02-10
boac-5898 - Note-templates CRUD for peer advising notes

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-5898

### DIFF
--- a/boac/api/note_templates_controller.py
+++ b/boac/api/note_templates_controller.py
@@ -39,6 +39,7 @@ from flask_login import current_user
 def create_note_template():
     params = request.get_json()
     note_id = params.get('noteId', None)
+    peer_advising_department_id = params.get('peer_advising_department_id', None)
     title = params.get('title', None)
     if not note_id or not title:
         raise BadRequestError('Invalid or missing parameters')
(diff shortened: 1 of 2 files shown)


boac/api/note_templates_controller.py
```diff
@@ -77,6 +79,13 @@ def get_my_note_templates():
     return tolerant_jsonify([t.to_api_json() for t in note_templates])
 
 
+@app.route('/api/note_templates/peer_advising_department_id/<peer_advising_department_id>')
+@advising_data_access_required
+def get_my_note_templates_for_peer_advising(peer_advising_department_id):
```
pauline2k: @mohamalnadi Minor, but we probably shouldn't have `_my_` in this method name, since the template association isn't with the current user.

mohamalnadi: @pauline2k updated the method name to `get_note_templates_for_peer_advising_department`, thank you!

pauline2k:
@mohamalnadi Looks good! Not as a merge blocker, but as a follow-up, we should think about access control for these new controller methods. In general, for any CRUD operations associated with peer advising departments, we probably want to limit access to people who have a db association with that department, and some operations should be limited specifically to managers.

PR #4775, opened 2025-02-13
BOAC-5913, Pax Man: user can have Peer Advising role in max one department

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5913


pfarestveit:
The Pax Man cometh.

PR #4783, opened 2025-02-14
BOAC-5917 Bring back the subplans

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-5917

Thanks @johncrossman for the Vue 3 crash course.


johncrossman:
I'm happy to help. I don't want front-end code to ruin your Friday.

PR #4790, opened 2025-02-18
BOAC-5879, no more mock data in /api/peer/department/<id>; front-end renders w/ real data

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5879

### DIFF
--- a/boac/models/note.py
+++ b/boac/models/note.py
@@ -33,7 +33,6 @@ from boac.models.base import Base
 from boac.models.note_attachment import NoteAttachment
 from boac.models.note_template_attachment import NoteTemplateAttachment
 from boac.models.note_topic import NoteTopic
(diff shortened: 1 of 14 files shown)


boac/models/note.py
```diff
@@ -83,8 +82,6 @@ class Note(Base):
         back_populates='note',
         lazy=True,
     )
-    note_template = db.relationship('NoteTemplate', back_populates='notes')
-    peer_advising_department = db.relationship(PeerAdvisingDepartment.__name__, back_populates='notes')
```
johncrossman, line 87: @mohamalnadi - I removed the `db.relationship` entries above and one below because it might impede performance. We don't want our Peer Advising work to slow down BOA's otherwise ordinary use cases.

PR #4793, opened 2025-02-18
BOAC-5928, PAM dashboard: get 'topic' out of element IDs

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5928


pfarestveit:
image

PR #4804, opened 2025-02-20
BOAC-5920, pytest to verify that /cachejob/refresh does not drop Peer Advisors

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5920


johncrossman:
@mohamalnadi @pfarestveit - Travis is happy now.

PR #4805, opened 2025-02-20
BOAC-5929, Pax Man: CSV downloads handles all flavors of filter and search

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5929


pfarestveit:
flavors

johncrossman:
airplane

PR #4819, opened 2025-02-26
boac-5935 - Peer Advising Note templates

mohamalnadi:
Successfully rebased! Minor conflicts resolved. 

**Jiras:**
- https://jira-secure.berkeley.edu/browse/BOAC-5898
- https://jira-secure.berkeley.edu/browse/BOAC-5899
- https://jira-secure.berkeley.edu/browse/BOAC-5900
- https://jira-secure.berkeley.edu/browse/BOAC-5935


boac/api/note_templates_controller.py
```diff
@@ -68,12 +67,7 @@ def get_note_template(note_template_id):
     note_template = NoteTemplate.find_by_id(note_template_id=note_template_id)
     if not note_template:
         raise ResourceNotFoundError('Template not found')
-    if not note_template.peer_advising_department_id and note_template.creator_id != current_user.get_id():
-        raise ForbiddenRequestError('Template not available')
-    if (note_template.peer_advising_department_id
-            and not PeerAdvisingDepartment.user_in_peer_advising_department(
-                user=current_user,
-                peer_advising_department_id=note_template.peer_advising_department_id)):
+    if note_template.creator_id != current_user.get_id():
```
johncrossman, line 70: @mohamalnadi - Peer Advisors will use note-templates created by PAMs. The if statement above will not allow that sharing to happen. We need a more fine-tuned auth check: If the `creator is not current_user` then the creator (PAM) and current_user (Peer Advisor) must both be members of the same `peer_advising_department`.

boac/api/peer_advising_notes_controller.py
```diff
@@ -64,6 +67,64 @@ def get_peer_advising_topics():
     return tolerant_jsonify([topic.to_api_json() for topic in PeerAdvisingTopic.get_all()])
 
 
+@app.route('/api/peer_advising/note_template/<note_template_id>')
+@login_required
```
johncrossman: @mohamalnadi - Two questions:
1. Is this API endpoint functionally the same as `/api/note_template/<note_template_id>`, with different auth checks? In this case, I would lean towards having a single API endpoint (in `note_templates_controller`) and then code the necessary authentication checks.
2.  The `@login_required` is overly permissive.  See my comment (above) in `note_templates_controller.py`.

boac/api/peer_advising_notes_controller.py
```diff
@@ -64,6 +67,64 @@ def get_peer_advising_topics():
...
+    note_templates = NoteTemplate.get_templates_created_by_peer_advising_department(peer_advising_department_id=peer_advising_department_id)
+    return tolerant_jsonify([t.to_api_json() for t in note_templates])
+
+
+@app.route('/api/peer_advising/note_template/update', methods=['POST'])
+@peer_advisor_manager_required
+def update_peer_advising_note_template():
+    params = request.get_json() or {}
+    note_template_id = params.get('id', None)
+    body = params.get('noteBody', None)
```
johncrossman: @mohamalnadi - Here is a minor concern: Instead of `noteBody` as the param name, I recommend `body`. Because that aligns with the front-end convention and we have a lot of shared code on the front-end.

boac/api/peer_advising_notes_controller.py
```diff
@@ -64,6 +67,64 @@ def get_peer_advising_topics():
...
+def delete_peer_advising_note_template(note_template_id):
+    note_template = NoteTemplate.find_by_id(note_template_id=note_template_id)
+    if not note_template:
+        raise ResourceNotFoundError('Template not found')
+    if not PeerAdvisingDepartmentMember.is_user_in_peer_advising_department(
+            user_id=current_user.get_id(),
+            peer_advising_department_id=note_template.peer_advising_department_id):
+        raise ForbiddenRequestError('Template not available')
+    NoteTemplate.delete(note_template_id=note_template_id)
+    return tolerant_jsonify({'message': f'Note template {note_template_id} deleted'}), 200
```
johncrossman, line 125: @mohamalnadi - I believe this is another case of functional equivalence across API endpoints. In other words, the endpoint above is equivalent to `/api/note_template/delete/<note_template_id>` in _note_templates_controller_. Is that true?  I understand that the auth checks are tricky (considering the range of use cases) but I still believe we should not have multiple API endpoints that have functional equivalence. What do you think?

src/components/peer/PeerAdvisingNoteTemplateModal.vue
```diff
@@ -0,0 +1,165 @@
...
+    createPeerAdvisingNoteTemplate(props.peerAdvisingDeptId, noteDetailsText.value, templateName.value, topicsSelected.value).then((newNoteTemplate) => {
+      model.value = false
+      isSaving.value = false
+      emit('note-template-updated', newNoteTemplate)
+      alertScreenReader(`Created ${newNoteTemplate.title} note template.`)
+    })
+  }
+
+}
+</script>
```
johncrossman, line 165: @mohamalnadi - The template code above looks good.  Going forward, I recommend we put all Vue components related to peer-advising notes in the directory `src/components/peer/note`.  You do not need to make that move now. Just a tip for future reference.

boac/api/peer_advising_notes_controller.py
mohamalnadi: @johncrossman 

1. Yes, this, along with delete are the same functionally, I put it here since we also have `create_peer_advising_note_template` and `update_peer_advising_note_template` in the `peer_advising_notes_controller.py`. Those 2 latter endpoints work a bit different functionally, which is why I decided to separate them out from the standard `note_template` endpoints, as it would get pretty convoluted there. Do you think `create_peer_advising_note_template` and `update_peer_advising_note_template` should remain in `peer_advising_notes_controller.py` and this endpoint (`get`) as well as the `delete` endpoint should be deleted and we stick to ones in `note_templates_controller`?

2. I am unsure what the alternative would be? Would it be `peer_advisor_or_peer_advisor_manager`? I think that would work if it is in the `peer_advising_notes_controller.py`. But if we move this endpoint back to the `note_templates_controller`, then that would not work since non-peer advisors/manager would also need to access these GET endpoints.

src/components/peer/PeerAdvisingNoteTopics.vue
```diff
@@ -101,6 +106,10 @@ onMounted(() => {
       })
     })
   }
+
+  if (props.topics) {
+    topicsSelected.value = props.topics
+  }
```
johncrossman: @mohamalnadi - Not sure if I'm reading this right... I think `if (props.topic) ...`  will always be true because its default value is `[]`.  Perhaps it doesn't matter. 😄

johncrossman:
Thank you @mohamalnadi. This is good work. I've posted questions & comments. I'll merge when amendments (if necessary) are made.

src/components/peer/PeerAdvisingNoteTopics.vue
mohamalnadi: Ahh, yes that is correct. I originally did not have a default value assigned, hence that if block, but then added the default value later when the linter complained. I'll remove that if statement!

mohamalnadi:
@johncrossman - thank for your feedback! I will make the changes for the `body` and the if statement.

But please let me know if you think having **GET** and **DELETE** in the `note_templates_controller` and **CREATE** and **UPDATE** in `peer_advising_notes_controller` is the best option or if all of them should remain in `peer_advising_notes_controller`. The decorator permissions for the GET endpoint will have to change based on which controller it resides in.

boac/api/peer_advising_notes_controller.py
johncrossman: @mohamalnadi - Thanks for the quick reply.  For question 1 above, I'd say:
* In a future PR, let's discard `/api/peer_advising/note_template/create` in favor of `/api/note_template/create`. (Do you mind creating a Jira for that task?)
* In this PR, let's discard '/api/peer_advising/note_template/<note_template_id>' and use `/api/note_template/<note_template_id>` instead. Of course, your peer-advising auth check should be copied over.
* Lastly, I do think we should use the existing `update` and `delete` note-template API endpoints in `note_templates_controller`.  

We have solid test coverage on all of the existing notes and note-templates related API endpoints. When it makes sense, we should reuse existing APIs and just extend the existing test-coverage to deal with peer-advising use cases.

Let's revisit question 2 after we make decisions on question 1.

johncrossman:
@mohamalnadi - Let me know if you'd like to have a quick Zoom call. Having a conversation via Github is a bit tough. 😄

boac/api/peer_advising_notes_controller.py
johncrossman: @mohamalnadi - I'm ready to merge this PR. Did you want to make the change above or shall I proceed w/ merge?

mohamalnadi: @johncrossman  - just made the change! I also changed `@login-required `to `@peer_advisor_or_peer_advisor_manager` for `get_peer_advising_note_template`.

johncrossman:
@mohamalnadi - The python linter failed. One more amendment and then I'll merge.

mohamalnadi:
@johncrossman sorry about that, just fixed it!

PR #4822, opened 2025-02-27
BOAC-5957, python API decorators to dedicated decorators.py

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5957

Our growing number of peer-advising decorators put us close to 1000 lines in `api/utils.py`


pfarestveit:
decorate

PR #4829, opened 2025-03-04
BOAC-5953, /peer_advisor/home: forgot to add component in previous PR

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5953


pauline2k:
womp womp 🎺

PR #4868, opened 2025-03-13
BOAC-6030, debounce=500 for v-comboboxes; EditNote: disable subject if peerAdvising

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6030

The `EditAdvisingNote.vue` is simple if reviewed with `?w=1`.


pfarestveit:
The more debounce de better

title

PR #4871, opened 2025-03-13
BOAC-6019, PAM edits a PA-authored note; v-html in case of special chars

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6019
https://jira-secure.berkeley.edu/browse/BOAC-6036


pfarestveit:
Pam loves her Pa!

PR #4881, opened 2025-03-19
BOAC-6038, /student: Peer department info per note

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6038

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -17,6 +17,7 @@ import {useContextStore} from '@/stores/context'
 import vuetify from '@/plugins/vuetify'
 import router from '@/router'
 import axiosPlugin from '@/plugins/axios'
+import {getDepartments} from '@/api/user'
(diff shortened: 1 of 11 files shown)


src/main.ts
```diff
@@ -49,6 +50,8 @@ axios.get(`${apiBaseUrl}/api/profile/my`).then(response => {
 
   axios.get(`${apiBaseUrl}/api/config`).then(response => {
     contextStore.setConfig({...response.data, apiBaseUrl, isVueAppDebugMode})
+    // Async API calls, whenever possible.
+    getDepartments().then(data => contextStore.setAllBerkeleyDepartments(data))
```
johncrossman, line 54: The line above is the cornerstone behind this rather low-risk refactor. The dept and peer-dept metadata is needed in many components and loading (async) that data during SPA init phase is best.

PR #4887, opened 2025-03-20
boac-6042 - Complete Peer Advising Search Implementation

mohamalnadi:
**Jira:**
- https://jira-secure.berkeley.edu/browse/BOAC-6042

### DIFF
--- a/src/components/peer/note/PeerAdvisingNoteDetails.vue
+++ b/src/components/peer/note/PeerAdvisingNoteDetails.vue
@@ -1,57 +1,62 @@
 <template>
   <div>
-    <div>
-      <div v-if="note.subject" :id="`note-${note.id}-subject`">{{ note.subject }}</div>
-      <div :id="`note-${note.id}-body`" v-html="note.body" />
+    <div v-if="!noteDetails">
+      Loading...
     </div>
-    <div :id="`note-${note.id}-is-open`" class="w-100">
-      <div v-if="note.subject && note.body" class="open-note-message-container pt-2">
-        <span :id="`note-${note.id}-message-open`" v-html="note.body" />
+    <div v-if="noteDetails">
+      <div>
+        <div v-if="noteDetails.subject" :id="`note-${noteDetails.id}-subject`">{{ noteDetails.subject }}</div>
+        <div :id="`note-${noteDetails.id}-body`" v-html="noteDetails.body" />
       </div>
-      <div class="mt-1">
-        <div v-if="note.author.name || note.author.email">
-          <span class="sr-only">Note created by </span>
-          <span v-if="note.author.uid && note.author.name">
-            <router-link
-              v-if="currentUser.isAdmin && note.peerAdvisingDepartmentId"
-              :id="`note-${note.id}-link-to-peer-advisor-home`"
-              :to="`/peer_advisor/${note.author.uid}/home`"
-            >
-              {{ note.author.name }}
-            </router-link>
-            <a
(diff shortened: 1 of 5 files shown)


src/components/peer/note/PeerAdvisingNoteDetails.vue
```diff
@@ -59,21 +64,48 @@
...
-    required: true,
-    type: Object as PropType<Note>
+    required: false,
+    type: Object as PropType<Note>,
+    default: undefined
+  },
+  noteId: {
+    required: false,
+    type: Number,
+    default: undefined
```
johncrossman, line 86: @mohamalnadi - First, I don't see this new `noteId` prop being used. What is the intention? Second, it's confusing that both `note` and `noteId` are not required props. I recommend we have a single (required) prop: `note`. If the invoking component only has `noteId` then make an API call to fetch the note _before_ you mount this component. Then we don't need the internal `noteDetails` object. Thoughts?

src/components/peer/note/PeerAdvisorNoteSearchResults.vue
```diff
@@ -80,7 +80,7 @@
                     @click="toggleShowHide(note)"
                   />
                 </div>
-                <PeerAdvisingNoteDetails class="my-3" :note="note" />
+                <PeerAdvisingNoteDetails class="my-3" :noteId="note.id" />
```
johncrossman, line 83: @mohamalnadi - Okay, I see its usage here. Performance-wise, this will be a problem. If we have 50 notes in search  results then this will result in 50 API calls to `getNote` by ID. Why not pass the entire `note` object and do away with the `noteId` prop?

src/views/PeerAdvisorSearch.vue
```diff
@@ -56,6 +47,10 @@ const route = useRoute()
 const router = useRouter()
 const totalNoteCount = ref(0)
 
+const resultWord = computed(() => {
+  return totalNoteCount.value === 1 ? 'result' : 'results'
+})
+
```
johncrossman: @mohamalnadi - The computed variable above makes sense. However, if you use the `pluralize` util then the code will be cleaner. Thus (1) remove the computed var above and (2) instead of `resultWord` (on line 7), use `pluralize('result', totalNoteCount)`

src/views/PeerAdvisorSearch.vue
```diff
@@ -2,14 +2,10 @@
...
-      </div>
-      <div v-if="!currentUser.isAdmin">
-        <EditPeerAdvisingNoteModal
-          v-if="noteStore.isCreateNoteModalOpen"
-          v-model="createNoteModal"
-          :peer-advising-department-id="peerAdvisingDepartmentId"
-        />
+        <h1>Peer Advising Notes</h1>
+        <div>
+          Showing {{ totalNoteCount }} {{ resultWord }} for <span class="font-weight-bold">{{ searchStore.queryText }}</span> | <span class="cursor-pointer text-blue-accent-2 select-none" @click="clearResults"> Clear Search Results </span>
```
johncrossman: @mohamalnadi - A couple changes needed here:
* A clickable `<span>` element is bad practice because (1) it's unreachable if you're a keyboard-only user, (2) a screen-reader will not identify it as a button (ie, interactive element). Thus, it's invisible to some users.
* Reloading the page in order to clear search results is an accessibility no-no. In fact, it will seem jarring to most users.

How about we convert that `<span>` to a button and then modify the `clearResults` function? I think we can "clear results" by simply setting `notes`, `totalNoteCount`, `offset`, etc. to _undefiend_.

PR #4893, opened 2025-03-21
boac-6042 - Complete Peer Advising Search Implementation with feedback

mohamalnadi:
**Jira:**

https://jira-secure.berkeley.edu/browse/BOAC-6042

### DIFF
--- a/src/components/peer/note/PeerAdvisingNoteDetails.vue
+++ b/src/components/peer/note/PeerAdvisingNoteDetails.vue
@@ -1,58 +1,63 @@
 <template>
   <div>
-    <div>
-      <div v-if="note.subject" :id="`note-${note.id}-subject`">{{ note.subject }}</div>
-      <div :id="`note-${note.id}-body`" v-html="note.body" />
+    <div v-if="!noteDetails">
+      Loading...
     </div>
-    <div :id="`note-${note.id}-is-open`" class="w-100">
-      <div v-if="note.subject && note.body" class="open-note-message-container pt-2">
-        <span :id="`note-${note.id}-message-open`" v-html="note.body" />
+    <div v-if="noteDetails">
+      <div>
+        <div v-if="noteDetails.subject" :id="`note-${noteDetails.id}-subject`">{{ noteDetails.subject }}</div>
+        <div :id="`note-${noteDetails.id}-body`" v-html="noteDetails.body" />
       </div>
-      <div class="mt-1">
-        <div v-if="note.author.name || note.author.email">
-          <span class="sr-only">Note created by </span>
-          <span v-if="note.author.uid && note.author.name">
-            <router-link
-              v-if="currentUser.isAdmin && note.peerAdvisingDepartmentId"
-              :id="`note-${note.id}-link-to-peer-advisor-home`"
-              :to="`/peer_advisor/${note.author.uid}/home`"
-            >
-              {{ note.author.name }}
-            </router-link>
-            <a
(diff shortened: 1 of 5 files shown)


mohamalnadi:
@johncrossman - Thank you for your feedback on my last PR! I closed my last PR as there was some merge conflicts after I made my changes and the commits got all messed up. 

Please see my comments below, thank you!

1. I added` /api/peer_advising/note/<note_id>` to the `peer_advising_notes_controller.py` this function gets peer advising note by ID and ensures that the user is **1. a peer advisor** and **2. has the correct peer advising department id**. I did this because the `get_note` in the `notes_controller.py` has the `@advising_data_access_required` decorator which will not work for a peer advisor
2. I fixed the span/button issue for "Clear Search Results" as you suggested
3. I kept the `noteId` prop for `PeerAdvisingNoteDetails.vue`. When, let's say, 30 search results come back, the `getNoteById` endpoint does NOT get called 30 times. Rather, it is only called when the user clicks on the note to expand it. I had to take this approach as the search endpoint does not return all the details of a note (e.g. does not return attachments and topics as well as other fields). So we display the data from the search result endpoint in the `PeerAdvisorNoteSearchResults.vue` and then once a user clicks on a note, we `getNoteById` and display all of the note's information. 
4. In the situation where `PeerAdvisingNoteDetails.vue` is not used in search (i.e. it is used by `PeerAdvisorPaginatedNotes.vue,` for example), we pass in the `note` as a prop and not the `noteId` and that will not call any endpoint, and `PeerAdvisingNoteDetails.vue` will just display the data that was passed through in the prop.

I hope that makes sense, please let me know if you have any other feedback!

src/components/peer/note/PeerAdvisingNoteDetails.vue
```diff
@@ -1,80 +1,112 @@
...
+            <span v-if="noteDetails.author.uid && noteDetails.author.name">
+              <router-link
+                v-if="currentUser.isAdmin && noteDetails.peerAdvisingDepartmentId"
+                :id="`note-${noteDetails.id}-link-to-peer-advisor-home`"
+                :to="`/peer_advisor/${noteDetails.author.uid}/home`"
+              >
+                {{ noteDetails.author.name }}
+              </router-link>
+              <a
+                v-if="!currentUser.isAdmin || !noteDetails.peerAdvisingDepartmentId"
```
johncrossman, line 27: @mohamalnadi - `noteDetails.peerAdvisingDepartmentId` will always be defined, right? I don't think it'll impact the way things render but perhaps the conditionals are extraneous.

src/components/peer/note/PeerAdvisingNoteDetails.vue
```diff
@@ -1,80 +1,112 @@
...
 })
 
 const currentUser = useContextStore().currentUser
+const noteDetails = ref()
+const noteLoaded = ref(false)
+
+onMounted(() => {
+  if (!props.note) {
+    getNoteById()
+  } else if (props.note) {
```
johncrossman, line 97: @mohamalnadi - I think the line above can just be an `else` without additional check on `props.note`.

johncrossman:
@mohamalnadi - Thank you for taking the time to put this new PR up. It looks good. I made a couple minor suggestions. Do you want to amend this PR or include changes in future PR?

mohamalnadi:
@johncrossman - thank you for reviewing my PR! I can make these changes in my next PR.

PR #4921, opened 2025-03-26
BOAC-6023, PAM reports: lazy-load the historical notes-related stats

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6023

Break off new API endpoint for historical notes report.


johncrossman:
Where is Travis? For the record, a local run of `tox` did pass, all green.

PR #4926, opened 2025-03-26
BOAC-6056, db/migrate SQL script to add all Peer Advising topics

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6056

Release instructions updated.

@pfarestveit  - I added the two new PA topics to BEA code, too.
- Change of College
- Change of Major


pfarestveit:
Thanks @johncrossman !

PR #4937, opened 2025-04-02
boac-6065 + boac-6064 + boac-6078 - PA: search now searches student name as well + templates fixes

mohamalnadi:
Jiras:
- https://jira-secure.berkeley.edu/browse/BOAC-6064
- https://jira-secure.berkeley.edu/browse/BOAC-6065
- https://jira-secure.berkeley.edu/browse/BOAC-6078

1. PA Search now searches the student names as well as note body
2. Took out Peer Advising Search out of `Note.search` and created a new method `Note.peer_advising_search` to simplify the search method as it was getting way too convoluted. 
3. PA advising templates - now you cannot have duplicate topics while editing a template. 
4. Added **Back to Home** button in search page

### DIFF
--- a/boac/models/note.py
+++ b/boac/models/note.py
@@ -403,6 +403,77 @@ class Note(Base):
         return api_json
 
     @classmethod
+    def peer_advising_notes_search(
+            cls,
+            search_phrase,
+            peer_advising_department_id,
+            matching_sids,
+            offset=0,
+            limit=40,
+    ):
+        notes_selector = """SELECT id, ts_rank(fts_index, plainto_tsquery('english', :search_phrase)) AS rank
+            FROM notes_fts_index
+            WHERE fts_index @@ plainto_tsquery('english', :search_phrase)"""
+        params = {
+            'search_phrase': search_phrase,
+        }
+
+        # Build a union branch that picks notes for which the student's SID is in the matching list.
+        student_branch = """
+            SELECT n.id, 0 AS rank
+            FROM notes n
+            WHERE n.sid = ANY(:matching_sids)
+            AND n.peer_advising_department_id = :peer_advising_department_id
+        """
+        params['matching_sids'] = matching_sids
+
+        # Combine both branches using UNION.
+        fts_selector = f"""
+            {notes_selector}
+            UNION
+            {student_branch}
+        """
+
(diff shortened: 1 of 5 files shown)


boac/models/note.py
```diff
@@ -402,6 +402,78 @@ def get_notes_report(cls):
...
+        fts_selector = f"""
+            {notes_selector}
+            UNION
+            {student_branch}
+        """
+
+        peer_advising_department_notes_filter = ''
+        if peer_advising_department_id:
+            peer_advising_department_notes_filter = 'AND notes.peer_advising_department_id = :peer_advising_department_id'
+            params.update({'peer_advising_department_id': peer_advising_department_id})
```
johncrossman: @mohamalnadi - In method above, `peer_advising_department_id` will always be defined (ie, non-null). And with that in mind, (1) should the `student_branch` query above include `peer_advising_department_id`, and (2) there's no need for if check above?

mohamalnadi: @johncrossman - thank you for your feedback! I made the changes and fixed the unit testing failing.

src/views/PeerAdvisorSearch.vue
```diff
@@ -76,7 +76,7 @@ const init = (user: BoaUser) => {
     alertScreenReader(`Searching for "${searchStore.queryText}"`)
     peerAdvisorSearch(searchStore.queryText, peerAdvisingDepartmentId.value).then(data => {
       notes.value = data.notes.reverse()
-      totalNoteCount.value = data.totalNoteCount
+      totalNoteCount.value = data.notes.length
```
johncrossman: @mohamalnadi - Is `notes.length` a reliable measure of `totalNoteCount`? If the backend query is using offset/limit then maybe `totalNoteCount` > `notes.length`.

johncrossman:
@mohamalnadi - My review is done. My most recent comment is the only one that needs your attention. Thanks again!

src/views/PeerAdvisorSearch.vue
mohamalnadi: Yes good catch, I reverted this change. There is a separate bug where the `totalNoteCount` does not return the correct amount of results - I'll create a separate bug for this. Thank you!

PR #4958, opened 2025-04-08
boac-6072 + boac-6065 - PA Search tweeks to improve search on student names and partial words

mohamalnadi:
Jiras:
- https://jira-secure.berkeley.edu/browse/BOAC-6072
- https://jira-secure.berkeley.edu/browse/BOAC-6065
- https://jira-secure.berkeley.edu/browse/BOAC-6076

### DIFF
--- a/boac/models/note.py
+++ b/boac/models/note.py
@@ -413,9 +413,11 @@ class Note(Base):
             offset=0,
             limit=40,
     ):
-        notes_selector = """SELECT id, ts_rank(fts_index, plainto_tsquery('english', :search_phrase)) AS rank
+        notes_selector = """
+            SELECT id, ts_rank(fts_index, to_tsquery('english', :search_phrase || ':*')) AS rank
             FROM notes_fts_index
-            WHERE fts_index @@ plainto_tsquery('english', :search_phrase)"""
+            WHERE fts_index @@ to_tsquery('english', :search_phrase || ':*')
+        """
         params = {
             'search_phrase': search_phrase,
         }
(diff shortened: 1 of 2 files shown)


boac/models/note.py
```diff
@@ -413,9 +413,11 @@ def peer_advising_notes_search(
             offset=0,
             limit=40,
     ):
-        notes_selector = """SELECT id, ts_rank(fts_index, plainto_tsquery('english', :search_phrase)) AS rank
+        notes_selector = """
+            SELECT id, ts_rank(fts_index, to_tsquery('english', :search_phrase || ':*')) AS rank
             FROM notes_fts_index
-            WHERE fts_index @@ plainto_tsquery('english', :search_phrase)"""
+            WHERE fts_index @@ to_tsquery('english', :search_phrase || ':*')
+        """
```
johncrossman, line 420: @mohamalnadi - For the sake of my understanding, can you explain why we want `to_tsquery` instead of `plainto_tsquery `? And why the fallback to `':*'`?

mohamalnadi: @johncrossman - `plainto_tsquery` only searches full words. Paul reported a bug where if a note had, for example the note "LS body 1743778442" and the user searched for "1743", `plainto_tsquery` would return no results. `to_tsquery` along side `':*' ` is more expansive, and as such, would return results.

This is the particular bug this change is addressing: https://jira-secure.berkeley.edu/browse/BOAC-6072

johncrossman:
Thanks @mohamalnadi

PR #4968, opened 2025-04-09
BOAC-6104, /student AcadTimeline: note null check after cancel delete-note

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6104

### DIFF
--- a/src/components/note/AdvisingNote.vue
+++ b/src/components/note/AdvisingNote.vue
@@ -180,10 +180,6 @@ const props = defineProps({
     required: true,
     type: Function
   },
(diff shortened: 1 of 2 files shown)


src/components/note/AdvisingNote.vue
```diff
@@ -180,10 +180,6 @@ const props = defineProps({
     required: true,
     type: Function
   },
-  deleteNote: {
-    required: true,
-    type: Function
-  },
```
johncrossman, line 186: This prop was unused.

PR #4999, opened 2025-04-15
boac-6132 - Advanced Search: improve partial-string result matching

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-6132

**A couple of changes:**
1. use `to_tsquery` vs. `plainto_tsquery` which is more inclusive, as has been done with Peer Advising Search 
2. Since there are some recent changes with having multiple search_phrases, the `*` operator that works with `to_tsquery` had to be appended to each token in the search_phrases
3. `to_tsquery` runs into errors with with this search phrase `horse; <- epitaph? ->`. So I had to update this test case to fix that. If this sort of search phrase is expected from the user, then I do not think `to_tsquery` will work for advanced search.
4. Additionally, the original regex caused 4 of the unit tests to fail with regards to special characters. The new regex does not cause this issue and with my testing, I see consist and more inclusive results. However, we would need to do more thorough testing to ensure this does accidentally exclude any searches or cause any errors.

### DIFF
--- a/boac/lib/util.py
+++ b/boac/lib/util.py
@@ -41,7 +41,7 @@ from titlecase import titlecase
 
 """Generic utilities."""
 
-TEXT_SEARCH_PATTERN = r'(\w*[.:/-@]\w+([.:/-]\w+)*)|[^\s?!(),;:.`]+'
+TEXT_SEARCH_PATTERN = r'[0-9A-Za-z/\-]+'
 
 
 def camelize(text):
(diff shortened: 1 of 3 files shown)


pauline2k:
Looks good to me! I'll merge, and once deployed we can do some follow-up comparisons on dev versus qa to see if there are discrepancies in search results. We'll be particularly interested in special characters and older SIS appointments that contain a lot of HTML entities, since I think the gnarly regex was originally written to handle those.

PR #5007, opened 2025-04-16
BOAC-6146: allows admin to download PA notes CSV

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6146

I refactored the authorization on peer advising endpoints so that each time we're checking if a user is a peer advisor manager, we're also making sure they belong to the right department.

@johncrossman @mohamalnadi Is this a good idea? I assume PAMs should not be allowed to request data for a department they aren't a member of.

### DIFF
--- a/boac/api/decorators.py
+++ b/boac/api/decorators.py
@@ -25,8 +25,10 @@ ENHANCEMENTS, OR MODIFICATIONS.
 
 from functools import wraps
 
+from boac.api.auth_utils import is_authorized_peer_advisor_manager
 from boac.api.util import can_access_admitted_students
 from boac.lib.berkeley import has_any_membership_role, is_peer_advisor, is_peer_advisor_manager
+from boac.models.peer_advising_department import PeerAdvisingDepartment
 from boac.models.peer_advising_department_member import PeerAdvisingDepartmentMember
 from boac.routes import login_manager
 from flask import current_app as app, request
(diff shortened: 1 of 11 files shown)


tests/test_api/test_peer_advising_users_controller.py
```diff
@@ -120,6 +120,8 @@ def test_authorized(self, client, fake_auth):
         # Log in PAM of CoE Mech Eng
         peer_advisor_manager = fake_auth.login(coe_mech_peer_advisor_manager_uid)
         assert is_peer_advisor_manager(peer_advisor_manager)
+        from flask import current_app
+        current_app.logger.error(peer_advising_department.id)
```
pauline2k: @lyttam is this error logging left over from development?

boac/api/decorators.py
```diff
@@ -187,6 +189,41 @@ def _peer_advisor_or_peer_advisor_manager(*args, **kw):
     return _peer_advisor_or_peer_advisor_manager
 
 
+def peer_advisor_manager_in_department(func):
+    # Checks if the peer_advisor_manager is in the peer_advising_department in the request
```
pauline2k, line 193: @lyttam restricting access by department membership makes sense to me!

tests/test_api/test_peer_advising_users_controller.py
lyttam: @pauline2k  oops, yes I'll remove it

boac/api/decorators.py
```diff
@@ -187,6 +189,41 @@ def _peer_advisor_or_peer_advisor_manager(*args, **kw):
...
+                and (
+                    current_user.is_admin
+                    or is_peer_advisor_manager(current_user)
+                    or _api_key_ok())
+                and (
+                    current_user.is_admin
+                    or is_authorized_peer_advisor_manager(
+                        peer_advising_department_id=peer_advising_department.id,
+                        peer_advisor_manager_user_id=current_user.get_id(),
+                    )
```
johncrossman, line 214: @lyttam - Looks good to me. Can the two parenthesized conditions above be combined into one:
```
current_user.is_admin
or is_authorized_peer_advisor_manager(
    peer_advising_department_id=peer_advising_department.id,
    peer_advisor_manager_user_id=current_user.get_id(),
)
```
The `is_peer_advisor_manager` check seems redundant.

lyttam: @johncrossman Yes, good idea. Amended.

PR #5008, opened 2025-04-16
boac-6138 - Fix issue with Deleted PA's not showing up in endpoint

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-6138

### DIFF
--- a/boac/models/authorized_user.py
+++ b/boac/models/authorized_user.py
@@ -377,36 +377,36 @@ def _peer_advising_users_sql(
         'role_type': role_type,
     }
     if peer_advising_department_id and role_type:
-        query_tables += """
+        query_tables += f"""
             JOIN peer_advising_departments d ON
                 d.id = :peer_advising_department_id
             JOIN peer_advising_department_members m ON
                 m.peer_advising_department_id = d.id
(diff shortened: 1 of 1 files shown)


boac/models/authorized_user.py
```diff
@@ -377,36 +377,36 @@ def _peer_advising_users_sql(
...
-                AND m.deleted_at IS NULL
+                {_users_sql_where_clause_peer_advising_member(status)}
         """
     elif not peer_advising_department_id and not role_type:
-        query_tables += """
+        query_tables += f"""
             JOIN peer_advising_department_members m ON
                 m.authorized_user_id = u.id
-                AND m.deleted_at IS NULL
+                {_users_sql_where_clause_peer_advising_member(status)}
```
johncrossman: @mohamalnadi - Looks good. Although I think you should change the name of  the method `_users_sql_where_clause_peer_advising_member` because it's not always in the context of peer advisors. Perhaps __get_deleted_at_condition_?

mohamalnadi: @johncrossman - makes sense! I just updated the name.

PR #5013, opened 2025-04-17
BOAC-5003, chip away at SQLAlchemy deprecations

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5003

Note: BOA's version of SQLAlchemy has not yet changed.
 
Migration guide: https://docs.sqlalchemy.org/en/20/changelog/migration_20.html


johncrossman:
I will retry later.

PR #5021, opened 2025-04-22
BOAC-6157: fixes add-to-curated-group bindings

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6157

v-model with argument (e.g. `v-model:selected`) is apparently a no-no. I think something in the recent package upgrades caused it to break in CuratedGroupSelector.vue and ManageStudent.vue.

It didn't seem to impact BoaUsers.vue and UserReport.vue, but I got rid of it there anyway so the linter won't complain.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -44,6 +44,7 @@ export default [
     languageOptions: {
       globals: {
         ...globals.node,
+        ...globals.browser
       },
       parser: parser,
       parserOptions: {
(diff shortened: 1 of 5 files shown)


eslint.config.js
```diff
@@ -44,6 +44,7 @@ export default [
     languageOptions: {
       globals: {
         ...globals.node,
+        ...globals.browser
```
lyttam, line 47: Adding this will get rid of lint warnings like `'window' is not defined  no-undef`

PR #5037, opened 2025-04-25
BOAC-6168, db-migrate SQL script to create peer-advising depts

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6168

The hard-coded IDs have been verified to line up in prod. (This is expected since we sync test envs w/ prod.) The release-instructions doc has been updated; this script will be run _after_ the KPUB dept is added.


johncrossman:
Closing this dev PR because we only need a QA PR, which is coming soon.

PR #5069, opened 2025-05-08
BOAC-5777, break cohort_controller into three controllers (readability, etc.)

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-5777

A slight refactor of the `cohort_filter_options` feed is less daunting when its code is not lost in the 500 lines of `cohort_controller`.


lyttam:
Agreed!

PR #5081, opened 2025-05-25
BOAC-6206, Calendly appointments in BOA

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6206


pauline2k:
Thanks, @johncrossman, looks good to me! Before doing a BOA-side merge, I think the team should get together and discuss timing for a 6.3 QA branch cut, and whether we want to include Calendly work in that release.

PR #5100, opened 2025-06-05
[QA] BOAC-6224, set lazy=False in db.replationship usage in note, alert and note_template.py

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6224

Jira description has the deets.


pauline2k:
It's never a _great_ idea to be lazy in a relationship.

PR #5114, opened 2025-06-20
BOAC-6205, show CalNet-sourced dept names in BOA monthly-notes-report

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6205

Read my latest comment in the Jira (above) to understand this workaround to a bad-data problem. I'm putting this fix on boa-dev so Flint can consider my lesser-evil solution.

*Note:* The server-side code is the important part of this PR.  The front-end code change is purely formatting.


pauline2k:
Seems reasonable to me.

PR #5186, opened 2025-08-05
BOAC-6289, fix breakage due to commas in note topics

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6289


johncrossman:
Fixing

PR #5195, opened 2025-08-07
BOAC-6292, PAM can add attachment to PA note

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6292


pfarestveit:
pam4pa

PR #5203, opened 2025-08-12
BOAC-6306, hang on to student metadata after edit; tidy up UX

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6306


johncrossman:
If a dev has a minute, this PR needs review. cc: @pauline2k @lyttam @mohamalnadi

PR #5206, opened 2025-08-13
BOAC-6311, front-end user session init is better with Promise.resolve

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6311

When user first authenticates, we load all department metadata and that takes time. Without a `promise`, the UI is rendered before user session init is complete. Thus, a `promise` is made.


pfarestveit:
Promises Promises

PR #5232, opened 2025-08-27
BOAC-6331: fixes sis_student_profile_data test failures

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6331

QA PR is #5231


johncrossman:
@lyttam - Looks good. Py linter has one small complaint.

PR #5240, opened 2025-08-28
BOAC-6337, enable some of our desired Ruff rules

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6337

Here is one worth noting: https://docs.astral.sh/ruff/rules/superfluous-else-break/


johncrossman:
I forgot to fix the violations in BEA code. Amendment coming soon.

PR #5255, opened 2025-10-03
BOAC-6367, /degrees: fix focus when archiving

johncrossman:
https://jira-secure.berkeley.edu/browse/BOAC-6367

### DIFF
--- a/src/views/degree/ManageDegreeChecks.vue
+++ b/src/views/degree/ManageDegreeChecks.vue
@@ -132,7 +132,7 @@ const onUpdateDegreeTemplate = (degreeTemplate: DegreeTemplate) => {
   if (degreeTemplate.archivedAt) {
     isShowingArchivedTemplates.value = true
   }
-  putFocusNextTick(`tr-degree-check-${degreeTemplate.id}`)
+  putFocusNextTick('page-header')
 }
 </script>
 


src/views/degree/ManageDegreeChecks.vue
```diff
@@ -132,7 +132,7 @@ const onUpdateDegreeTemplate = (degreeTemplate: DegreeTemplate) => {
   if (degreeTemplate.archivedAt) {
     isShowingArchivedTemplates.value = true
   }
-  putFocusNextTick(`tr-degree-check-${degreeTemplate.id}`)
+  putFocusNextTick('page-header')
```
lyttam, line 135: @johncrossman It's best to return focus to wherever the user last interacted with the page, if possible. If the button that the user clicked no longer exists on the page (or is disabled), it's nice to put focus on the next or previous focusable element. In this case I think it would be (a) one of the other buttons or links on the table row of the degree template that was just updated, or (b) if the degree template was deleted, the first button on the next row.
If none of that is feasible (like if there are no other interactable elements on the page, or if it feels counterintuitive), then putting focus on the page header makes sense. 
DAP dinged me on this same issue when testing the CSpace portals and provided this reference: https://www.w3.org/TR/2016/NOTE-WCAG20-TECHS-20161007/G13

johncrossman: Thank you @lyttam . Because I'm on vacation, I'm going to close this PR and unassign the Jira. We'll use your comment to do it right in the next PR.

PR #5271, opened 2026-01-07
BOAC-6434 Block login for disabled accounts

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-6434


pauline2k:
I believe Travis is failing on this PR because associated code in #5270 needs to be merged first.

PR #5338, opened 2026-02-03
boac-6503 - Add a custom directive to v-data-tables to render an sr-only caption

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-6503


lyttam:
Nice solution!

PR #5424, opened 2026-02-24
BOAC-6486: prevents JAWS saying 'clickable' on pill-like components

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6486

### DIFF
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -2,6 +2,18 @@
   max-width: max-content;
   width: 100%;
 }
+.alerts-count {
+  .v-badge__badge {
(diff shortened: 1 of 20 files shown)


src/assets/styles/global.css
```diff
@@ -489,6 +501,9 @@
   max-width: inherit;
   width: inherit;
 }
+.v-dialog--fullscreen.overflow-y-hidden > .v-overlay__content {
+  overflow-y: hidden !important;
+}
```
lyttam, line 506: Prevents double scrollbar when modal is in fullscreen mode.

PR #5475, opened 2026-03-10
BOAC-6539: puts modals in fullscreen mode on small screensf

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6539

### DIFF
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -306,9 +306,9 @@
   justify-content: flex-end !important;
   padding: 0 20px !important;
 }
-.modal-fullscreen {
-  min-width: 350px !important;
-  width: 100% !important;
+.modal-height-unset.v-dialog--fullscreen > .v-overlay__content {
+  height: unset !important;
+  top: unset !important;
 }
 .modal-open {
   overflow: hidden;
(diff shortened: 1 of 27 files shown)


src/assets/styles/global.css
```diff
@@ -555,6 +555,13 @@ hr {
   border: 0;
   border-top: 1px solid rgba(var(--v-border-color), var(--v-border-opacity));
 }
+html:has(body > .v-overlay-container > .v-overlay.v-overlay--active) {
+  overflow: hidden;
+}
+html:not(:has(body > .v-overlay-container > .v-overlay.v-overlay--active.v-dialog--fullscreen:not(.modal-height-unset))) {
+  scrollbar-gutter: stable !important;
+}
+
```
lyttam, line 564: No more need to manually add the `modal-open` class to the `<html>` tag when a modal opens.

src/assets/styles/global.css
```diff
@@ -306,9 +306,9 @@
   justify-content: flex-end !important;
   padding: 0 20px !important;
 }
-.modal-fullscreen {
-  min-width: 350px !important;
-  width: 100% !important;
+.modal-height-unset.v-dialog--fullscreen > .v-overlay__content {
+  height: unset !important;
+  top: unset !important;
```
lyttam, line 311: Small modals don't need to take up the full height in fullscreen mode.

PR #5530, opened 2026-04-01
boac-6626 - Fix Curated group select menu accessibility issues + update respective BEA tests

mohamalnadi:
**Jira:** https://jira-secure.berkeley.edu/browse/BOAC-6626


pauline2k:
Thanks for updating BEA too!

PR #5532, opened 2026-04-21
BOAC-6654: adds role to timeline messages

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6654

### DIFF
--- a/src/components/search/AdvancedSearchModal.vue
+++ b/src/components/search/AdvancedSearchModal.vue
@@ -1,10 +1,17 @@
 <template>
   <div>
-    <v-tooltip location="bottom" text="Advanced search options">
+    <v-tooltip
+      id="advanced-search-btn-tooltip"
+      aria-hidden="true"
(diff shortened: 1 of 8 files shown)


src/components/search/AdvancedSearchModal.vue
```diff
@@ -1,10 +1,17 @@
...
+      aria-hidden="true"
+      aria-roledescription="tooltip"
+      location="bottom"
+      text="Advanced search options"
+    >
       <template #activator="{props}">
         <v-btn
           id="search-options-panel-toggle"
           v-bind="props"
+          aria-describedby="advanced-search-btn-tooltip"
```
lyttam, line 14: Vuetify correctly implements the `aria-describedby` association, but I wanted to make it explicit.  This attribute causes a screen reader to announce the tooltip text when focus lands on the button, so also I added `aria-hidden` to the tooltip to prevent it from being announced twice.

PR #5539, opened 2026-04-29
BOAC-6377 Refactor batch degree check creation for performance

yijunge-ucb:
The "create_batch_degree_checks" function nows calls "_bulk_clone_all" instead of "clone".
The "clone" function is kept but refactored a little bit so that other functions that call it still work.
Each table is flushed once and there is only one commit for each chunk of sids.

### DIFF
--- a/boac/api/degree_progress_api_utils.py
+++ b/boac/api/degree_progress_api_utils.py
@@ -27,11 +27,18 @@ from datetime import datetime
 
 from flask_login import current_user
 
+from boac import db, std_commit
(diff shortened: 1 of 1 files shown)


pauline2k:
@yijunge-ucb, this all looks great. Thanks for the comments helping to make sense of the flow. Let's get it on dev and see how it behaves.

pauline2k:
Jira link for reference: https://jira-secure.berkeley.edu/browse/BOAC-6377

PR #5540, opened 2026-04-30
BOAC-6663 - API's to edit and delete comments

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-6663

### DIFF
--- a/boac/api/notes_controller.py
+++ b/boac/api/notes_controller.py
@@ -30,7 +30,11 @@ from flask import current_app as app
 from flask_login import current_user
 
 from boac import std_commit
-from boac.api.decorators import advising_data_access_required, director_advising_data_access_required
+from boac.api.decorators import (
+    admin_required,
+    advising_data_access_required,
+    director_advising_data_access_required,
+)
 from boac.api.errors import BadRequestError, ForbiddenRequestError, ResourceNotFoundError
 from boac.api.util import (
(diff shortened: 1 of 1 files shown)


boac/api/notes_controller.py
```diff
@@ -135,6 +135,79 @@ def add_note_comment():
...
+    return tolerant_jsonify(get_boac_note_as_compatible_json(note=comment, note_read=note_read))
+
+
+@app.route('/api/note/delete_comment/<comment_id>', methods=['DELETE'])
+@admin_required decorator
+def delete_note_comment(comment_id):
+    comment = Note.find_by_id(note_id=comment_id)
+    if not comment or not comment.parent_note_id:
+        raise ResourceNotFoundError('Note not found')
+    can_user_delete = False
```
lyttam: @mohamalnadi We actually don't need lines 196-206 because `@admin_required` does all the authorization checking we need. It returns a 401 error before the `delete_note_comment` function is called when a non-admin tries to use this API. See https://github.com/ets-berkeley-edu/boac/blob/1c29e0fdbec6304af7eda5a96d5a2d3a6e7c20da/boac/api/decorators.py#L51

mohamalnadi: @lyttam - Thank you for your feedback! Removed the duplicate code.

PR #5636, opened 2026-06-25
BOAC-6701 - Fix issues in Safari with menu focusability

mohamalnadi:
Jira: https://jira-secure.berkeley.edu/browse/BOAC-6701

This bug took me down a rabbit hole, but I think I figured it out! 

A few links that helped: 
- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/button#clicking_and_focus
- https://github.com/vuetifyjs/vuetify/issues/22540
- https://mayank.co/blog/safari-focus/
- https://github.com/vuetifyjs/vuetify/issues/16739
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Browser_detection_using_the_user_agent

Consulted ChatGPT as well which helped!

### DIFF
--- a/src/assets/styles/global.css
+++ b/src/assets/styles/global.css
@@ -311,6 +311,15 @@
   overflow: hidden;
   scrollbar-gutter: stable !important;
 }
+/* Safari does not show :focus-visible on programmatically-focused menu items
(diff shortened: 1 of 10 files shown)


lyttam:
Thanks for figuring this out!

PR #5658, opened 2026-07-20
BOAC-6759: Merge beanstalk and custom cloudwatch agent config

dg-ucb:
The fetch-config post hook is not merging beanstalk provided config and our custom cloudwatch config in a way that reliably streams all logs we need. The 'append-config' amazon-cloudwatch-agent-ctl  argument is the proper way to handle this.


pauline2k:
Hi @dg-ucb - the change looks good, but something seems to be off with the commit history; that second "Merge branch 'ets-berkeley-edu:dev' into dev" commit shouldn't be part of the PR.

dg-ucb:
> Hi @dg-ucb - the change looks good, but something seems to be off with the commit history; that second "Merge branch 'ets-berkeley-edu:dev' into dev" commit shouldn't be part of the PR.

PR #5660, opened 2026-07-27
BOAC-6377: fix bugs

yijunge-ucb:
@pauline2k  I found the bug in the function, instead of passing "chunk", I passed the entire range of sids. You mentioned in the JIRA card that it takes about 30 seconds to run a chunk, so maybe it took 30 seconds to run 2000 users. 

Other changes are minor. I added an index on degree_progress_templates, though I don't expect it to have a significant impact on performance. I can remove that change if it could cause any complications in the production database.

Also added a test to run 5000 users (it will pass if it takes less than 120s). Locally it took less than 30s but the template is really tiny. A limit of number of users were added recently in the code so I had to bypass that.


pauline2k:
Makes sense, @yijunge-ucb, thanks!

PR #5677, opened 2026-08-18
BOAC-6773 BEA: fix timing issue on homepage load

pauline2k:
https://jira-secure.berkeley.edu/browse/BOAC-6773


lyttam:
The Travis job was stuck in queued status, so I canceled and re-started the build.

PR #5693, opened 2026-08-26
BOAC-6668: caches alert counts for curated groups

lyttam:
https://jira-secure.berkeley.edu/browse/BOAC-6668

For a user with a ton of cohorts and curated groups, this cuts the response time of `/api/profile/my` from ~33 seconds in my local environment to 15-18 seconds. The size of the response is reduced from ~900 KB to ~50 KB.

- removes the list of SIDs from each cohort in the API response because it isn't used.
- removes the LIMIT and ORDER params on the backend; the front-end never sends them.
- caches the alert count per each curated group and viewer_id.
- deletes from the cache when an alert is dismissed so that the count will be recalculated the next time it's needed.
- refreshes all cached curated group alert counts during the nightly job run, after the alerts are refreshed.
- fixes this bug:  when a cohort has only one alert and that alert gets dismissed, the alert_count was never getting set to zero because the SQL was excluding cohorts with no alerts.

### DIFF
--- a/boac/api/alerts_controller.py
+++ b/boac/api/alerts_controller.py
@@ -30,6 +30,7 @@ from boac.api.decorators import advisor_required
 from boac.lib.http import tolerant_jsonify
 from boac.models.alert import Alert
 from boac.models.cohort_filter import CohortFilter
+from boac.models.curated_group import CuratedGroup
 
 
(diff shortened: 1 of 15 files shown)


pauline2k:
Thanks, this looks efficient!

PR #5699, opened 2026-08-28
BOAC-6786 - Fix issue with Degree Checks undesired reordering list of courses

mohamalnadi:
**Jira:** https://jira-secure.berkeley.edu/browse/BOAC-6786

The underlying cause appears to be BOAC-6377  ([link](https://github.com/ets-berkeley-edu/boac/pull/5539/)) (not the color coding) as it improved degree-check creation performance by bulk copying template categories instead of copying them one at a time in display order. The bulk copy loaded categories from the database without a defined order. It seems postgres returns rows in an order that doesn’t match how they were originally created.

The UI sorts requirements by `createdAt`. On a new student check, those timestamps are set at copy time. So if the copy happened in the wrong order, the student check shows the wrong order. It seems that advisors typically add course requirements in order (i.e. they created MATH 51 first, and then create MATH 52 next...etc.) 

**This fix will only fix any newly created degree checks in the future.** However, any degree checks assigned to students after v.6.13 was released will need to be manually fixed. We can probably write a script that matches the `createdAt` times of the cloned Degree Check requirements to the original degree check requirements, but this may be a bit risky in production.

### DIFF
--- a/boac/api/degree_progress_api_utils.py
+++ b/boac/api/degree_progress_api_utils.py
@@ -368,6 +368,9 @@ def _get_categories(template):
     """Fetch all categories associated with a template."""
     return DegreeProgressCategory.query.filter_by(
         template_id=template.id,
+    ).order_by(
+        DegreeProgressCategory.created_at,
+        DegreeProgressCategory.id,
     ).all()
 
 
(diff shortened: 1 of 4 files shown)


lyttam:
@mohamalnadi Thanks for getting to the bottom of this!

